PR #3, opened 2024-09-05
fix : fixing retrieval for rag setup to get better response

viplismism:
- Used `query_engine` to get the response, instead of depeding upon `QueryEngineTool`


bolna/agent_manager/task_manager.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, assistant_name, task_id, task, ws, input_parameters=None, con
         self.average_synthesizer_latency = 0.0
         self.average_transcriber_latency = 0.0
         self.task_config = task
+        logger.info(f"Task config is : {self.task_config}")
```
prateeksachan, line 50: can you remove this log?

PR #93, opened 2024-12-16
feat: function calling in litellm

mubashshir-cloudmate:


prateeksachan:
Thanks for contributing this PR. I’m still testing it thoroughly.

PR #108, opened 2025-01-18
Added endpoints to gets all agent, single agent and Edit an agent

khushaljethava:
ambient noise only works when wav files are local_setup file and need to copy them using dockerfile.


prateeksachan:
hey @khushaljethava thanks for the PR. Can you also put in a basic doc for consuming these API endpoints? let me know if you want to collaborate on this.

Issue #110, opened 2025-01-24
error on fetching /all agents endpoint


anantsparrow:
bolna-app-1   | INFO:     223.185.62.122:30735 - "GET /all HTTP/1.1" 500 Internal Server Error
bolna-app-1   | 2025-01-24 19:48:32.808 ERROR  {quickstart_server} [get_all_agents] Error fetching all agents: Expecting value: line 5 column 1 (char 4)
bolna-app-1   | Traceback (most recent call last):
bolna-app-1   |   File "/app/quickstart_server.py", line 165, in get_all_agents
bolna-app-1   |     agents = [{ "agent_id": key, "data": json.loads(data) } for key, data in zip(agent_keys, agents_data) if data]
bolna-app-1   |   File "/app/quickstart_server.py", line 165, in <listcomp>
bolna-app-1   |     agents = [{ "agent_id": key, "data": json.loads(data) } for key, data in zip(agent_keys, agents_data) if data]
bolna-app-1   |   File "/usr/local/lib/python3.10/json/__init__.py", line 346, in loads
bolna-app-1   |     return _default_decoder.decode(s)
bolna-app-1   |   File "/usr/local/lib/python3.10/json/decoder.py", line 337, in decode
bolna-app-1   |     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
bolna-app-1   |   File "/usr/local/lib/python3.10/json/decoder.py", line 355, in raw_decode
bolna-app-1   |     raise JSONDecodeError("Expecting value", s, err.value) from None
bolna-app-1   | json.decoder.JSONDecodeError: Expecting value: line 5 column 1 (char 4)
bolna-app-1   | INFO:     223.185.62.122:14421 - "GET /all HTTP/1.1" 500 Internal Server Error

prateeksachan:
cc @khushaljethava

khushaljethava:
It seems there is an issue with fetching data from the Redis database.

prateeksachan:
hey @anantsparrow can you try again. @khushaljethava has pushed a fix for this issue.

anantsparrow:
Hi yes it has been fixed

PR #111, opened 2025-01-26
added error handling

khushaljethava:
it will  solve #110  this issue.

### DIFF
--- a/local_setup/quickstart_server.py
+++ b/local_setup/quickstart_server.py
@@ -157,9 +157,14 @@ async def get_all_agents():
         
         if not agent_keys:
             return {"agents": []}  
-
-
-        agents_data = await asyncio.gather(*(redis_client.get(key) for key in agent_keys))
+        agents_data = []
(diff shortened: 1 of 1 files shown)


local_setup/quickstart_server.py
```diff
@@ -157,9 +157,14 @@ async def get_all_agents():
...
-
-        agents_data = await asyncio.gather(*(redis_client.get(key) for key in agent_keys))
+        agents_data = []
+        for key in agent_keys:
+            try:
+                data = await redis_client.get(key)
+                agents_data.append(data)
+            except Exception as e:
+                logger.error(f"An error occurred with key {key}: {e}")
+                agents_data.append(None)
```
prateeksachan: @khushaljethava Can youo remove adding `None` to the array in case of exception. I don't think it'll server any purpose.

khushaljethava: @prateeksachan please check.

PR #193, opened 2025-05-22
Fea: Logger sepration 

belikewater2:
One major issue I encountered while working on Bolna was the volume of logging. While the system understandably has many loggers (which is necessary), it becomes overwhelming when making changes to a specific module — for example, the synthesizer — and still seeing logs from unrelated components like the transcriber or LLM.

To address this, I’ve implemented logger bifurcation controlled via environment variables. This allows selective enabling/disabling of logs for specific modules without needing to modify the code.

Leave blank in env for all logs
Mention specific module to log that only 

Feel free too reach me out if pr needs any fixes


prateeksachan:
+1. this was much needed. Thanks for raising this PR. I'll review it and merge it.

.env.sample
```diff
@@ -1,3 +1,7 @@
+# TRANSCRIBER,LLM,SYNTHESIZER,INFO
+# Error always gets logged
+LOGGING_LEVELS=SYNTHESIZER
```
prateeksachan, line 3: Apart from having this control in the `.env` file, can you pipe the logs from different components to different file patters? As pratically this wouldn't be useful as everytime one would need to change the env variable and restart the application in production.

Maybe you can pipe the logs to `<some-common-file-name>.llm.log`, `<some-common-file-name>.synth.log`, etc?

prateeksachan (changes requested):
as mentioned here: https://github.com/bolna-ai/bolna/pull/193/files#r2102083340

.env.sample
belikewater2: sure

belikewater2:
image
@prateeksachan  now we are rotation the logs on the basis of size and file count of 5 file backup 
We will show all the logs but store specific log in specific file

image
Hardcoded or kind off closed file_name because otherwise it will create log file for every class

PR #308, opened 2025-10-18
fix(agent): Speak welcome message on outbound calls

yashsahane-eng:
### Summary
- Added logic for playing the agent welcome message on outbound calls.
- Tested via local setup in Codespaces with Twilio integration.
- Verified that inbound and outbound call behavior remains consistent.

### Details
- Updated `bolna/agent_manager/assistant_manager.py`
- Ensured agent config supports `agent_welcome_message` field.


prateeksachan:
this is already handled in the project code.

PR #370, opened 2025-11-21
refactor: split models.py into organized modules

HarshMethwani:
Split  models.py into 4 modules (providers, agents, rag, tasks) for better organization. All imports remain backward compatible.


prateeksachan:
Hey, thanks for the contribution - but it has a lot of conflicts. We'll try to pick this up internally too. Thank you.

PR #444, opened 2026-01-12
Feature/graph agent multilingual examples

san0808:


bolna/agent_manager/task_manager.py
```diff
@@ -1457,8 +1490,6 @@ async def __execute_function_call(self, url, method, param, api_token, headers,
         self.check_if_user_online = False
 
         if called_fun.startswith("transfer_call"):
-            await asyncio.sleep(2)
```
prateeksachan, line 1467: why removing this?

bolna/agent_manager/task_manager.py
```diff
@@ -1496,6 +1527,9 @@ async def __execute_function_call(self, url, method, param, api_token, headers,
                 logger.info(f"Gotten response {resp}")
                 payload = {**payload, **resp}
 
+            if self.tools['input'].io_provider != 'default':
```
prateeksachan, line 1538: why this addition??

bolna/agent_manager/task_manager.py
san0808: this is adding latency to transfer call , was it meant to be present for some reason ?

bolna/agent_manager/task_manager.py
san0808: I think it maybe redundant bt this ensures call_sid is set correctly in the final payload. The original call_sid assignment happens earlier (line 1509-1511), but payload = {**payload, **resp} on line 1528 can overwrite it if resp contains a call_sid key. This re-assignment after the merge guarantees the real provider call_sid is always in the payload sent to the transfer webhook. (had faced this bug when i was testing with rajnish)

bolna/agent_manager/task_manager.py
san0808: tell me if i am wrong but It was likely intended to wait for the TTS filler to finish playing, but there's already a proper dynamic wait later in the same function: while self.tools["input"].is_audio_being_played_to_user(): await asyncio.sleep(1) — which waits exactly until the audio finishes

so basically this is simply adding latency then

prateeksachan (approved):

PR #467, opened 2026-01-30
fix: use safe .get() for sequence key access in task_manager

### LINKED ISSUE #461: [Sentry] Fix KeyError 'sequence' in transcriber processing
## Problem

The transcriber task throws a KeyError when the 'sequence' key is missing from message metadata.

**Sentry Issue:** [DASHBOARD-BACKEND-K](https://bolna-voice-ai.sentry.io/issues/6804813978/)

**Error Message:**
```
Error in transcriber 'sequence'
```

**Impact:** ~70 errors/day, 13,571 total events (7.5% of all Sentry noise)

## Analysis

In `bolna/agent_manager/task_manager.py:2120`, the transcriber processing attempts to access `message["meta_info"]["sequence"]` without checking if the key exists.

Some messages may have `sequence_id` instead of `sequence`, or the key may be missing entirely.

**Related code:** `bolna/agent_manager/task_manager.py:2120`

san0808:
## Summary
- Uses `.get()` with fallback for `sequence` key access in 3 locations

## Problem
When `transcriber_connection_closed` message arrives with minimal meta_info `{'io': 'default', 'eos': True}`, it lacks `sequence` key. `process_transcriber_request()` is called before the message type check, causing KeyError.

**Sentry Issue:** [DASHBOARD-BACKEND-K](https://bolna-voice-ai.sentry.io/issues/6804813978/) - ~13,571 events

Fixes #461

## Changes
- `_extract_sequence_and_meta()`: Use `.get("sequence", .get("sequence_id", 0))`
- `process_transcriber_request()`: Same pattern + safe `request_id` access
- `__send_first_message()`: Same pattern

## Why 0 is safe
- For `transcriber_connection_closed`: we `break` immediately, sequence unused
- For normal messages: pipeline[0] is the standard audio pipeline
- `_get_next_step()` has its own try/except for invalid sequences

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -1149,8 +1149,8 @@ class TaskManager(BaseManager):
         sequence, meta_info = None, None
         if isinstance(message, dict) and "meta_info" in message:
             self._set_call_details(message)
-            sequence = message["meta_info"]["sequence"]
             meta_info = message["meta_info"]
+            sequence = meta_info.get("sequence", meta_info.get("sequence_id", 0))
         return sequence, meta_info
 
     def _is_extraction_task(self):
@@ -1778,10 +1778,11 @@ class TaskManager(BaseManager):
     # Transcriber task
     #################################################################
     async def process_transcriber_request(self, meta_info):
-        if not self.current_request_id or self.current_request_id != meta_info["request_id"]:
-            self.previous_request_id, self.current_request_id = self.current_request_id, meta_info["request_id"]
+        request_id = meta_info.get("request_id")
+        if request_id and (not self.current_request_id or self.current_request_id != request_id):
+            self.previous_request_id, self.current_request_id = self.current_request_id, request_id
 
-        sequence = meta_info["sequence"]
+        sequence = meta_info.get("sequence", meta_info.get("sequence_id", 0))
 
         # check if previous request id is not in transmitted request id
         if self.previous_request_id is None:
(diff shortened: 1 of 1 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -1149,8 +1149,8 @@ def _extract_sequence_and_meta(self, message):
         sequence, meta_info = None, None
         if isinstance(message, dict) and "meta_info" in message:
             self._set_call_details(message)
-            sequence = message["meta_info"]["sequence"]
             meta_info = message["meta_info"]
+            sequence = meta_info.get("sequence", meta_info.get("sequence_id", 0))
```
SubhrajitPrusty: The default value here for sequence is returning the id. Is that intentional?

SubhrajitPrusty (changes requested):

bolna/agent_manager/task_manager.py
san0808: Updated to use `0` as the fallback instead.

This fallback triggers when `transcriber_connection_closed` message arrives with minimal meta_info `{'io': 'default', 'eos': True}` which lacks the `sequence` key.

`sequence` is the pipeline index (0, 1) used in `self.pipelines[sequence]`. `sequence_id` is a turn counter (1, 2, 3...) for conversation tracking - using it as fallback could cause `pipelines[5]` access when only 1-2 pipelines exist.

Defaulting to `0` ensures we use the first/default pipeline. For `transcriber_connection_closed` case, we break immediately anyway so sequence is unused.

https://github.com/bolna-ai/bolna/pull/467/commits/eb36165

SubhrajitPrusty (approved):

PR #468, opened 2026-01-30
fix: add null check for message data in duration calculation

### LINKED ISSUE #465: [Sentry] Handle duration calculation exception in forced first message
## Problem

Duration calculation throws exceptions during forced first message handling.

**Sentry Issue:** [DASHBOARD-BACKEND-M](https://bolna-voice-ai.sentry.io/issues/)

**Error Message:**
```
Exception in __forced_first_message for duration calculation...
```

**Impact:** ~400 errors/day, 5,423 total events

## Analysis

The forced first message flow has an edge case where duration calculation fails, likely due to:
- Missing timestamp data
- Race condition in message timing
- Uninitialized variables

**Related code areas:**
- `bolna/agent_manager/task_manager.py` - `__forced_first_message` method
- Duration/timing calculation logic

san0808:
## Summary
Fixes #465

When `__forced_first_message` is called with empty text, `audio_chunk` is set to `None`. This results in `message["data"]` being `None`, and `len(None)` fails.

**Root cause flow:**
1. `meta_info['text'] == ''` → `audio_chunk = None` (line 680-681)
2. `create_ws_data_packet(None, meta_info)` → `message["data"] = None`
3. `len(message["data"])` → `TypeError: object of type 'NoneType' has no len()`

## Fix
Add null check before calling `len()`. When `message["data"]` is None, skip duration calculation and recording (no audio data to record).

**Sentry:** https://bolna-voice-ai.sentry.io/issues/6804869977/
**Events:** 5,428 total, 122 in last 24h

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -698,11 +698,12 @@ class TaskManager(BaseManager):
                     convert_to_request_log(message=text, meta_info=meta_info, component="synthesizer", direction="response", model=self.synthesizer_provider, is_cached=meta_info.get("is_cached", False), engine=self.tools['synthesizer'].get_engine(), run_id=self.run_id)
                     await self.tools["output"].handle(message)
                     try:
-                        if message["data"] is not None:
-                            duration = calculate_audio_duration(len(message["data"]), self.sampling_rate,
+                        data = message.get("data")
+                        if data is not None:
+                            duration = calculate_audio_duration(len(data), self.sampling_rate,
                                                                 format=message['meta_info']['format'])
                             self.conversation_recording['output'].append(
-                                {'data': message['data'], "start_time": time.time(), "duration": duration})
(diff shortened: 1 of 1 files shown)


SubhrajitPrusty (changes requested):
Multiple repeats of access to message["data"]. Let's first use .get to avoid KeyErrors. Then use a variable to reduce code repetition.

Then value of duration is never filled if data is not present. Is that intentional?

san0808:
Updated to use `.get()` and a variable for cleaner code.

For `duration` not being filled when data is None - yes, intentional. When `data` is None, we skip the recording (nothing to record since there's no audio data). We immediately `break` after this block, and `duration` is only used inside the `append()` call which we skip.

Let me know if you were thinking about something else that might be affected.

https://github.com/bolna-ai/bolna/pull/468/commits/da3ca06

SubhrajitPrusty (approved):

PR #470, opened 2026-01-30
fix: prevent asyncio tasks from being garbage collected while pending

### LINKED ISSUE #462: [Sentry] Fix async tasks destroyed while pending
## Problem

Async tasks are being garbage collected while still pending, indicating improper task lifecycle management.

**Sentry Issue:** [DASHBOARD-BACKEND-1G4](https://bolna-voice-ai.sentry.io/issues/7112549521/)

**Error Message:**
```
Task was destroyed but it is pending!
task: <Task pending name='Task-27634' coro=<...> wait_for=<Future pending cb=[Task.task_wakeup()]>>
```

**Impact:** ~130 errors/day, 6,650 total events

## Analysis

Background tasks created with `asyncio.create_task()` are not being properly awaited or cancelled when the parent context ends. This typically happens when:
- WebSocket connection closes while tasks are running
- Tasks are created but references aren't stored
- Cleanup doesn't await/cancel all spawned tasks

**Related code areas:**
- `bolna/agent_manager/task_manager.py` - task creation and cleanup
- Any place using `asyncio.create_task()` without storing reference

## Related Issues
- Also see: DASHBOARD-BACKEND-1XB (480 events), DASHBOARD-BACKEND-1XC (238 events)

san0808:
## Summary
Fixes #462

Asyncio tasks created with `asyncio.create_task()` but not stored can be garbage collected while still running, causing "Task was destroyed but it is pending!" errors.

## Root Cause

| Location | Issue |
|----------|-------|
| `telephony.py:45` | `disconnect_stream()` task not stored - API call to Plivo/Vobiz can be GC'd mid-request |
| `utils.py:583` | `write_request_logs()` task not stored - file write can be GC'd before completion |

## Fix

1. **telephony.py**: Store task in `self.disconnect_task` instance variable
2. **utils.py**: Add `create_background_task()` helper using standard pattern:
   - Tasks added to module-level set
   - `add_done_callback(set.discard)` removes task when complete
   - No memory leak

## Testing
```python
task = create_background_task(slow_task())
del task  # Remove local reference
gc.collect()  # Task survives GC
await asyncio.sleep(0.3)  # Task completes and is cleaned up
```

**Sentry:** https://bolna-voice-ai.sentry.io/issues/7112549521/
**Events:** 944 in 7 days

### DIFF
--- a/bolna/helpers/utils.py
+++ b/bolna/helpers/utils.py
@@ -28,6 +28,16 @@ from pydub import AudioSegment
 
 logger = configure_logger(__name__)
 load_dotenv()
+
+# Background task set to prevent fire-and-forget tasks from being garbage collected
+_background_tasks = set()
+
+def create_background_task(coro):
+    """Create a background task that won't be garbage collected until complete."""
+    task = asyncio.create_task(coro)
+    _background_tasks.add(task)
+    task.add_done_callback(_background_tasks.discard)
+    return task
 BUCKET_NAME = os.getenv('BUCKET_NAME')
 RECORDING_BUCKET_NAME = os.getenv('RECORDING_BUCKET_NAME')
 RECORDING_BUCKET_URL = os.getenv('RECORDING_BUCKET_URL')
@@ -580,7 +590,7 @@ def convert_to_request_log(message, meta_info, model, component="transcriber", d
     else:
         log['is_final'] = False #This is logged only for users to know final transcript from the transcriber
     log['engine'] = engine
-    asyncio.create_task(write_request_logs(log, run_id))
+    create_background_task(write_request_logs(log, run_id))
 
 
 def get_route_info(message, route_layer):
--- a/bolna/input_handlers/telephony.py
+++ b/bolna/input_handlers/telephony.py
@@ -25,6 +25,7 @@ class TelephonyInputHandler(DefaultInputHandler):
         self.last_media_received = 0
         self.io_provider = None
         self.websocket_listen_task = None
+        self.disconnect_task = None
 
     def get_stream_sid(self):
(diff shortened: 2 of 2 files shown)


bolna/helpers/utils.py
```diff
@@ -28,6 +28,16 @@
 
 logger = configure_logger(__name__)
 load_dotenv()
+
+# Background task set to prevent fire-and-forget tasks from being garbage collected
```
SubhrajitPrusty: nitpick: can you please move this below, after the constants?

SubhrajitPrusty:
BTW how did you find the offending tastes that were causing this issue?

There was a bug in Deepgram's Python SDK that used to throw the same error, ie Tasks destroyed. I don't remember if they fixed it or not. So it maybe coming from there.

san0808:
Based on my analysis, I found these by searching for `asyncio.create_task()` calls without stored references. Most of our code stores tasks in instance variables (e.g., `self.sender_task = asyncio.create_task(...)`), but two places were fire-and-forget without storing the reference:

- `telephony.py:45` - `asyncio.create_task(self.disconnect_stream())`
- `utils.py:583` - `asyncio.create_task(write_request_logs(log, run_id))`

Checked Sentry breadcrumbs - the errors happen during call cleanup (pricing, S3 upload, wallet update) which aligns with these locations. The stack traces don't show Deepgram SDK.

My hypothesis is that the Deepgram SDK bug you mentioned was in their library code , while this issue is in our own fire-and-forget task patterns. But let me know if you've seen something different - happy to investigate further but ya will check out once on sentry a bit more before i think i merge this

PR #478, opened 2026-02-09
Fix/audio playing state deadlock

san0808:


bolna/agent_manager/task_manager.py
```diff
@@ -1266,7 +1266,15 @@ async def agent_hangup_observer(self, is_agent_hangup):
             await self.__process_end_of_conversation()
 
     async def wait_for_current_message(self):
+        start_time = time.time()
+        max_wait = 10
```
SubhrajitPrusty: should this be configurable or a constant?

SubhrajitPrusty (approved):

PR #480, opened 2026-02-09
feat: sarvam v3 models

RG-04:
Added support for saaras:v3 and bulbul:v3.

Notes:
- Using speech to text for saaras:v3 with `mode` parameter as transcribe. Currently hardcoded as this is our only use case. In the future can consider using translate
- bulbul:v3 sends WAV header and then PCM chunks, unlike older versions which send full WAV files in each chunk
- Sampling rate as per docs for bulbul:v3 is 24000, but this is not observed in WAV header, so currently set to 22050
- The resample function now accepts PCM as well

### DIFF
--- a/bolna/helpers/utils.py
+++ b/bolna/helpers/utils.py
@@ -378,17 +378,73 @@ def convert_audio_to_wav(audio_bytes, source_format = 'flac'):
     return buffer.getvalue()
 
 
-def resample(audio_bytes, target_sample_rate, format = "mp3"):
+import io
+import torch
+import torchaudio
+import numpy as np
+
+def resample(audio_bytes, target_sample_rate, format="mp3", pcm_channels=1, original_sample_rate=None):
+    """
+    Resample audio bytes
+    
+    Args:
+        audio_bytes: Audio data as bytes
+        target_sample_rate: Target sample rate
+        format: Audio format ('wav', 'mp3', 'pcm', etc.)
+        original_sample_rate: Required if format='pcm'
+        pcm_channels: Number of channels for PCM (default: 1 mono)
+    """
+    
+    # Handle PCM separately
+    if format == "pcm":
+        if original_sample_rate is None:
+            raise ValueError("original_sample_rate must be provided for PCM format")
+        
+        audio_array = np.frombuffer(audio_bytes, dtype=np.int16)
+        
+        waveform = torch.from_numpy(audio_array).float() / 32768.0
+        
+        if pcm_channels == 1:
+            waveform = waveform.unsqueeze(0)  # (1, samples)
+        else:
+            # For multi-channel, interleaved PCM
+            waveform = waveform.reshape(pcm_channels, -1)
+        
+        if original_sample_rate == target_sample_rate:
+            return audio_bytes
+        
+        # Resample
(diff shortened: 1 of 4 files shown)


bolna/synthesizer/sarvam_synthesizer.py
```diff
@@ -294,8 +314,27 @@ async def generate(self):
...
+                        try:
+                            resampled_audio = resample(audio, int(self.sampling_rate), format=format, original_sample_rate=self.original_sampling_rate)
+                        except Exception as e:
+                            logger.error(f"Error in resampling audio: {e}")
+                            continue
+                        
+                        if format == "wav":
+                            audio = wav_bytes_to_pcm(resampled_audio)
+                        else:
+                            audio = resampled_audio
```
SubhrajitPrusty: very situational logic in the main method, can you move to a helper?

bolna/transcriber/sarvam_transcriber.py
```diff
@@ -58,8 +58,9 @@ def __init__(
         self.api_key = kwargs.get("transcriber_key", os.getenv("SARVAM_API_KEY"))
         self.api_host = os.getenv("SARVAM_HOST", "api.sarvam.ai")
 
-        # saaras models use translate endpoint, saarika models use transcription endpoint
-        if model.startswith("saaras"):
+        # old saaras models use translate endpoint, saarika models use transcription endpoint
+        # saaras:v3 supports direct transcription
+        if model.startswith("saaras") and model != "saaras:v3":
```
SubhrajitPrusty: can you move these logic that are checking for specific models to a separate helper?

RG-04: not sure what you mean, isn't _get_ws_url already a helper that is doing these model checks

bolna/synthesizer/sarvam_synthesizer.py
RG-04: addressed in https://github.com/bolna-ai/bolna/pull/480/commits/d2a48df0a14bb6a1b04c241adca215ad3ff291b6

bolna/transcriber/sarvam_transcriber.py
SubhrajitPrusty: what i mean is that _get_ws_url is also repeating the same checks, as well as the __init__. Why not move this to a singular place. ie. line 63 and line 128 are almost the same condition.

bolna/helpers/utils.py
```diff
@@ -378,19 +378,70 @@ def convert_audio_to_wav(audio_bytes, source_format = 'flac'):
...
+    """
+    
+    # Handle PCM separately
+    if format == "pcm":
+        if original_sample_rate is None:
+            raise ValueError("original_sample_rate must be provided for PCM format")
+        
+        audio_array = np.frombuffer(audio_bytes, dtype=np.int16)
+        
+        waveform = torch.from_numpy(audio_array).float() / 32768.0
```
SubhrajitPrusty: Can you move this magic number to some const?

SubhrajitPrusty (changes requested):

SubhrajitPrusty (approved):

PR #486, opened 2026-02-11
call audio flow fix

amansrivstv:


bolna/agent_manager/task_manager.py
```diff
@@ -2257,7 +2261,6 @@ async def _listen_transcriber(self):
                             welcome_played=self.tools["input"].welcome_message_played()
                         ):
                             logger.info(f"Continuing the loop and ignoring the transcript received ({transcript_content}) in speech final as it is false interruption")
-                            continue
```
prateeksachan, line 2260: why removing this change?

amansrivstv: resolved

san0808 (approved):

PR #490, opened 2026-02-13
fix: cancel all orphaned asyncio tasks on conversation cleanup

san0808:
## Summary
- Cancel 6 previously missed asyncio tasks (`llm_task`, `llm_queue_task`, `filler_task`, `first_message_task_new`, `execute_function_call_task`, `synthesizer_tasks` list) in the `finally` block of `TaskManager.run()`
- Fix `handle_cancellation()` to only cancel tasks owned by the current `TaskManager` instance instead of `asyncio.all_tasks()` which killed all concurrent conversations
- Fix fire-and-forget `_safe_disconnect_stream` task in telephony input handler by storing reference and awaiting completion

## Problem
[DASHBOARD-BACKEND-1G4](https://bolna-voice-ai.sentry.io/issues/7112549521/) — 9,040 events of "Task was destroyed but it is pending!" across `/{provider}/{call_type}/{run_id}`, `/healthz`, and `/call-me` endpoints.

Root cause: `TaskManager` creates ~25 asyncio tasks per conversation but only cancelled ~12 when the conversation ended. The remaining ~13 leaked as orphaned pending tasks.

### DIFF
--- a/bolna/input_handlers/telephony.py
+++ b/bolna/input_handlers/telephony.py
@@ -51,8 +51,7 @@ class TelephonyInputHandler(DefaultInputHandler):
     async def stop_handler(self):
         logger.info("stopping handler")
         self.running = False
-        # Fire and forget disconnect_stream - don't block the disconnection flow
-        asyncio.create_task(self._safe_disconnect_stream())
+        self._disconnect_task = asyncio.create_task(self._safe_disconnect_stream())
         logger.info("sleeping for 2 seconds so that whatever needs to pass is passed")
         await asyncio.sleep(2)
         try:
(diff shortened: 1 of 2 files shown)


bolna/input_handlers/telephony.py
```diff
@@ -51,15 +51,20 @@ async def _safe_disconnect_stream(self):
...
+        self._disconnect_task = asyncio.create_task(self._safe_disconnect_stream())
         logger.info("sleeping for 2 seconds so that whatever needs to pass is passed")
         await asyncio.sleep(2)
         try:
             await self.websocket.close()
             logger.info("WebSocket connection closed")
         except Exception as e:
             logger.info(f"Error closing WebSocket: {e}")
+        # Ensure disconnect task is complete before returning
+        if self._disconnect_task and not self._disconnect_task.done():
```
prateeksachan, line 63: will this block the execution?

PR #494, opened 2026-02-15
refactor: extract ConversationHistory and ToolCallAccumulator

san0808:
Extracts two reusable classes from task_manager and LLM providers:

- **ConversationHistory** — encapsulates all history append/pop/merge/sync operations previously scattered across 30+ locations in task_manager.py
- **ToolCallAccumulator** — consolidates near-identical tool call streaming logic from openai, azure, and litellm providers into one class

Pure refactor, zero behavioral change.

PR #495, opened 2026-02-15
feat: add OpenAI Responses API support

san0808:
Adds opt-in OpenAI Responses API support behind `use_responses_api` flag.

- Streaming and non-streaming Responses API paths in OpenAiLLM
- `previous_response_id` chaining for server-side conversation state
- Stale response ID detection with automatic fallback to full history
- `response.failed` / `response.incomplete` stream event handling
- MessageFormatAdapter for Chat Completions ↔ Responses API translation
- LLMStreamChunk, LatencyData, FunctionCallPayload typed definitions
- Constants for all event types, model prefixes, and error hints

Depends on #494.


bolna/agent_manager/task_manager.py
```diff
@@ -262,6 +262,10 @@ def __init__(self, assistant_name, task_id, task, ws, input_parameters=None, con
                 if 'reasoning_effort' in self.llm_agent_config:
                     self.llm_config['reasoning_effort'] = self.llm_agent_config['reasoning_effort']
 
+                llm_level_config = self.llm_agent_config.get('llm_config', self.llm_agent_config)
```
SubhrajitPrusty: default is the value itself?

bolna/llms/format_adapter.py
```diff
@@ -0,0 +1,73 @@
...
+
+class MessageFormatAdapter:
+    @staticmethod
+    def chat_to_responses_input(messages: list[dict]) -> tuple[str, list[dict]]:
+        """Chat Completions messages -> (instructions, Responses API input items)."""
+        instructions = ""
+        input_items = []
+
+        for msg in messages:
+            role = msg.get("role")
```
SubhrajitPrusty: LLM Messages are very well documented fixed structures, please use a typed struct / pydantic model to deserialize

bolna/llms/llm.py
```diff
@@ -13,3 +13,6 @@ async def respond_back_with_filler(self, messages):
 
     async def generate(self, messages, stream=True, ret_metadata=False):
         pass
+
+    def invalidate_response_chain(self):
```
SubhrajitPrusty, line 19: User defined base classes should be abstract and raise NotImplementedError [ref](https://docs.python.org/3/library/exceptions.html#NotImplementedError)

bolna/llms/types.py
```diff
@@ -0,0 +1,34 @@
...
+    param: Any
+    api_token: str | None
+    headers: dict | None
+    model_args: dict
+    meta_info: dict
+    called_fun: str
+    model_response: list[dict]
+    tool_call_id: str
+    textual_response: str | None
+    resp: Any
```
SubhrajitPrusty: use pydantic for easier conversion + serialization support.

bolna/constants.py
```diff
@@ -8,6 +8,25 @@
...
+RESPONSE_INCOMPLETE = "response.incomplete"
+RESPONSE_OUTPUT_TEXT_DELTA = "response.output_text.delta"
+RESPONSE_OUTPUT_ITEM_ADDED = "response.output_item.added"
+RESPONSE_FUNCTION_CALL_ARGS_DELTA = "response.function_call_arguments.delta"
+
+# Responses API input item types
+ITEM_TYPE_MESSAGE = "message"
+ITEM_TYPE_FUNCTION_CALL = "function_call"
+ITEM_TYPE_FUNCTION_CALL_OUTPUT = "function_call_output"
+ITEM_TYPE_FUNCTION = "function"
```
SubhrajitPrusty: These should be Enum classes

SubhrajitPrusty (changes requested):

bolna/agent_manager/task_manager.py
san0808: Done. Removed the extra `.get('llm_config', ...)` — by that point `llm_agent_config` already holds the flat config dict.

Commit: https://github.com/bolna-ai/bolna/commit/48ce26eae2f686140321e928ea71a0f6d3568449

bolna/llms/format_adapter.py
san0808: Done. Added Pydantic models (`ChatMessage`, `ChatToolCall`, `ChatToolDefinition`, etc.) and the adapter now parses dicts into them before accessing attributes.

Let me know if you think this is overkill for what's essentially a dict-to-dict conversion — happy to simplify if so.

Commit: https://github.com/bolna-ai/bolna/commit/6181fa2a93a4e02e09cb07b2b4e13162661a0640

bolna/llms/llm.py
san0808: Added docstrings to all three methods. Kept `pass` instead of `NotImplementedError` because `invalidate_response_chain` is an optional hook — only OpenAI implements it, and raising would force empty overrides in Azure/LiteLLM. The existing `generate` and `respond_back_with_filler` also already use `pass`.

Commit: https://github.com/bolna-ai/bolna/commit/219b94a014f98023e3dabce548bbd3f6d834dabd

bolna/llms/types.py
san0808: Done. Converted all three to Pydantic BaseModel. Also updated all producers (openai, azure, litellm, graph_agent, knowledgebase_agent) and the consumer in task_manager to use attribute access and `.model_dump()` at serialization boundaries.

Commit: https://github.com/bolna-ai/bolna/commit/12c277c038782445de95dcb44ff74ff757c4e6b7

bolna/constants.py
san0808: Done. Moved to `ResponseStreamEvent` and `ResponseItemType` str Enum classes in `bolna/enums.py`. Removed the constants from `constants.py` and updated all usages in `openai_llm.py` and `format_adapter.py`.

Commit: https://github.com/bolna-ai/bolna/commit/ff28de634d40f0480bffd3a09ffb87d460099d27

bolna/agent_types/knowledgebase_agent.py
```diff
@@ -295,9 +296,9 @@ async def generate(self, message: List[dict], **kwargs) -> AsyncGenerator[Union[
...
-                "first_token_latency_ms": 0,
-                "total_stream_duration_ms": now_ms() - start_time
-            }
-            yield f"An error occurred: {str(e)}", True, latency_data, False, None, None
+            latency_data = LatencyData(
+                sequence_id=meta_info.get("sequence_id") if meta_info else None,
+                first_token_latency_ms=0,
+                total_stream_duration_ms=now_ms() - start_time,
+            )
+            yield LLMStreamChunk(data=f"An error occurred: {str(e)}", end_of_stream=True, latency=latency_data)
```
SubhrajitPrusty, line 304: I see this same pattern in other files. Can be moved to a helper / base class method if repeated.

bolna/llms/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .openai_llm import OpenAiLLM
 from .litellm import LiteLLM
 from .azure_llm import AzureLLM
+from .types import LLMStreamChunk, LatencyData, FunctionCallPayload
```
SubhrajitPrusty: if this is define here in __init__, then these imports should be directly referenced in same level.

eg:
```
from bolna.llm import LLMStreamChunk, ...
```

But I see we're still using dot imports, which is not a clean way of writing.

bolna/llms/message_models.py
SubhrajitPrusty, line 1: maybe the filename has to change, now that it has more Pydantic models

SubhrajitPrusty (approved):

bolna/agent_types/knowledgebase_agent.py
san0808: Acknowledged will keep it inline for now since it's only 2 files. Can extract to a helper if more agents need it later but got the gist

bolna/llms/__init__.py
san0808: yepp makes sense
 Commit: https://github.com/bolna-ai/bolna/commit/9b0c77aa265b4a2fe54436bd74cc10dba068486b

PR #504, opened 2026-02-18
chore: remove unused deps, restructure pyproject.toml

san0808:
## Summary
- Removed unused packages: `llama_index`, `llama-index-vector-stores-lancedb`, `lancedb`, `pymongo` (was also duplicated)
- Restructured `pyproject.toml` with inline dependencies instead of dynamic loading from requirements.txt
- Separated provider-specific deps into optional extras: `azure`, `google`, `groq`, `plivo`, `twilio`, `classification`
- Added `[all]` extra to install everything at once
- Bumped `requires-python` to `>=3.10`
- Verified all remaining imports and resolution with `uv pip compile`


requirements.txt
```diff
@@ -1,36 +1,34 @@
...
 sentence-transformers==3.0.1
 tokenizers==0.15.2
+
+# Optional provider dependencies (included here for backward compatibility)
+azure-cognitiveservices-speech==1.38.0
+google-cloud-speech==2.25.0
+groq>=0.4.0
+plivo==4.47.0
+twilio==8.9.0
 optimum[onnxruntime]
```
SubhrajitPrusty, line 34: no version pinning on this?

SubhrajitPrusty: btw this is failing on mac arm systems. esp for 3.10 python.

SubhrajitPrusty:
since this is a significant change, we should upgrade the pkg version as well. should be a minor bump IMO

requirements.txt
san0808: anyway this optimum is getting resolved in #505

san0808: as in removing it , as for the failure i guess we can sit together once to understand the issue

MujeerAhmed (approved):

PR #514, opened 2026-02-20
fix: Handle missing pre-cached welcome audio in local/OSS telephony setup

abdulrahmanmajid:
## Summary

When running Bolna locally via the quickstart Docker setup with a telephony provider (e.g. Twilio), **no audio is heard on the phone** — the call connects, transcription works, the LLM responds, the synthesizer generates audio, but the caller hears complete silence. This PR fixes three related issues that cause this.

---

### Root Cause: `__forced_first_message()` sends `None` audio → crashes `TelephonyOutputHandler`

In `task_manager.py`, the method `__forced_first_message()` (used for telephony calls) expects pre-cached welcome audio to be passed via the `welcome_message_audio` kwarg. In the **production Bolna SaaS**, this audio is pre-synthesized and passed as base64 data. However, in the **local quickstart setup** (`quickstart_server.py`), no `welcome_message_audio` is ever provided:

```python
# quickstart_server.py line 198
assistant_manager = AssistantManager(agent_config, websocket, agent_id)
# ^ No welcome_message_audio kwarg
```

This means `preloaded_welcome_audio` is `None`, so `__forced_first_message()` creates a WebSocket data packet with `data=None` and sends it to the output handler.

In `TelephonyOutputHandler.handle()`, the very first operation on the audio chunk is `len(audio_chunk)`. When `audio_chunk` is `None`:

```python
len(None)  # → TypeError: object of type 'NoneType' has no len()
```

This exception is caught by the handler's `except` block, which sets `self._closed = True` — **permanently disabling the output handler**. Every subsequent audio chunk (including all LLM responses) is silently dropped because of the `if self._closed: return` guard at the top of `handle()`.

### Why it went undetected

1. **Production never hits this path**: The Bolna SaaS platform always pre-synthesizes and caches welcome audio before passing it to the agent. The `None` case simply never occurs in production.

2. **The failure is silent**: The `TypeError` is caught internally, sets `_closed = True`, and no error is logged. All subsequent audio silently returns early. Logs show the full pipeline working (transcription ✓, LLM response ✓, synthesis ✓, "Sending media event" ✓) — but no audio reaches the phone. This makes it extremely hard to diagnose.

3. **Other code paths handle this correctly**: The web-based `__first_message()` method properly calls `_synthesize()` as a fallback. The `__send_preprocessed_audio()` method explicitly handles `None` with a "File doesn't exist in S3. Hence we're synthesizing it from synthesizer" fallback. Only `__forced_first_message()` was missing this fallback.

4. **The OSS quickstart is a secondary path**: The quickstart Docker setup is primarily for evaluation/development, not the production-deployed path, so this gap was missed.

---

### 1. `task_manager.py` — Real-time welcome audio synthesis fallback (primary fix)

In `__forced_first_message()`, when no pre-cached welcome audio exists, synthesize it in real-time instead of sending a `None` packet. This matches the existing fallback behavior in `__first_message()` and `__send_preprocessed_audio()`:

```python
# If no pre-cached welcome audio, synthesize it in real-time
if audio_chunk is None and meta_info.get('text'):
    logger.info("No pre-cached welcome audio, synthesizing in real-time")
    meta_info['cached'] = False
    self.tools["input"].update_is_audio_being_played(True)
    await self._synthesize(create_ws_data_packet(meta_info['text'], meta_info=meta_info))
    break
```

**Backward compatible**: When cached audio IS present (production path), the original code path executes unchanged.

### 2. `telephony.py` — Defensive null-check guard (safety net)

Added a null check for `audio_chunk` before calling `len()` in `TelephonyOutputHandler.handle()`:

```python
if audio_chunk is None:
    logger.info("No audio data in packet, skipping media send")
    return
```

This prevents `None` audio from any source from crashing the output handler and permanently disabling it. This is a defensive measure — the task_manager fix above is the real solution.

### 3. `task_manager.py` — Fix `return` in `finally` block (code quality)

The `run()` method had a `return output` inside a `finally` block, which silently suppressed any exception raised in the `except` blocks (making `raise Exception(e)` dead code). Fixed by:
- Initializing `output = None` before the `try` block
- Removing the dead `raise Exception(e)` 
- Moving `return output` outside the `finally` block

This ensures exceptions propagate correctly while still returning conversation output after cleanup.

---

## Testing

Tested with the local Docker quickstart setup:
- **Twilio telephony** → Deepgram nova-3 transcriber → Groq llama-3.1-8b-instant LLM → Cartesia sonic-3-latest synthesizer
- Welcome message is now heard on phone (synthesized in real-time by Cartesia)
- Full conversation works: user speaks → transcribed → LLM responds → synthesized → heard on phone
- Call ends cleanly with full conversation transcript and latency data returned

## Files Changed

- `bolna/agent_manager/task_manager.py` — Welcome audio fallback + finally-block fix
- `bolna/output_handlers/telephony.py` — Null audio guard


abdulrahmanmajid:
Hi @MujeerAhmed 
Could someone please review this PR when available? I’d appreciate any feedback, and I’m happy to make further changes if needed. Thank you.

PR #518, opened 2026-02-24
fix: replace hardcoded sleep with mark-event wait before tool API call

san0808:
## Problem

Two sources of unnecessary latency on every tool call path:

1. `trigger_api` had a hardcoded `asyncio.sleep(0.7)` — a blind 700ms delay before every API call
2. `wait_for_current_message()` polled `mark_event_meta_data` every 500ms via `asyncio.sleep(0.5)` to check if telephony mark ACKs arrived — adding up to 500ms of dead time after marks clear

Together these added ~1.2s of wasted latency to every tool call, hangup, and end-of-conversation path.

## Fix

- Remove the hardcoded `sleep(0.7)` from `trigger_api`
- Call `wait_for_current_message()` before `trigger_api` in `__execute_function_call` so audio is properly flushed before the API fires (replacing the blind sleep with the real signal)
- Add `asyncio.Event` (`mark_changed`) to `MarkEventMetaData` — set on `fetch_data` (mark popped) and `clear_data` (interruption), not on `update_data`
- Replace `asyncio.sleep(0.5)` in `wait_for_current_message` with `asyncio.wait_for(mark_changed.wait(), ...)` so the waiter wakes immediately when marks are removed

All exit conditions in the wait loop are unchanged — only the sleep mechanism between checks is replaced.

## Files changed

- `bolna/helpers/function_calling_helpers.py` — remove `sleep(0.7)` from `trigger_api`
- `bolna/agent_manager/task_manager.py` — add `wait_for_current_message()` before tool API call, replace polling sleep with event wait
- `bolna/helpers/mark_event_meta_data.py` — add `mark_changed` event, signal on fetch/clear

### DIFF
--- a/bolna/helpers/mark_event_meta_data.py
+++ b/bolna/helpers/mark_event_meta_data.py
@@ -1,3 +1,4 @@
+import asyncio
 import copy
 
 from bolna.helpers.logger_config import configure_logger
@@ -10,6 +11,7 @@ class MarkEventMetaData:
         self.mark_event_meta_data = {}
         self.previous_mark_event_meta_data = {}
         self.counter = 0
+        self.mark_changed = asyncio.Event()
 
     def update_data(self, mark_id, value):
         value['counter'] = self.counter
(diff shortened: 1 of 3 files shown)


bolna/helpers/mark_event_meta_data.py
```diff
@@ -1,3 +1,4 @@
+import asyncio
```
prateeksachan, line 1: what's the use of this import here?

prateeksachan (approved):

bolna/helpers/mark_event_meta_data.py
san0808: for the line 14 below where i am using it ```  self.mark_changed = asyncio.Event()```

prateeksachan (approved):

PR #521, opened 2026-02-25
fix: wait for synth pipeline to flush before hangup using asyncio.Event

san0808:
## Problem

When the LLM triggers a hangup, `wait_for_current_message()` jumps straight to mark polling. If the final agent response is still being synthesized (no marks exist yet), it finds only a `pre_mark_message` and exits immediately — terminating the call before the last message is spoken.

This causes a mismatch: the transcript shows the final agent response, but the call recording cuts off before it's played.

## Fix

- Introduces an `asyncio.Event` (`_turn_audio_flushed`) that gates `wait_for_current_message()` until the synthesizer pipeline has flushed the current turn's audio
- The event is **cleared** when text enters the synthesizer and **set** when `end_of_synthesizer_stream` is detected across all output paths (streaming, non-streaming, blocked)
- A 3s timeout prevents indefinite blocking if something goes wrong
- Error/cancel paths in `_synthesize()` and `__listen_synthesizer()` also set the event to avoid unnecessary delays on failure

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -2287,9 +2287,11 @@ class TaskManager(BaseManager):
                 except asyncio.CancelledError:
                     logger.info("Synthesizer task was cancelled.")
                     #await self.handle_cancellation("Synthesizer task was cancelled.")
+                    self._turn_audio_flushed.set()
                     break
                 except Exception as e:
                     logger.error(f"Error in synthesizer: {e}", exc_info=True)
+                    self._turn_audio_flushed.set()
                     break
 
             logger.info("Exiting __listen_synthesizer gracefully.")
@@ -2392,6 +2394,7 @@ class TaskManager(BaseManager):
         except Exception as e:
             traceback.print_exc()
             logger.error(f"Error in synthesizer: {e}")
+            self._turn_audio_flushed.set()
 
     ############################################################
     # Output handling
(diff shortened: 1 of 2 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -2243,6 +2253,8 @@ async def __listen_synthesizer(self):
                                 logger.info("Stream not enabled, sending entire audio")
                                 # TODO handle is audio playing over here
                                 await self.tools["output"].handle(message)
+                                if meta_info.get('end_of_llm_stream', False):
+                                    self._turn_audio_flushed.set()
```
RG-04, line 2267: Why is this llm and not synth?

san0808: my bad ,have fixed it

RG-04:
We should also probably set in cleanup_downstream_tasks, because of the following 2 points

- https://github.com/bolna-ai/bolna/pull/521/changes#diff-fbc2323ec7b8dd29655f841e27d6c23a91b38f67867b25826d8ced517f9a3597L2083
- https://github.com/bolna-ai/bolna/pull/521/changes#diff-fbc2323ec7b8dd29655f841e27d6c23a91b38f67867b25826d8ced517f9a3597L2098

Plus its safer as well if missed anywhere.

prateeksachan (approved):

PR #522, opened 2026-02-25
fix: handle empty welcome message crashing telephony output handler

san0808:
## Problem

When an agent has no welcome message (empty string), `__forced_first_message` passes `data=None` to the telephony output handler. `len(None)` throws a `TypeError`, which silently sets `self._closed = True` (logged at DEBUG level). This permanently kills the output handler — **no audio reaches the caller for the entire call**.

## Changes

**`bolna/output_handlers/telephony.py`**
- Added None guard for `audio_chunk` before calling `len()`, preventing the crash

**`bolna/agent_manager/task_manager.py`**
- When welcome message is empty, skip calling `handle()` entirely
- Mark `is_welcome_message_played = True` immediately so the system doesn't wait for a mark event that will never arrive
- Don't set `is_audio_being_played = True` for empty audio (was causing user transcripts to be silently dropped)

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -670,19 +670,26 @@ class TaskManager(BaseManager):
                     logger.info(f"Got stream sid and hence sending the first message {stream_sid}")
                     self.stream_sid = stream_sid
                     await self.tools["output"].set_stream_sid(stream_sid)
-                    self.tools["input"].update_is_audio_being_played(True)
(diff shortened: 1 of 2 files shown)


RG-04 (approved):

prateeksachan (approved):

abdulrahmanmajid:
@san0808 Could you please check for the welcome audio not playing or respecting the "agent_welcome_message" in the agent config, it should generate it if no premade audio is provided.

PR #535, opened 2026-03-05
fix: defer stream_sid assignment until first audio frame 

amansrivstv:


bolna/input_handlers/telephony_providers/sip_trunk.py
```diff
@@ -167,6 +181,15 @@ async def _listen(self):
                     media_audio = message["bytes"]
                     if not media_audio:
                         continue
+                    # First audio frame from Asterisk proves the channel is now
```
SubhrajitPrusty, line 184: Does this mean that, before this fix, the call could appear connected before asterisk actually had put the call in the mixing bridge, hence the call appeared silent?

SubhrajitPrusty (approved):

bolna/input_handlers/telephony_providers/sip_trunk.py
amansrivstv: not sure of the fix, but maybe the bug is because the welcome audio is sent before the agent channel is added to the mixing bridge. that is why its getting blocked. This way the welcome message is sent only after the channel joins the mixing bridge. will test it out on prod and see if it fixes that race condition.

PR #543, opened 2026-03-09
Feat/gemini as llm

web1havv:
1.Created GeminiLLM class with full streaming, tool calling, and history translation from OpenAI format to Gemini format.

2.Registered GOOGLE in the LLMProvider enum and mapped it to GeminiLLM in the provider registry.

3.Added gemini-1.5-flash as the default Gemini model in constants and google-genai>=1.0.0 to requirements.txt

### DIFF
--- a/bolna/llms/gemini_llm.py
+++ b/bolna/llms/gemini_llm.py
@@ -139,13 +139,15 @@ class GeminiLLM(BaseLLM):
         return system_instruction, history
 
     def _build_config(self, system_instruction, request_json=False):
-        config = types.GenerateContentConfig(
+        config_kwargs = dict(
             system_instruction=system_instruction or None,
             max_output_tokens=self.max_tokens,
             temperature=self.temperature,
             response_mime_type="application/json" if request_json else "text/plain",
-            thinking_config=types.ThinkingConfig(thinking_budget=self.thinking_budget),
         )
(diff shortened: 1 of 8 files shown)


bolna/llms/gemini_llm.py
```diff
@@ -0,0 +1,278 @@
...
+        return system_instruction, history
+
+    def _build_config(self, system_instruction, request_json=False):
+        config = types.GenerateContentConfig(
+            system_instruction=system_instruction or None,
+            max_output_tokens=self.max_tokens,
+            temperature=self.temperature,
+            response_mime_type="application/json" if request_json else "text/plain",
+            thinking_config=types.ThinkingConfig(thinking_budget=0),
+        )
```
san0808, line 147: Avoid hardcoding `thinking_budget=0` here. Instead, accept it as a configurable kwarg (defaulting to `0` for voice AI latency), similar to how `openai_llm.py` handles `reasoning_effort` for GPT-5 series.

This way dashboard-backend can pass it through as a configurable param, giving users the option to trade latency for better quality on complex tool-call decisions if they want to.

Something like:
```python
self.thinking_budget = kwargs.get("thinking_budget", 0)
```

and then in `_build_config`:
```python
thinking_config=types.ThinkingConfig(thinking_budget=self.thinking_budget),
```

san0808 (approved):

PR #555, opened 2026-03-15
fix: add tool call message in prompt logs

RG-04:
Includes tools messages in the logs. This is correct behaviour as the formatted messages must be a true copy of the prompt passed to the LLM, which do include tool call messages.

### DIFF
--- a/bolna/helpers/utils.py
+++ b/bolna/helpers/utils.py
@@ -274,7 +274,8 @@ def format_messages(messages, use_system_prompt=False, include_tools=False):
                 formatted_string += "user: " + content + "\n"
         elif include_tools and role == 'tool':
             if content:
-                formatted_string += "tool_response: " + content + "\n"
+                tool_call_id = message.get('tool_call_id', '')
+                formatted_string += f"tool_response ({tool_call_id}): " + content + "\n"
 
     return formatted_string
 
--- a/bolna/helpers/utils.py
+++ b/bolna/helpers/utils.py
@@ -275,7 +275,7 @@ def format_messages(messages, use_system_prompt=False, include_tools=False):
         elif include_tools and role == 'tool':
             if content:
                 tool_call_id = message.get('tool_call_id', '')
-                formatted_string += f"tool_response ({tool_call_id}): " + content + "\n"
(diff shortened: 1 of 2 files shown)


san0808:
This change correctly includes `role: tool` messages in the log —> nice fix.

One thing I noticed while looking at this: `format_messages` in [`helpers/utils.py:252`](https://github.com/bolna-ai/bolna/blob/main/bolna/helpers/utils.py#L252) does `if message['content'] is None: continue`, which skips the entire message. When the LLM returns a pure tool call with no accompanying text, the assistant message has `content=None` (set at [`task_manager.py:1669-1670`](https://github.com/bolna-ai/bolna/blob/main/bolna/agent_manager/task_manager.py#L1669) via `textual_response = resp.get("textual_response", None)`). So the assistant's tool call decision itself still gets dropped from the log . you'd see the tool response but not *what function the LLM chose to call*.

Not something this PR introduces (pre-existing behavior), but since the goal is making logs a true copy of the prompt —> is this something you'd want to handle here or as a follow-up?

bolna/helpers/utils.py
```diff
@@ -249,22 +249,30 @@ def format_messages(messages, use_system_prompt=False, include_tools=False):
...
+                    except Exception as e:
+                        logger.warning(f"Error formatting tool call content: {e}")
         elif role == 'user':
-            formatted_string += "user: " + content + "\n"
+            if content:
+                formatted_string += "user: " + content + "\n"
         elif include_tools and role == 'tool':
-            formatted_string += "tool_response: " + content + "\n"
+            if content:
+                formatted_string += "tool_response: " + content + "\n"
```
san0808: Nit: since `assistant_tool_call` already includes the `id` via `str(tc)`, it'd be helpful to echo the `tool_call_id` here too so you can match responses to their calls when there are multiple tool calls:

```suggestion
                tool_call_id = message.get('tool_call_id', '')
                formatted_string += f"tool_response ({tool_call_id}): " + content + "\n"
```

san0808 (approved):
just solve the small nit and baki toh teek lag raha hai

PR #564, opened 2026-03-18
chore: refactor synth base class

RG-04:


bolna/synthesizer/azure_synthesizer.py
```diff
@@ -102,127 +87,94 @@ async def __generate_http(self, text):
...
-                synthesizer.synthesizing.connect(speech_synthesizer_synthesizing_handler)
-                synthesizer.synthesis_completed.connect(speech_synthesizer_completed_handler)
-
-                # Stamp synthesizer turn start time
-                try:
-                    meta_info['synthesizer_start_time'] = time.perf_counter()
-                except Exception:
-                    pass
+                            self._log_cancellation_error(cancellation)
+                    asyncio.run_coroutine_threadsafe(_set_done(), self.loop)
```
san0808, line 172: **Bug: this will deadlock on Azure TTS errors.**

`_log_cancellation_error()` raises an `Exception` (line 108). Since `on_completed` runs in a **thread callback** from the Azure SDK, the raise kills this callback before the next line executes. Result: `done_event` is never set → the `while not done_event.is_set()` loop at line 193 hangs forever.

The old code logged errors inline and always let execution reach the done-event setter. Was the raise intentional here?

Suggested fix — log inline, don't call the raising helper from a thread callback:
```python
def on_completed(evt):
    if evt.result.reason == speechsdk.ResultReason.Canceled:
        cancellation = evt.result.cancellation_details
        if cancellation.reason == speechsdk.CancellationReason.Error:
            error_code = cancellation.error_code
            error_details = cancellation.error_details
            logger.error(f"Azure TTS error: {error_details}")
            error_map = {
                CancellationErrorCode.AuthenticationFailure: "authentication failed",
                CancellationErrorCode.Forbidden: "forbidden",
                CancellationErrorCode.BadRequest: "bad request",
                CancellationErrorCode.ConnectionFailure: "connection failure",
            }
            msg = error_map.get(error_code, f"error (code: {error_code})")
            logger.error(f"Azure TTS {msg} - Region: {self.region}")
    asyncio.run_coroutine_threadsafe(_set_done(), self.loop)
```

bolna/synthesizer/base_synthesizer.py
```diff
@@ -48,6 +74,93 @@ async def cleanup(self):
...
+    async def _generate_http_loop(self):
+        """Standard HTTP (non-streaming) generate loop with caching support."""
+        while True:
+            message = await self.internal_queue.get()
+            logger.info(f"Generating TTS response for message: {message}")
+            meta_info, text = message.get("meta_info"), message.get("data")
+
+            if not self.should_synthesize_response(meta_info.get("sequence_id")):
+                logger.info(f"Not synthesizing: sequence_id {meta_info.get('sequence_id')} not current")
+                return
```
san0808, line 123: **`return` here permanently kills the async generator**: no more messages will ever be processed by this synthesizer instance.

The WS streaming path doesn't terminate the generator for stale sequence_ids. Should the HTTP path also keep going?

The old per-provider code also used `return`, so not a regression, but feels like it should be `continue`?

```suggestion
            if not self.should_synthesize_response(meta_info.get("sequence_id")):
                logger.info(f"Not synthesizing: sequence_id {meta_info.get('sequence_id')} not current")
                continue
```

bolna/synthesizer/base_synthesizer.py
```diff
@@ -48,6 +74,93 @@ async def cleanup(self):
...
+
+            audio = await self._fetch_http_audio(text, meta_info)
+            audio = self._process_http_audio(audio)
```
san0808, line 126: **What happens when `_generate_http` returns `None`?**

Polly returns `None` on `BotoCoreError`/`ClientError`, Azure returns `None` on cancellation. `_process_http_audio(None)` will either crash (e.g. Polly's `convert_audio_to_wav(None)`) or pass `None` through to `create_ws_data_packet`.

```suggestion
            audio = await self._fetch_http_audio(text, meta_info)
            if audio is None:
                logger.warning(f"No audio generated for text: {text[:50]}")
                continue
            audio = self._process_http_audio(audio)
```

bolna/synthesizer/base_synthesizer.py
```diff
@@ -48,6 +74,93 @@ async def cleanup(self):
...
+    def _get_http_audio_format(self):
+        """Output format string for HTTP mode (e.g. 'wav', 'mulaw')."""
+        return "wav"
+
+    async def _generate_http_loop(self):
+        """Standard HTTP (non-streaming) generate loop with caching support."""
+        while True:
+            message = await self.internal_queue.get()
+            logger.info(f"Generating TTS response for message: {message}")
+            meta_info, text = message.get("meta_info"), message.get("data")
```
san0808, line 119: **Missing `synthesizer_start_time` stamping for latency tracking.**

The old Deepgram/Azure HTTP paths stamped `meta_info['synthesizer_start_time'] = time.perf_counter()` before the TTS call and then computed `synthesizer_first_result_latency` / `synthesizer_total_stream_duration`. With this gone, latency tracking for HTTP-mode synths (Polly, OpenAI, non-streaming Deepgram) is lost. Was that intentional?

```suggestion
            message = await self.internal_queue.get()
            logger.info(f"Generating TTS response for message: {message}")
            meta_info, text = message.get("meta_info"), message.get("data")
            meta_info["synthesizer_start_time"] = time.perf_counter()
```

bolna/synthesizer/base_synthesizer.py
```diff
@@ -48,6 +74,93 @@ async def cleanup(self):
...
+                if meta_info is not None:
+                    meta_info["is_cached"] = True
+                return cached
+            logger.info("Not a cache hit")
+            if meta_info is not None:
+                meta_info["is_cached"] = False
+            self.synthesized_characters += len(text)
+            audio = await self._generate_http(text)
+            self.cache.set(text, audio)
+            return audio
```
san0808, line 153: **Caches `None` on provider failure.**

If `_generate_http(text)` returns `None` (error), this caches `None` —> all future requests for the same text will get a "cache hit" returning `None`, permanently breaking that text.

```suggestion
            self.synthesized_characters += len(text)
            audio = await self._generate_http(text)
            if audio is not None:
                self.cache.set(text, audio)
            return audio
```

bolna/synthesizer/stream_synthesizer.py
```diff
@@ -0,0 +1,336 @@
...
+
+class StreamSynthesizer(BaseSynthesizer):
+    """Base class for all WebSocket-streaming synthesizers."""
+
+    def __init__(self, stream=True, provider_name="stream", **kwargs):
+        super().__init__(
+            task_manager_instance=kwargs.get("task_manager_instance"),
+            stream=stream,
+            buffer_size=kwargs.get("buffer_size", 400),
+        )
```
san0808: **`**kwargs` silently swallows mistyped params** — e.g. `bufer_size=500` silently falls through to the default of 400 with no error. Would it be cleaner to make these explicit?

```suggestion
    def __init__(self, stream=True, provider_name="stream",
                 task_manager_instance=None, buffer_size=400, **kwargs):
        super().__init__(
            task_manager_instance=task_manager_instance,
            stream=stream,
            buffer_size=buffer_size,
        )
```

Subclasses would then pass `task_manager_instance=kwargs.get("task_manager_instance")` explicitly instead of forwarding the whole bag.

bolna/synthesizer/stream_synthesizer.py
```diff
@@ -0,0 +1,336 @@
...
+        """
+        return self._get_audio_format()
+
+    def _on_stream_end(self):
+        """Hook called when end-of-stream (b'\\x00') is received. Override for extra cleanup."""
+        pass
+
+    def _enrich_meta_info(self, meta_info, message):
+        """Hook to add provider-specific fields to meta_info per chunk."""
+        pass
```
san0808: **Neither `_on_stream_end` nor `_enrich_meta_info` is overridden by any subclass in this PR.** Are these speculative hooks for future use, or is there a planned consumer?

If they're anticipatory, I'd suggest removing them and adding when actually needed — fewer hooks = less cognitive overhead for contributors reading the base class.

```suggestion
```

bolna/synthesizer/base_synthesizer.py
RG-04: Yes did observe this, did not want to make it a part of this PR

bolna/synthesizer/base_synthesizer.py
RG-04: This is old behaviour, would prefer to address in a different PR

bolna/synthesizer/base_synthesizer.py
san0808: cool can u note it down or make sure we do it in a followup pr if necessary

bolna/synthesizer/base_synthesizer.py
san0808: again maybe note it down so we address if needed in followup pr

bolna/synthesizer/base_synthesizer.py
RG-04: The old code for HTTP would populate these yes, but as per my understanding synthesizer_start_time, synthesizer_total_stream_duration etc. were never used. My guess is at some point we migrated to turn_latencies logic and didn't remove this. Currently for HTTP I dont think we are tracking latencies at all, correct me if I am wrong

bolna/synthesizer/base_synthesizer.py
RG-04: Once again, old behaviour. Will prefer to address separately.

bolna/synthesizer/stream_synthesizer.py
RG-04: Were used in pixa which I reverted later. Forgot to remove, addressed in latest commit

bolna/synthesizer/base_synthesizer.py
san0808: `synthesizer_latency` is actively consumed in two places:

1. **`bolna/helpers/utils.py:669`** : structured logging reads `meta_info.get('synthesizer_latency')` for every synthesizer response log
2. **`dashboard-backend/routes/latency.py:77`** + **`utils/datadog/latency_updater.py:195`** : Datadog metrics pipeline fetches and stores synthesizer latency metrics

For WS synths the PR preserves it via `stream_synthesizer.py:259`, but for HTTP synths (Polly, OpenAI) the new `_generate_http_loop` never stamps it — so latency will always be `None` in logs/metrics for those providers.

`task_manager.py:653` already stamps `synthesizer_start_time` before pushing, so the fix is just computing the delta in `_generate_http_loop` after the HTTP call:

```python
audio = await self._fetch_http_audio(text, meta_info)
meta_info["synthesizer_latency"] = time.perf_counter() - meta_info.get("synthesizer_start_time", time.perf_counter())
```

san0808: note i understand maybe it is not important but just thinking out loud if it can be handled better or do we even need to handle or just remove it

let us ignore if too much hassle

bolna/synthesizer/base_synthesizer.py
RG-04: Yup will make sure to address these

bolna/synthesizer/stream_synthesizer.py
RG-04: Makes sense

bolna/synthesizer/stream_synthesizer.py
```diff
@@ -0,0 +1,330 @@
...
+    def __init__(self, stream=True, provider_name="stream",
+                 task_manager_instance=None, buffer_size=400, **kwargs):
+        super().__init__(
+            task_manager_instance=task_manager_instance,
+            stream=stream,
+            buffer_size=buffer_size,
+        )
+        self.provider_name = provider_name
+
+        # WebSocket state
```
san0808, line 47: **`websocket_holder = {"websocket": None}` => unnecessary dict wrapper**

This dict indirection is used dozens of times across 6+ files (`self.websocket_holder["websocket"]`). There's no reason for the wrapping , it adds noise to every read/write. A refactor PR is the ideal time to simplify this to just `self.ws` or `self.websocket`.

```suggestion
        self.websocket = None
```

And then all usages become e.g. `self.websocket` instead of `self.websocket_holder["websocket"]`.

bolna/synthesizer/base_synthesizer.py
RG-04: Hmm will revisit this, we are writing it to trace data but even that is not being used anywhere. It does not make sense that WS is being tracked in latency_dict while this is being tracked in trace data. Ideally either both should be tracked in both places, or we should get rid of the latency column in trace data altogether

bolna/synthesizer/stream_synthesizer.py
```diff
@@ -0,0 +1,330 @@
...
+            if self.stream:
+                async for packet in self._generate_ws_loop():
+                    yield packet
+            else:
+                async for packet in self._generate_http_loop():
+                    yield packet
+        except Exception as e:
+            traceback.print_exc()
+            logger.error(f"Error in {self.provider_name} generate: {e}")
+            raise
```
san0808, line 198: **Use logger instead of `traceback.print_exc()`**

`traceback.print_exc()` writes to stderr, bypassing structured logging. In production this means the traceback won't show up in log aggregation (Datadog, CloudWatch, etc.). Use `logger.error` with `exc_info=True` instead — you can also drop the `import traceback`.

```suggestion
        except Exception as e:
            logger.error(f"Error in {self.provider_name} generate: {e}", exc_info=True)
            raise
```

RG-04: Done

bolna/synthesizer/stream_synthesizer.py
RG-04: Addressed, makes sense

bolna/synthesizer/azure_synthesizer.py
RG-04: Addressed :+1:

san0808 (approved):

PR #566, opened 2026-03-19
feat: per-language handoff message for multilingual switch

san0808:
## Summary
Synthesizes a configurable handoff message with the current voice before switching languages. Each language can have its own handoff message template with `{voice_name}` and `{language}` placeholders.

## Changes
- Per-language `switch_handoff_messages` dict and `agent_names` dict read from `tools_config`
- `switch_language` block in `__execute_function_call`: resolves placeholders, synthesizes with current voice using `sequence_id: -1`, waits for playback, then switches pools
- `{language}` resolves to display name ("Hindi") via `LANGUAGE_NAMES` mapping, not code ("hi")
- `_get_voice_name_for_label()` helper looks up agent name for target language
- Skip synthesis for empty text in `_handle_llm_output` — eliminates 3s delay from empty pre-function-call push on tool calls
- Skip `wait_for_current_message` when no audio is pending, flush stale synth requests before handoff
- No-op when handoff message is empty — no regression for existing agents

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -1526,6 +1526,8 @@ class TaskManager(BaseManager):
             meta_info['message_category'] = 'filler'
 
         if next_step == "synthesizer" and not should_bypass_synth:
+            if not text_chunk or not text_chunk.strip():
+                return
             self._turn_audio_flushed.clear()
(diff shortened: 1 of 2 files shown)


RG-04:
switch_handoff_messages dict and agent_names should be added as an optional in attribute in ToolsConfig, otherwise looks good to me

RG-04 (approved):

PR #572, opened 2026-03-21
fix: log pre function call message correctly

RG-04:
Issue:
1) Any message before the function call in the same stream is not logged to trace data
2) The same response if spoken is not visible in the transcript if the user interrupts while the function call is running. This also means that the next LLM call gets incomplete context.

Fix:
Log this message both to trace data and conversation history early. If the tool call succeeds, edit this message to contain the tool call as well.

### DIFF
--- a/bolna/helpers/conversation_history.py
+++ b/bolna/helpers/conversation_history.py
@@ -40,6 +40,9 @@ class ConversationHistory:
             "content": content,
         })
 
+    def attach_tool_calls_to_last_response(self, tool_calls: list):
+        self._messages[-1]["tool_calls"] = tool_calls
+
     def update_system_prompt(self, content: str):
         if self._messages and self._messages[0].get("role") == ChatRole.SYSTEM:
             self._messages[0]["content"] = content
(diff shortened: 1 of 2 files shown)


prateeksachan (approved):

bolna/helpers/conversation_history.py
```diff
@@ -40,6 +40,9 @@ def append_tool_result(self, tool_call_id: str, content: str):
             "content": content,
         })
 
+    def attach_tool_calls_to_last_response(self, tool_calls: list):
+        self._messages[-1]["tool_calls"] = tool_calls
```
san0808: `_messages[-1]` can be a non-assistant message if `sync_after_interruption` removes the assistant msg during the async gap (between `__store_into_history` and this call, there's `wait_for_current_message` + the API call). Would cause `IndexError` on empty history or `tool_calls` on a user/tool message → invalid OpenAI conversation format.

```suggestion
    def attach_tool_calls_to_last_response(self, tool_calls: list):
        if self._messages and self._messages[-1].get("role") == ChatRole.ASSISTANT:
            self._messages[-1]["tool_calls"] = tool_calls
        else:
            logger.warning("attach_tool_calls_to_last_response: last message is not assistant, appending new")
            self.append_assistant(None, tool_calls=tool_calls)
```

PR #594, opened 2026-03-24
API Call log details added

vikramsingh-bolna:


bolna/agent_manager/task_manager.py
```diff
@@ -464,6 +468,95 @@ def __init__(self, assistant_name, task_id, task, ws, input_parameters=None, con
...
+
+    def _start_api_call_detail(
+        self,
+        *,
+        called_fun,
+        url,
+        method,
+        param,
+        headers,
+        model_args,
```
san0808: nit: `model_args` is accepted here but never stored in the `api_call_detail` dict below. Intentional or should it be included?

san0808 (approved):

PR #603, opened 2026-03-27
fix: choose correct language for filler messages

RG-04:
**Issue**
1) Incorrect language being chosen for filler messages after language switch
2) User online message stops functioning after language switch

**Fix**
1) Disable the language detector when a multilingual agent is chosen. This PR makes `language` a property of the `TaskManager` class, giving priority to detected language in the getter method.

2) `check_if_user_online` is temporarily set to `False` at the beginning of  `_execute_function_call`. It must be reset on the return path of `switch_language` tool as well, similar to the normal tools.

**Notes**
 - `check_if_user_online` should ideally never be changed. This flow should be handled better via `response_in_pipeline`. This should be addressed in a later PR since the changes are not related to the issues mentioned above.
- `_inject_language_instruction` behaviour remains the same (only functions if language detector is active).

### DIFF
--- a/bolna/helpers/language_detector.py
+++ b/bolna/helpers/language_detector.py
@@ -15,9 +15,10 @@ logger = configure_logger(__name__)
 class LanguageDetector:
     """Detects dominant language from user transcripts using LLM."""
 
-    def __init__(self, config: dict, run_id: str = None):
+    def __init__(self, config: dict, run_id: str = None, enabled: bool = True):
         self.turns_threshold = config.get('language_detection_turns') or 0
         self.run_id = run_id
+        self.enabled = enabled
 
         self._transcripts = []
         self._result = None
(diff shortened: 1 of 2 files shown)


bolna/helpers/language_detector.py
```diff
@@ -66,6 +63,19 @@ async def collect_transcript(self, transcript: str):
             self._in_progress = True
             self._task = asyncio.create_task(self._run_detection())
 
+    def set_enabled_status(self, enabled: bool):
+        """Enable or disable language detection."""
+        self.enabled = enabled
+        logger.info(f"[DEBUG] Language detection enabled set to {enabled}")
```
san0808: use logger.debug or remove the prefix if info

san0808 (approved):

PR #612, opened 2026-03-30
feat: track http synth latencies;  misc bugs

RG-04:
- Add Latency tracking for HTTP synthesizers
- Prevents early return in generate_http_loop (bug)
- Prevents setting cache on null responses from the synthesizer implementations, usually arising due to API errors

### DIFF
--- a/bolna/synthesizer/base_synthesizer.py
+++ b/bolna/synthesizer/base_synthesizer.py
@@ -1,4 +1,5 @@
 import io
+import time
 import uuid
 import asyncio
 import re
@@ -21,6 +22,7 @@ class BaseSynthesizer:
         self.first_chunk_generated = False
         self.synthesized_characters = 0
         self.model = "default"
+        self.current_turn_start_time = None
 
     # ------------------------------------------------------------------
     # Common accessors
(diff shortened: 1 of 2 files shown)


bolna/synthesizer/stream_synthesizer.py
```diff
@@ -146,7 +146,7 @@ async def push(self, message):
         if self.stream:
             await self._push_stream(message)
         else:
-            self.internal_queue.put_nowait(copy.deepcopy(message))
+            super().push(message)
```
san0808: `BaseSynthesizer.push()` is `async def` — doesn't this need an `await`? Without it the coroutine is created but never executed, so the message never reaches `internal_queue` and `_generate_http_loop` will hang on `get()`.

san0808: The original code did `copy.deepcopy(message)` before enqueuing. `BaseSynthesizer.push()` does a plain `put_nowait(message)` with no copy. Since `_generate_http_loop` mutates `meta_info` in-place (sets `format`, `text`, `mark_id`, etc.), this could corrupt the caller's original dict. Should the deepcopy be preserved here?

bolna/synthesizer/base_synthesizer.py
```diff
@@ -21,6 +22,7 @@ def __init__(self, task_manager_instance=None, stream=True, buffer_size=40, even
         self.first_chunk_generated = False
         self.synthesized_characters = 0
         self.model = "default"
+        self.current_turn_start_time = None
```
san0808, line 25: `current_turn_id` is set in `_stamp_turn_start` (line 108) and read in `_record_turn_latency` (line 116) but isn't initialized here. Might be worth adding `self.current_turn_id = None` alongside `current_turn_start_time` for safety.

PR #506, opened 2026-02-18
feat: add typed error attribution for call-breaking errors

san0808:
## Summary
- Replaces stringly-typed component/direction params with `LogComponent` and `LogDirection` enums across all CSV logging call sites
- Adds `BolnaComponentError` exception hierarchy (`LLMError`, `SynthesizerError`, `TranscriberError`) carrying component/provider/model context
- Call-breaking errors now log user-friendly attributed messages instead of raw exception strings
- Bad API keys (user-provided) now surface as call-breaking errors and terminate the call
- Propagates transcriber `connection_error` across all transcriber providers

## Test plan
- [ ] Verify CSV error messages show attributed friendly text for bad API keys (LLM, synthesizer, transcriber)
- [ ] Verify call terminates on bad API key
- [ ] Verify dashboard-backend still parses CSV correctly

PR #627, opened 2026-04-08
fix: break exception traceback reference cycles in TaskManager

san0808:
## Summary

- Fixes memory leak caused by CPython exception reference cycles that prevented GC of `TaskManager` instances after PR #506
- Clears `__traceback__` / `__cause__` / `__context__` before storing exceptions on `self._component_error`
- Removes `from e` exception chains in LLM, synthesizer, and transcriber error paths
- Reconstructs fresh exceptions in `run()` error surfacing block instead of re-raising originals
- Clears `_component_error` in `finally` block as safety net

## Context

After #506 introduced typed error attribution, storing caught exceptions on `self` creates `self -> exception -> __traceback__ -> frame -> f_locals -> self` reference cycles. Under high concurrency the cyclic GC can't keep up, so `TaskManager` instances (holding conversation history, audio buffers, WS connections) accumulate until OOM.

Details: https://www.notion.so/33ce00facdd681aab526dcdd8caf9056

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -1401,7 +1401,7 @@ class TaskManager(BaseManager):
                     str(e),
                     provider=self.llm_config.get("provider"),
                     model=self.llm_config.get("model")
-                ) from e
+                )
             latency_ms = (time.time() - start_time) * 1000
             
             if self.task_config["task_type"] == "summarization":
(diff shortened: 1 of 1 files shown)


bolna/agent_manager/task_manager.py
SubhrajitPrusty:
```python
exc = task.exception()
...
if exc is not None:
    raise cls(str(exc), provider=provider)
```
do this

san0808: but we will lose the error attribution right with this

san0808: acha nahi got it , this will be for that specific case

san0808: [a09f5d7](https://github.com/bolna-ai/bolna/pull/627/commits/a09f5d707c30c8c95f095337ef0fce20a1f5b3a5)

SubhrajitPrusty (approved):

PR #641, opened 2026-04-13
fix: cancel websocket listener task on call end

san0808:
## Summary

`stop_handler()` closes the WebSocket but never cancels the `_listen()` task. Since `_listen()` uses `while True` (not `while self.running`), the flag set in `stop_handler()` has no effect and the task can only exit via a "stop" event or exception.

Also fixes `_listen()` to `break` on `WebSocketDisconnect` instead of `pass` and the old code possibly looped back to `receive_text()` on a closed socket for codes 1000/1001/1006.

Same cancellation added to SIP trunk's `stop_handler()` for consistency.

### DIFF
--- a/bolna/input_handlers/telephony.py
+++ b/bolna/input_handlers/telephony.py
@@ -70,6 +70,14 @@ class TelephonyInputHandler(DefaultInputHandler):
         asyncio.create_task(self._safe_disconnect_stream())
         logger.info("sleeping for 2 seconds so that whatever needs to pass is passed")
         await asyncio.sleep(2)
+        # Cancel the listener task before closing the WebSocket to prevent zombie SSL reads
+        if self.websocket_listen_task and not self.websocket_listen_task.done():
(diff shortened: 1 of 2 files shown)


bolna/input_handlers/telephony.py
```diff
@@ -159,12 +167,11 @@ async def _listen(self):
                     break
 
             except WebSocketDisconnect as e:
-                if e.code in (1000, 1001, 1006):
-                    pass
-                else:
+                if e.code not in (1000, 1001, 1006):
```
prateeksachan, line 170: this change is functionally identical right?

prateeksachan (approved):

bolna/input_handlers/telephony.py
san0808: Not exactly na the old code did pass for 1000/1001/1006 which continued the while True loop back to receive_text() on a dead socket. The new code breaks out for all disconnect codes. The if inversion is just cleanup.

PR #670, opened 2026-04-29
feat: Deepgram Flux integration with multilingual + conversation tuning

san0808:
## Summary

* v2 Flux streaming with EagerEndOfTurn / TurnResumed / EndOfTurn machinery
* Multilingual support via repeatable language_hint param (flux-general-multi)
* Conversation defaults tuned: eager_eot_threshold=0.5, eot_timeout_ms=3000 — turns the speculative-LLM path on by default

## Knobs

* eot_threshold (0.5–0.9, default 0.7)
* eager_eot_threshold (default 0.5 for flux-* models)
* eot_timeout_ms (default 3000)
* language_hints (List[str], flux-general-multi only)


bolna/agent_manager/task_manager.py
```diff
@@ -230,7 +230,9 @@ def __init__(
 
         # Setup IO SERVICE, TRANSCRIBER, LLM, SYNTHESIZER
         self.llm_task = None
+        self.eager_llm_task = None
         self.llm_queue_task = None
+        self.eager_llm_task = None
```
san0808: is this a duplicate of line 233? same assignment twice.

bolna/agent_manager/task_manager.py
```diff
@@ -3211,6 +3218,45 @@ async def _listen_transcriber(self):
...
+                            logger.info(f"Skipping speculative LLM (audio playing or welcome not done)")
+
+                    elif isinstance(message.get("data"), dict) and message["data"].get("type", "") == "turn_resumed":
+                        logger.info(f"TurnResumed: Cancelling speculative LLM task")
+
+                        if self.eager_llm_task is not None:
+                            self.eager_llm_task.cancel()
+                            self.eager_llm_task = None
+
+                            if self.history and self.history[-1].get("role") == "user":
```
san0808: if the speculative LLM has already completed and pushed an assistant turn before TurnResumed lands, history[-1] would be the assistant, not the user. wouldn't this guard then no-op and leave a stale user/assistant pair behind?

bolna/agent_manager/task_manager.py
```diff
@@ -3245,8 +3291,15 @@ async def _listen_transcriber(self):
                         self.interruption_manager.reset_delay_for_speech_final(len(self.history))
 
                         transcriber_message = message["data"].get("content")
-                        meta_info = self.__get_updated_meta_info(meta_info)
-                        await self._handle_transcriber_output(next_task, transcriber_message, meta_info)
+                        was_eager = message["data"].get("was_eager", False)
+
+                        if was_eager and self.eager_llm_task is not None:
+                            logger.info(f"EndOfTurn follows EagerEndOfTurn - using speculative LLM")
+                            self.llm_task = self.eager_llm_task
```
san0808, line 3327: by jumping straight to llm_task = eager_llm_task, aren't we skipping everything _handle_transcriber_output normally does on a transcript? specifically:

* conversation_history.append_user (only self.history is touched on the eager path, so the canonical history loses the user turn on every adopted turn)
* is_duplicate_user dedupe
* voicemail_handler trigger
* language_detector.collect_transcript
* convert_to_request_log
* pop_unheard_responses / pop_and_merge_user when response_in_pipeline
* interruption_manager.revalidate_sequence_id (the comment in that helper warns skipping this can leave audio permanently BLOCKed)

do we want to run those side effects against the final transcriber_message before reusing the eager task?

bolna/transcriber/deepgram_transcriber.py
```diff
@@ -640,6 +759,128 @@ async def receiver(self, ws: ClientConnection):
...
+                                "content": transcript,
+                                "was_eager": self.eager_transcript_pending is not None,
+                            }
+
+                            self._reset_turn_state()
+                            self.eager_transcript_pending = None
+
+                            yield create_ws_data_packet(data, self.meta_info)
+                        else:
+                            logger.warning("Flux: EndOfTurn received with empty transcript")
```
san0808, line 872: if EagerEndOfTurn already fired with content and EndOfTurn arrives empty, the eager LLM task in task_manager never gets an adopt or cancel signal. wouldn't it keep running and emit speculative audio for a turn the user didn't actually finish? should we send a turn_resumed packet here to cancel it?

bolna/constants.py
```diff
@@ -7,6 +7,11 @@
 OPENAI_TRANSCRIBER_HEARTBEAT_INTERVAL_S = 5
 OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S = 3.0
 
+# Deepgram Flux defaults — all overridable via agent transcriber config
+DEEPGRAM_FLUX_EOT_THRESHOLD = 0.7          # confidence to declare end-of-turn
+DEEPGRAM_FLUX_EAGER_EOT_THRESHOLD = 0.5    # confidence to trigger speculative LLM early
+DEEPGRAM_FLUX_EOT_TIMEOUT_MS = 500         # max silence before forcing end-of-turn
```
san0808: PR description says eot_timeout_ms=3000 but this is 500. which one is right? 500ms forces end-of-turn much more aggressively than 3s.

bolna/transcriber/deepgram_transcriber.py
```diff
@@ -102,7 +107,24 @@ def __init__(
         self.utterance_timeout_task = None
         self.connection_error = None
 
+        # Flux model support
+        self.is_flux_model = model.startswith("flux-")
+        self.is_flux_multi = model == "flux-general-multi"
+        self.eot_threshold = kwargs.get("eot_threshold") or DEEPGRAM_FLUX_EOT_THRESHOLD
```
san0808: `kwargs.get(...) or DEFAULT` treats 0 / 0.0 as falsy. if someone configures eot_threshold=0 it'll silently get replaced. should this be `kwargs.get("eot_threshold", DEFAULT)` or an explicit `is None` check? same on line 117 for eot_timeout_ms.

bolna/transcriber/deepgram_transcriber.py
```diff
@@ -640,6 +759,128 @@ async def receiver(self, ws: ClientConnection):
...
+                                logger.info(f"Flux LID Update: languages={languages} hinted={languages_hinted}")
+
+                            data = {"type": "interim_transcript_received", "content": transcript}
+                            yield create_ws_data_packet(data, self.meta_info)
+
+                    elif event == "EagerEndOfTurn":
+                        if languages:
+                            logger.info(f"Flux LID EagerEndOfTurn: languages={languages} hinted={languages_hinted}")
+                        logger.info(f"Flux: EagerEndOfTurn (confidence={eot_confidence}, transcript={transcript})")
+                        self.eager_transcript_pending = transcript
```
san0808: should this only be set when transcript is non-empty? right now if EagerEndOfTurn arrives with empty content, this becomes `""`, then at EndOfTurn `was_eager = self.eager_transcript_pending is not None` evaluates True for the empty string and trips the adoption branch in task_manager.

bolna/constants.py
puneet-bolna: initially it was 3sec but it was too much so i thought endpointing is default at 400 so i  made this as 500 default and this is configurable in UI

bolna/agent_manager/task_manager.py
puneet-bolna: removed

bolna/agent_manager/task_manager.py
puneet-bolna: added
Snapshot len(self.history) right before appending the speculative user message. On TurnResumed, truncate history back to that snapshot — works regardless of whether the
LLM has already replied or not

bolna/agent_manager/task_manager.py
puneet-bolna: added 
Store the eager task's meta_info (which carries its sequence_id) so we can revalidate the right ID at EndOfTurn. Then run the critical side effects manually at
  adoption time.

bolna/transcriber/deepgram_transcriber.py
puneet-bolna: added
in the empty EndOfTurn branch, if eager_transcript_pending is set (i.e. an EagerEndOfTurn did fire), yield a turn_resumed packet first so task_manager  
  cancels the eager task and reverts history. turn_resumed is the right reuse here — its handler already does exactly what we need (cancel + history rollback)

bolna/transcriber/deepgram_transcriber.py
puneet-bolna: added
All three threshold parameters now use the same is None pattern — explicit, consistent, and zero-safe

bolna/transcriber/deepgram_transcriber.py
puneet-bolna: added 
guard the entire EagerEndOfTurn handling on a non-empty transcript — same principle as the task_manager's own guard

bolna/transcriber/deepgram_transcriber.py
```diff
@@ -204,6 +292,40 @@ async def send_heartbeat(self, ws: ClientConnection):
...
+                    logger.warning("Flux heartbeat ping timeout - connection may be stale")
+                    break
+                except ConnectionClosed as e:
+                    rcvd_code = getattr(e.rcvd, "code", None)
+                    sent_code = getattr(e.sent, "code", None)
+
+                    if rcvd_code == 1000 or sent_code == 1000:
+                        logger.info("WebSocket closed normally (1000 OK) during Flux heartbeat.")
+                    else:
+                        logger.warning(
```
SubhrajitPrusty: this should be an error

bolna/transcriber/deepgram_transcriber.py
```diff
@@ -204,6 +292,40 @@ async def send_heartbeat(self, ws: ClientConnection):
...
+                        )
+                    break
+                except Exception as e:
+                    logger.error(f"Error sending Flux heartbeat ping: {e}")
+                    break
+
+                await asyncio.sleep(5)  # Send ping every 5 seconds
+        except asyncio.CancelledError:
+            logger.info("Flux heartbeat task cancelled")
+            raise
```
SubhrajitPrusty: should be an error log or not raised, incase of graceful cancel

SubhrajitPrusty (approved):

PR #671, opened 2026-04-29
Feat/audio lid

web1havv:
Automatic language detection that runs parallel to ASR. When a user switches language mid-call, the active transcriber switches instantly — no manual switch_language tool call needed.

### How
SarvamLID opens a separate WebSocket to Sarvam saaras:v3 with language-code=unknown and feeds every audio chunk to it in parallel  ASR path untouched, zero added latency
Detections go through debounce (2 consecutive same-language hits) + cooldown (10s after a switch) before triggering
On confirmed switch: TranscriberPool.switch() + TaskManager.switch_language() fires — swaps transcriber, synthesizer, and system prompt
### Changes
lid_provider.py -new file, SarvamLID class + LIDProvider factory shim
transcriber_pool.py -LID tap in run()/cleanup(), _handle_lid_signal() with debounce + cooldown, _lang_to_label map from multilingual config
task_manager.py -wires lid_provider, lid_config, on_lid_switch into TranscriberPool for multilingual agents; controlled via LID_PROVIDER env var (default: sarvam)
models.py - added multilingual and active to Transcriber Pydantic model

Why models.py?
Transcriber didn't declare multilingual/active. When an agent is created via API, dashboard-backend calls model_dump() which silently strips undeclared fields before saving to Redis so the multilingual block was lost and the pool never started. These two fields fix that round-trip.

Env
LID_PROVIDER=sarvam (default, only supported value for now)

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -348,6 +348,9 @@ class TaskManager(BaseManager):
 
         # Stores structured API call records for dashboard/backend persistence.
         self.function_tool_api_call_details = []
+        # Records every manual switch_language tool call — used post-call to
+        # compare against LID shadow detections (precision / latency analysis).
+        self.language_switch_events: list[dict] = []
         self.hangup_task = None
 
         self.conversation_config = None
@@ -1041,7 +1044,11 @@ class TaskManager(BaseManager):
                         if label == active_label:
                             self.transcriber_provider = cfg.get("provider", cfg.get("model"))
 
-                    # Audio LID tap — configurable via LID_PROVIDER env var (default: sarvam).
+                    # Audio LID tap.
+                    # LID_PROVIDER — which backend to use (default: "sarvam").
+                    # LID_MODE     — "shadow" (default) logs detections without switching;
+                    #                "active" performs live transcriber/synthesizer/prompt swap.
+                    #                Keep "shadow" until detection quality is validated.
                     LID_PROVIDER = os.getenv("LID_PROVIDER", "sarvam")
                     _lid_config = {"telephony_provider": provider}
 
@@ -1056,7 +1063,8 @@ class TaskManager(BaseManager):
                         on_lid_switch=self.switch_language,
                     )
                     logger.info(
-                        f"TranscriberPool created with labels={list(transcribers.keys())}, active='{active_label}'"
+                        f"TranscriberPool created with labels={list(transcribers.keys())}, "
+                        f"active='{active_label}', lid_provider={LID_PROVIDER!r}"
                     )
                     return
 
@@ -3336,14 +3344,27 @@ class TaskManager(BaseManager):
         """Get agent name for a language label from configured agent_names."""
         return self.agent_names.get(label, "")
 
-    async def switch_language(self, label, components=None):
+    async def switch_language(self, label, components=None, triggered_by: str = "manual"):
         """Switch the active language for multilingual pools.
 
         Args:
             label: language label to switch to (e.g. "hi", "en").
             components: list of component names to switch. Defaults to both.
+            triggered_by: "manual" (LLM tool call) or "lid" (automatic detection).
+                          Used in post-call telemetry to compare LID shadow detections
+                          against actual LLM-decided switches.
         """
         components = components or ["transcriber", "synthesizer"]
+
+        # Record every switch so shadow-eval can compare LID detections vs.
+        # actual LLM-decided switches on the same call.
+        self.language_switch_events.append({
+            "to_label":       label,
+            "from_label":     self.language,
+            "triggered_by":   triggered_by,
+            "switched_at":    time.time(),
+        })
+
         if "transcriber" in components and isinstance(self.tools.get("transcriber"), TranscriberPool):
             await self.tools["transcriber"].switch(label)
         if "synthesizer" in components and isinstance(self.tools.get("synthesizer"), SynthesizerPool):
@@ -4240,6 +4261,8 @@ class TaskManager(BaseManager):
                     "conversation_time": time.time() - self.start_time,
                     "label_flow": self.label_flow,
                     "function_tool_api_call_details": copy.deepcopy(self.function_tool_api_call_details),
+                    "lid_detection_events": list(getattr(self.tools.get("transcriber"), "lid_detection_events", [])),
+                    "language_switch_events": list(self.language_switch_events),
                     "call_sid": self.call_sid,
                     "stream_sid": self.stream_sid,
                     "transcriber_duration": self.transcriber_duration,
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -1041,12 +1041,20 @@ class TaskManager(BaseManager):
                         if label == active_label:
                             self.transcriber_provider = cfg.get("provider", cfg.get("model"))
 
+                    # Audio LID tap — always Sarvam for multilingual pools.
+                    # To switch provider, change LID_PROVIDER below; agents have no control over this.
+                    LID_PROVIDER = "sarvam"
+                    _lid_config = {"telephony_provider": provider}
+
                     self.tools["transcriber"] = TranscriberPool(
                         transcribers=transcribers,
                         shared_input_queue=self.audio_queue,
                         output_queue=self.transcriber_output_queue,
                         active_label=active_label,
                         multilingual_config=multilingual,
+                        lid_provider=LID_PROVIDER,
+                        lid_config=_lid_config,
+                        on_lid_switch=self.switch_language,
                     )
                     logger.info(
                         f"TranscriberPool created with labels={list(transcribers.keys())}, active='{active_label}'"
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -1041,9 +1041,9 @@ class TaskManager(BaseManager):
                         if label == active_label:
                             self.transcriber_provider = cfg.get("provider", cfg.get("model"))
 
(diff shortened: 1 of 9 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -1041,12 +1041,19 @@ def __setup_transcriber(self):
...
+
                     self.tools["transcriber"] = TranscriberPool(
                         transcribers=transcribers,
                         shared_input_queue=self.audio_queue,
                         output_queue=self.transcriber_output_queue,
                         active_label=active_label,
                         multilingual_config=multilingual,
+                        lid_provider=LID_PROVIDER,
+                        lid_config=_lid_config,
+                        on_lid_switch=self.switch_language,
```
san0808, line 1063: For shadow-testing the model, should the default be shadow (log detections, do nothing) rather than wiring `on_lid_switch=self.switch_language`? Right now `LID_PROVIDER=sarvam` (default) actively swaps transcriber, synthesizer, and system prompt on every confirmed detection. A `LID_MODE=shadow|active` env var with `shadow` default would let us collect data first.

bolna/transcriber/transcriber_pool.py
```diff
@@ -137,6 +188,85 @@ async def _standby_keepalive(self):
...
+        """
+        # Skip confidence check for Sarvam — it doesn't return language_probability
+        # in unknown-language mode; the language_code itself is the signal.
+        if self._lid_provider_name != "sarvam" and confidence < self._LID_CONFIDENCE_THRESHOLD:
+            logger.debug(f"TranscriberPool LID: {lang} conf={confidence:.2f} below threshold, ignoring")
+            return
+
+        # Already on this language — reset debounce and do nothing
+        active_lang = (self._multilingual_config.get(self.active_label, {})
+                       .get("language_code", "")).split("-")[0].lower()
```
san0808: Inconsistency vs the map build at line 81, which falls back to `cfg.get("language")` if `language_code` is missing. This compare only checks `language_code`. If a multilingual config has only `language`, `active_lang` becomes `""` here, so `lang == active_lang` is always false and every detection bypasses the "already on this language" early-return. Should both use the same fallback?

bolna/transcriber/transcriber_pool.py
```diff
@@ -137,6 +188,85 @@ async def _standby_keepalive(self):
...
+                    f"TranscriberPool LID: switching {self.active_label} → {target_label} "
+                    f"(lang={lang}, conf={confidence:.2f})"
+                )
+                self._lid_last_switch_time = now
+                self._lid_pending_lang = None
+                self._lid_pending_count = 0
+                await self.switch(target_label)
+                if self._on_lid_switch:
+                    try:
+                        await self._on_lid_switch(target_label)
```
san0808: `self._on_lid_switch` is `TaskManager.switch_language`, which itself calls `self.tools["transcriber"].switch(label)` at task_manager.py:3348. The second call is a no-op (short-circuits at line 280-282), but between line 257 and `switch_language` updating the synthesizer + system prompt + `self.language`, there is a small window where the transcriber has flipped but the rest has not. Could we drop the explicit `self.switch()` here and let `switch_language` own the whole transition?

bolna/transcriber/lid_provider.py
```diff
@@ -0,0 +1,163 @@
...
+import asyncio
+import base64
+import io
+import json
+import logging
+import os
+import wave
+from typing import Awaitable, Callable, Optional
+
+logger = logging.getLogger(__name__)
```
san0808: The rest of the transcriber module uses `bolna.helpers.logger_config.configure_logger` (see `transcriber_pool.py:4`). Any reason this file uses stdlib `logging.getLogger` instead? Mixing the two means LID logs miss the project formatter and any context vars.

bolna/transcriber/lid_provider.py
```diff
@@ -0,0 +1,163 @@
...
+        self.on_language = on_language
+        self.config = config
+        self._api_key = config.get("sarvam_api_key") or os.getenv("SARVAM_API_KEY", "")
+        self._host = config.get("sarvam_host") or os.getenv("SARVAM_HOST", "api.sarvam.ai")
+        self._telephony = config.get("telephony_provider", "")
+        self._sr = int(config.get("sampling_rate", 16000))
+        self._input_sr = 8000 if self._telephony in ("twilio", "plivo") else self._sr
+        self._encoding = "mulaw" if self._telephony == "twilio" else "linear16"
+
+        self._queue: asyncio.Queue = asyncio.Queue()
```
san0808: Should this queue have a `maxsize`? If the Sarvam WS stalls or slows down, audio chunks would buffer here for the full call duration with no backpressure. LID is best-effort, so bounding it (say, a few hundred packets) and dropping on full feels safer than unbounded growth.

bolna/transcriber/lid_provider.py
```diff
@@ -0,0 +1,163 @@
...
+                    if data.get("type") == "data":
+                        payload = data.get("data", {})
+                        lang = payload.get("language_code", "")
+                        # Sarvam returns language_probability=None in unknown mode;
+                        # the language_code itself is the detection signal.
+                        conf = float(payload.get("language_probability") or 0.0)
+                        if lang and lang != "unknown":
+                            short = lang.split("-")[0].lower()
+                            logger.info(f"SarvamLID: detected {lang!r} (short={short!r})")
+                            await self.on_language(short, conf)
```
san0808, line 148: Per the comment above, Sarvam returns `language_probability=None` in unknown mode, so `conf` is always 0.0 here. Combined with the `self._lid_provider_name != "sarvam"` skip in `_handle_lid_signal`, the `_LID_CONFIDENCE_THRESHOLD` gate is effectively dead code for the only supported provider. For shadow eval it would also be useful to have real confidence numbers. Is there another field in the payload we can pull a score from, or should we drop the confidence param?

bolna/transcriber/lid_provider.py
```diff
@@ -0,0 +1,163 @@
...
+                except Exception as e:
+                    logger.error(f"SarvamLID receiver parse error: {e}")
+        except asyncio.CancelledError:
+            pass
+        except Exception as e:
+            logger.error(f"SarvamLID receiver error: {e}")
```
san0808, line 154: If the WS drops mid-call, both `_sender_loop` and `_receiver_loop` catch, log, and exit. There is no reconnect, so LID is dead for the rest of the call after a single transient hiccup, with no signal to the caller. For shadow eval this will silently bias the stats (calls where LID died early will look like "no detection"). Should we add a reconnect-with-backoff loop, or at least surface a "LID dead" flag the pool can act on?

bolna/agent_manager/task_manager.py
web1havv: Added LID_MODE=shadow|active env var, default shadow. In shadow mode, detections are logged with suppressed_reason=shadow_mode and no switch is ever called. Verified on staging — calls run fully in shadow with DB rows confirming would_switch=false.

bolna/transcriber/transcriber_pool.py
web1havv: handle_lid_signal now uses cfg.get("language_code") or cfg.get("language") — same fallback as the _lang_to_label map build at __init__. The "already on this language" early-return now works correctly for configs that only have language.

bolna/transcriber/transcriber_pool.py
web1havv: Removed the explicit self.switch(target_label) call from _handle_lid_signal. In active mode, we now only call on_lid_switch(target_label, triggered_by="lid") which is TaskManager.switch_language

bolna/transcriber/lid_provider.py
web1havv: Fixed

bolna/transcriber/lid_provider.py
web1havv: Queue is now asyncio.Queue(maxsize=200). On QueueFull, chunks are dropped with a debug log (backpressure).

bolna/transcriber/lid_provider.py
web1havv: Added a comment in _receiver_loop explaining this. The confidence gate in _handle_lid_signal is intentionally skipped for Sarvam (if self._lid_provider_name != "sarvam"), kept for future providers.

bolna/transcriber/transcriber_pool.py
```diff
@@ -137,6 +201,152 @@ async def _standby_keepalive(self):
...
+                f"(lang={lang}, conf={confidence:.2f}, suppressed_reason=shadow_mode)"
+            )
+            self._record_lid_event(
+                detected_lang=lang,
+                confidence=confidence,
+                target_label=target_label,
+                would_switch=False,
+                suppressed_reason="shadow_mode",
+            )
+            return
```
san0808, line 329: Should `self._lid_last_switch_time = now` be set here too, before the `return`? Right now `_lid_last_switch_time` is only updated in the active branch at line 336, so cooldown never engages in shadow mode. In a multilingual call where the user is steadily speaking the target language, Sarvam keeps emitting that detection, the debounce keeps re-arming (since `active_label` never changes in shadow), and `_record_lid_event` fires once every 2 detections. A multi-minute Hindi monologue will produce dozens of duplicate "would_switch" events instead of one, inflating detections_total, FP rate, and agreement metrics in the eval data. One-line fix: set `self._lid_last_switch_time = now` in the shadow branch as well.

web1havv: done now instead of firing every 2 detections for a steady monologue, shadow mode respects the 10s cooldown

san0808 (approved):

PR #686, opened 2026-05-05
Feat/azure lid

web1havv:
Adds AzureLID as default natively handles 8kHz mulaw, no upsampling, uses existing Azure credentials
Drops detections for languages outside the agent's pool; uses utterance duration as confidence since Azure doesn't expose scores
Adds lid_provider and detected_at to telemetry events for per-provider tracking


bolna/transcriber/transcriber_pool.py
```diff
@@ -212,7 +212,7 @@ async def _standby_keepalive(self):
     async def _start_lid_tap(self) -> None:
         """Instantiate and connect the configured LID provider."""
         try:
-            from bolna.transcriber.lid_provider import LIDProvider
+            from bolna.lid import LIDProvider
```
san0808: move the import to the top

san0808 (approved):

PR #689, opened 2026-05-06
Feat/function call manual trigger

puneet-bolna:


bolna/llms/openai_llm.py
```diff
@@ -333,16 +336,60 @@ async def _generate_stream_chat(
...
+                if text_tool_buffer is not None:
+                    text_tool_buffer += content
+                    end_pos = self._find_tool_call_end(text_tool_buffer)
+                    if end_pos != -1:
+                        captured_tool_text = text_tool_buffer[:end_pos]
+                        remainder = text_tool_buffer[end_pos:].lstrip("\n")
+                        text_tool_buffer = None
+                        answer += remainder
+                        buffer = remainder
+                elif "functions." in (buffer + content):
```
san0808: entry condition is just `"functions." in (buffer + content)`, but `_find_tool_call_end` requires the full `functions\.\w+\s*[({]` shape. if the assistant says something like "i support several functions. they include..." in prose, this latches and holds back everything after that token until end-of-stream (the tail never gets streamed to TTS in real time).

repro in a real call: give the agent a system prompt that nudges it to talk about its tools out loud, e.g. "if asked what you can do, list the functions you support." then ask "what can you do?" — odds are good the response contains the literal word "functions." mid-sentence and you'll hear a noticeable pause / tail dump.

should we tighten the entry to the same regex `_find_tool_call_end` uses, so prose can't latch?

bolna/llms/openai_llm.py
```diff
@@ -333,16 +336,60 @@ async def _generate_stream_chat(
...
+                    idx = combined.find("functions.")
+                    before = combined[:idx]
+                    after = combined[idx:]
+                    if before:
+                        answer += before
+                        if synthesize:
+                            yield LLMStreamChunk(data=before, end_of_stream=False, latency=latency_data)
+                    end_pos = self._find_tool_call_end(after)
+                    if end_pos != -1:
+                        captured_tool_text = after[:end_pos]
```
san0808, line 370: if a single stream contains two `functions.foo(...)` calls back-to-back, this line overwrites the previous `captured_tool_text` and only the last one is rescued at end-of-stream. the first call gets silently dropped (no log, no TTS fallback). verified locally with chunks like `['functions.foo({"a":1})\n', 'functions.bar({"b":2})']`.

should we loop / yield per detected call instead of single-shotting at end-of-stream? at minimum a warning when overwriting would make this debuggable.

bolna/llms/openai_base.py
```diff
@@ -22,6 +29,128 @@ class OpenAICompatibleLLM(BaseLLM):
...
+                    "function": {"name": func_name, "arguments": args_str},
+                    "type": "function",
+                }
+            ],
+            tool_call_id=f"rescued_{func_name}",
+            textual_response=_clean_rescue_answer(answer),
+        )
+
+        try:
+            parsed_args = json.loads(args_str)
```
san0808, line 148: rescue path applies all parsed args via setattr without checking required keys against the tool spec. `ToolCallAccumulator.build_api_payload` does that check (`tool_call_accumulator.py` line ~106) and zeroes `resp` if any required key is missing. so the rescue can fire a call with missing required args while the structured path can't.

should we mirror the required-key check here for parity?

bolna/llms/openai_llm.py
puneet-bolna: can this happen in real scenario two function call back to back in same llm response.  i can add some logging here to identify

bolna/llms/openai_llm.py
puneet-bolna: here i have tighten more 
now it will check after functions. we still have letters it is not a text which llm spoke

bolna/llms/openai_base.py
puneet-bolna: added the  validation before firing the api

san0808 (approved):

PR #690, opened 2026-05-06
feat: add end-to-end observability timestamps for call progression

puneet-bolna:
Captures user_start_ms, asr_start/finalized_ms, llm_start_ms, tts_start_ms, agent_end_ms, call_start_epoch_ms and turn_id across transcriber, LLM, and synthesizer so the progression endpoint can reconstruct the full call timeline.

### DIFF
--- a/bolna/agent_manager/interruption_manager.py
+++ b/bolna/agent_manager/interruption_manager.py
@@ -60,6 +60,7 @@ class InterruptionManager:
 
         self._adjusted_user_stop_ts: Optional[float] = None
         self._last_user_start_ts: float = -1
+        self._pending_user_start_ts: float = -1  # snapshot of user_start paired with _adjusted_user_stop_ts
         self.user_bot_latencies: List[Dict] = []
 
         logger.info(
@@ -182,6 +183,7 @@ class InterruptionManager:
                 self._adjusted_user_stop_ts = user_stop_ts_wall
             else:
                 self._adjusted_user_stop_ts = now_s - (stop_offset_ms / 1000.0)
+            self._pending_user_start_ts = self._last_user_start_ts
 
         if self.callee_speaking_start_time > 0:
             self._total_user_speaking_ms += (now_s - self.callee_speaking_start_time) * 1000
@@ -258,13 +260,14 @@ class InterruptionManager:
             self.user_bot_latencies.append(
                 {
                     "sequence_id": sequence_id,
-                    "user_start_s": self._last_user_start_ts if self._last_user_start_ts > 0 else None,
+                    "user_start_s": self._pending_user_start_ts if self._pending_user_start_ts > 0 else None,
                     "user_end_s": self._adjusted_user_stop_ts,
                     "agent_start_s": now_s,
                     "latency_ms": round(latency_ms, 2),
                 }
             )
             self._adjusted_user_stop_ts = None
+            self._pending_user_start_ts = -1
 
         logger.info(f"Agent speech started (sequence_id={sequence_id})")
 
--- a/bolna/agent_manager/interruption_manager.py
+++ b/bolna/agent_manager/interruption_manager.py
@@ -59,6 +59,7 @@ class InterruptionManager:
         self.longest_agent_monologue_ms: float = 0.0
 
         self._adjusted_user_stop_ts: Optional[float] = None
+        self._last_user_start_ts: float = -1
         self.user_bot_latencies: List[Dict] = []
 
         logger.info(
@@ -145,6 +146,7 @@ class InterruptionManager:
         if not self.callee_speaking:
             self.callee_speaking = True
             self.callee_speaking_start_time = time.time()
+            self._last_user_start_ts = self.callee_speaking_start_time
             logger.info("User started speaking")
 
     def on_interim_transcript_received(self) -> None:
@@ -203,6 +205,7 @@ class InterruptionManager:
 
         event: Dict = {
             "type": "user_interrupted_agent",
+            "turn_id": self.turn_id,
             "user_start_s": self.callee_speaking_start_time if self.callee_speaking_start_time > 0 else None,
             "user_end_s": None,
             "recovery_completed": False,
@@ -224,6 +227,7 @@ class InterruptionManager:
 
         event: Dict = {
             "type": "agent_interrupted_user",
+            "turn_id": self.turn_id,
             "user_start_s": self.callee_speaking_start_time if self.callee_speaking_start_time > 0 else None,
             "user_end_s": None,
             "recovery_completed": False,
@@ -254,6 +258,7 @@ class InterruptionManager:
             self.user_bot_latencies.append(
                 {
                     "sequence_id": sequence_id,
+                    "user_start_s": self._last_user_start_ts if self._last_user_start_ts > 0 else None,
                     "user_end_s": self._adjusted_user_stop_ts,
                     "agent_start_s": now_s,
                     "latency_ms": round(latency_ms, 2),
--- a/bolna/agent_manager/interruption_manager.py
+++ b/bolna/agent_manager/interruption_manager.py
@@ -260,7 +260,8 @@ class InterruptionManager:
             self.user_bot_latencies.append(
                 {
                     "sequence_id": sequence_id,
-                    "user_start_s": self._pending_user_start_ts if self._pending_user_start_ts > 0 else None,
+                    "user_start_s": self._last_user_start_ts if self._last_user_start_ts > 0 else None,
+                    "user_first_start_s": self._pending_user_start_ts if self._pending_user_start_ts > 0 else None,
                     "user_end_s": self._adjusted_user_stop_ts,
                     "agent_start_s": now_s,
                     "latency_ms": round(latency_ms, 2),
--- a/bolna/agent_manager/interruption_manager.py
+++ b/bolna/agent_manager/interruption_manager.py
@@ -269,6 +269,20 @@ class InterruptionManager:
             )
             self._adjusted_user_stop_ts = None
             self._pending_user_start_ts = -1
+        else:
+            # Tool-call follow-up or any agent response without a preceding user turn
+            # in this sequence (e.g. seq=5 after a tool call consumed seq=4's user timing).
+            # Record agent_start_s so agent_speech_start appears in the progression.
+            self.user_bot_latencies.append(
+                {
+                    "sequence_id": sequence_id,
+                    "user_start_s": None,
+                    "user_first_start_s": None,
+                    "user_end_s": None,
+                    "agent_start_s": now_s,
+                    "latency_ms": None,
+                }
+            )
 
         logger.info(f"Agent speech started (sequence_id={sequence_id})")
 
--- a/bolna/agent_manager/interruption_manager.py
+++ b/bolna/agent_manager/interruption_manager.py
@@ -220,7 +220,7 @@ class InterruptionManager:
             f"Interruption triggered — turn_id={self.turn_id}, user_interrupted_agent_count={self.user_interrupted_agent_count}"
         )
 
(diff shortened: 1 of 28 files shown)


bolna/transcriber/deepgram_transcriber.py
```diff
@@ -684,16 +705,23 @@ async def receiver(self, ws: ClientConnection):
...
                                         "turn_id": self.current_turn_id,
-                                        "sequence_id": self.current_turn_id,
+                                        "asr_start_epoch_ms": self.speech_start_time,
+                                        "asr_turn_start_epoch_ms": self._turn_first_speech_epoch_ms,
+                                        "final_transcript": self.final_transcript,
+                                        "asr_finalized_epoch_ms": timestamp_ms(),
                                         "interim_details": self.current_turn_interim_details,
                                         "first_interim_to_final_ms": first_interim_to_final_ms,
                                         "last_interim_to_final_ms": last_interim_to_final_ms,
+                                        "asr_start_epoch_ms": self.current_turn_start_time,
```
san0808: duplicate dict keys here (and at L373 / L772). `asr_start_epoch_ms`, `asr_finalized_epoch_ms`, `final_transcript` are set twice in the same literal, so Python silently keeps the second. that makes the `self.speech_start_time` line above dead code. was that intentional, or should we pick one source per field?

bolna/transcriber/deepgram_transcriber.py
```diff
@@ -705,9 +733,7 @@ async def receiver(self, ws: ClientConnection):
                                 )
                                 pass
                             self.meta_info["user_stop_offset_ms"] = self.endpointing_ms
-                            last_word_end_wall = self._compute_last_word_end_wall(msg)
-                            if last_word_end_wall is not None:
-                                self.meta_info["user_stop_ts_wall"] = last_word_end_wall
+                            self.meta_info["user_stop_ts_wall"] = self._compute_last_word_end_wall(msg)
```
san0808, line 733: we dropped the `if last_word_end_wall is not None` guard. `_compute_last_word_end_wall` returns None on empty `words` or any exception. was the intent to allow it to clobber a previously valid `user_stop_ts_wall` with None?

bolna/agent_manager/task_manager.py
```diff
@@ -4869,6 +5084,107 @@ async def run(self):
...
+                        _t.pop(_f, None)
+
+                _asr_turns = (output["latency_dict"]["transcriber_latencies"] or {}).get("turn_latencies", [])
+                for _t in _asr_turns:
+                    for _f in (
+                        "asr_start_ms",
+                        "asr_finalized_ms",
+                        "asr_turn_start_ms",
+                        "final_transcript",
+                        "was_interrupted",
```
san0808: `was_interrupted` was on master `latency_dict.transcriber_latencies.turn_latencies`. stripping it here looks like a regression for any current consumer reading it from `latency_dict`. should we keep it on both, or only strip the genuinely PR-new fields?

bolna/agent_manager/task_manager.py
```diff
@@ -3239,6 +3394,8 @@ async def process_call_hangup(self):
                 "end_of_llm_stream": True,
             }
             await self._synthesize(create_ws_data_packet(message, meta_info=meta_info))
+            # Stamp after goodbye is queued — actual disconnect happens after it plays
+            self.hangup_triggered_at = time.time()
```
san0808, line 3417: the docstring at L3374 says `hangup_triggered_at` is set AFTER the goodbye plays, but `_synthesize` only queues to the synth pipeline and returns; this stamps at queue-time, not playback-end. should we wait on the final-chunk ACK for the goodbye sequence, or update the comment to say "queued time"?

bolna/transcriber/deepgram_transcriber.py
puneet-bolna: this we did because lets say it returns None in current scenario then the user stop ts wall will still have the previous turn data but after this it will clear it.

The guard in interruption_manager (line 182) will handle the None gracefully there 

code 

            if user_stop_ts_wall is not None:
                self._adjusted_user_stop_ts = user_stop_ts_wall
            else:
                self._adjusted_user_stop_ts = now_s - (stop_offset_ms / 1000.0)
            self._pending_user_start_ts = self._last_user_start_ts

bolna/agent_manager/task_manager.py
puneet-bolna: my bad will add it back

bolna/transcriber/deepgram_transcriber.py
puneet-bolna: my bad will remove it

bolna/agent_manager/task_manager.py
puneet-bolna: changed it to hangup_queued in progression endpoint

bolna/agent_manager/task_manager.py
```diff
@@ -4798,24 +4991,46 @@ async def run(self):
...
+                        "user_end_ms": round(e["user_end_s"] * 1000 - _call_start_ms, 2)
+                        if e.get("user_end_s") is not None
+                        else None,
                         "agent_start_ms": round(e["agent_start_s"] * 1000 - _call_start_ms, 2),
+                        "agent_end_ms": self._agent_end_timestamps.get(e["sequence_id"]),
                         "latency_ms": e["latency_ms"],
+                        # agent_end_ms: actual playback end from provider mark ACK (more accurate
+                        # than ElevenLabs stream end). Present only when the final mark was ACKed.
+                        **(
+                            {"agent_end_ms": round(e["agent_end_s"] * 1000 - _call_start_ms, 2)}
```
san0808: `agent_end_ms` is set twice in the same literal (L5025 from `_agent_end_timestamps`, L5030 from `agent_end_s` via spread). it happens to do the right thing (mark-ACK wins, synth-stream-end falls back) but precedence is implicit. could we either make it an explicit conditional, or use distinct field names so consumers can tell which source?

bolna/agent_manager/task_manager.py
```diff
@@ -4869,6 +5085,106 @@ async def run(self):
...
+                for _ub in output["progression_data"]["user_bot_latencies"]:
+                    _seq = _ub.get("sequence_id")
+                    if _seq is not None and _seq in _seq_to_asr_turn:
+                        _ub["turn_id"] = _seq_to_asr_turn[_seq]
+
+                for _tts_t in output["progression_data"]["synthesizer_latencies"].get("turn_latencies", []):
+                    _seq = _tts_t.get("sequence_id")
+                    if _seq is not None and _seq in _seq_to_asr_turn:
+                        _tts_t["turn_id"] = _seq_to_asr_turn[_seq]
+
```
san0808, line 5176: this enrich-then-strip pattern is brittle: every new field has to be added to the strip list or it leaks into `latency_dict`. e.g. `characters` (stamped on synth turns in `stream_synthesizer._record_turn_latency`) and the changed `sequence_id` semantics on synth turns aren't stripped. would it be cleaner to snapshot `latency_dict` at the top, then write all PR-new fields only onto `progression_data` and leave the original untouched?

bolna/agent_manager/task_manager.py
```diff
@@ -4798,24 +4992,46 @@ async def run(self):
                 # Annotate each transcriber turn with was_interrupted so callers
                 # can see which ASR turns had a user barge-in without cross-referencing
                 # the separate interruption_events list.
+                _call_start_ms = self.conversation_start_init_ts
+
                 _interrupted_ids = self.interruption_manager.interrupted_transcriber_turn_ids
                 for _turn in self.transcriber_latencies.turn_latencies:
-                    _tid = _turn.get("turn_id") or _turn.get("sequence_id")
+                    _tid = _turn.get("turn_id")
```
san0808, line 5029: we dropped the `or _turn.get("sequence_id")` fallback. any transcriber that historically wrote only `sequence_id` (no `turn_id`) will now miss the `was_interrupted` annotation. did you verify all transcribers in the pool always set `turn_id` post-PR?

bolna/agent_manager/task_manager.py
```diff
@@ -3033,6 +3158,21 @@ async def __do_llm_generation(
...
+                    )
+                    _t = self.tools.get("transcriber")
+                    if hasattr(_t, "transcribers") and hasattr(_t, "active_label"):
+                        _t = _t.transcribers.get(_t.active_label, _t)
+                    latency_dict["asr_turn_id"] = getattr(_t, "turn_counter", None)
+                    latency_dict["model"] = self.llm_config.get("model") if self.llm_config else None
+                    latency_dict["input_tokens"] = actual_input_tokens
+                    latency_dict["output_tokens"] = actual_output_tokens
+                    latency_dict["reasoning_tokens"] = actual_reasoning_tokens
+                    latency_dict["cached_tokens"] = actual_cached_tokens
```
san0808: this 9-field stamp (`turn_id`, `llm_start_ms`, `asr_turn_id`, `model`, the four token counts, plus `response_text`) is duplicated almost verbatim in the function-call branch at L3118-3136. could we extract a `_stamp_llm_turn(latency_dict, meta_info, tokens, response_text)` helper so the two paths can't drift?

bolna/agent_manager/interruption_manager.py
```diff
@@ -254,19 +261,50 @@ def on_agent_speech_started(self, sequence_id: int) -> None:
...
+            # Tool-call follow-up or any agent response without a preceding user turn
+            # in this sequence (e.g. seq=5 after a tool call consumed seq=4's user timing).
+            # Record agent_start_s so agent_speech_start appears in the progression.
+            self.user_bot_latencies.append(
+                {
+                    "sequence_id": sequence_id,
+                    "user_start_s": None,
+                    "user_first_start_s": None,
+                    "user_end_s": None,
+                    "agent_start_s": now_s,
```
san0808, line 284: the comment says "tool-call follow-up" but the predicate is just `_adjusted_user_stop_ts is None`, which fires for any agent speech with no preceding `on_user_speech_ended` (e.g. back-to-back agent messages, or a timing race where ended fired after started). risk of spurious agent-only entries. should we gate this on a tool-call-follow-up signal rather than the absence of user-stop?

bolna/agent_manager/interruption_manager.py
```diff
@@ -224,6 +230,7 @@ def on_agent_interrupted_user(self) -> None:
 
         event: Dict = {
             "type": "agent_interrupted_user",
```
san0808, line 232: `on_interruption_triggered` now takes `asr_turn_id` and writes it on the event, but `on_agent_interrupted_user` (this fn) doesn't. is the asymmetry intentional, or should this one also be tagged so the progression service can place both interruption types in the right turn lane?

bolna/agent_manager/task_manager.py
puneet-bolna: All transcribers set both turn_id and sequence_id to the same value (self.current_turn_id). Dropping the or _turn.get("sequence_id") fallback has no runtime impact today.

bolna/agent_manager/task_manager.py
puneet-bolna: discussed.

bolna/agent_manager/task_manager.py
puneet-bolna: refactored

bolna/agent_manager/interruption_manager.py
puneet-bolna: discussed

bolna/agent_manager/interruption_manager.py
puneet-bolna: added it

bolna/agent_manager/task_manager.py
puneet-bolna: this removed

san0808 (approved):

PR #699, opened 2026-05-09
feat: separate DEEPGRAM_FLUX_HOST for self-hosted Flux endpoint

puneet-bolna:
Flux runs on a different cluster than Nova on India self-hosted, so the Flux WebSocket URL needs its own host/protocol independent of DEEPGRAM_HOST. Read DEEPGRAM_FLUX_HOST and DEEPGRAM_FLUX_HOST_PROTOCOL in _get_flux_ws_url (both default to api.deepgram.com / wss, matching prior behavior). Nova URL builder is untouched.

### DIFF
--- a/bolna/transcriber/deepgram_transcriber.py
+++ b/bolna/transcriber/deepgram_transcriber.py
@@ -245,9 +245,7 @@ class DeepgramTranscriber(BaseTranscriber):
         if self.run_id:
             dg_params["tag"] = self.run_id
 
-        websocket_api = "{}://{}/v2/listen?".format(
-            os.getenv("DEEPGRAM_HOST_PROTOCOL", "wss"), self.deepgram_flux_host
-        )
+        websocket_api = "{}://{}/v2/listen?".format(os.getenv("DEEPGRAM_HOST_PROTOCOL", "wss"), self.deepgram_flux_host)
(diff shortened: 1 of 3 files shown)


bolna/transcriber/deepgram_transcriber.py
```diff
@@ -244,7 +245,9 @@ def _get_flux_ws_url(self):
         if self.run_id:
             dg_params["tag"] = self.run_id
 
-        websocket_api = "{}://{}/v2/listen?".format(os.getenv("DEEPGRAM_HOST_PROTOCOL", "wss"), self.deepgram_host)
```
san0808, line 247: isn't this going to affect the nova-3 flows ?

san0808: why is this needed ? DEEPGRAM_HOST_PROTOCOL ka hi same hoga na for flux

puneet-bolna: isn't this going to affect the nova-3 flows ? - nahi it is different nova3 listens on v1 not v2 and i think that initilisation his happening around lin 180.

yes protocol will be same removed a new variable

san0808: got it just saw the get fc is diff for both

san0808 (approved):

PR #702, opened 2026-05-09
fix: avoid premature end_of_synthesizer_stream on empty alignment chunks

san0808:
**What was happening**

The ElevenLabs receiver decides "stream is ending" by checking whether the current text ends with the last few characters of the chunk just received. ElevenLabs occasionally sends mid-stream chunks where `alignment.chars` is empty (silence padding, breath frames). When that happens, the check becomes `text.endswith("")` which is always True, so we falsely declare end-of-stream. Subsequent legitimate audio chunks for the same response then get dropped as `Skipping message with sequence_id`.

User-visible effect: agent's reply gets cut off mid-sentence even though the LLM produced the full text and ElevenLabs was still streaming the tail.

**The fix**

Only treat the chunk as a possible end-of-stream signal if its characters are non-empty. Real end-of-stream is unaffected because the explicit `isFinal: True` branch above handles it directly, and the legitimate final chunk has non-empty alignment.

### DIFF
--- a/bolna/synthesizer/elevenlabs_synthesizer.py
+++ b/bolna/synthesizer/elevenlabs_synthesizer.py
@@ -231,9 +231,8 @@ class ElevenlabsSynthesizer(StreamSynthesizer):
                         if last_four and current_norm.endswith(last_four):
                             logger.info("send end_of_synthesizer_stream")
                             yield b"\x00", ""
-                        elif (
+                        elif last_four_no_space and current_norm.replace('"', "").replace(" ", "").strip().endswith(
                             last_four_no_space
-                            and current_norm.replace('"', "").replace(" ", "").strip().endswith(last_four_no_space)
                         ):
                             logger.info("send end_of_synthesizer_stream on fallback")
                             yield b"\x00", ""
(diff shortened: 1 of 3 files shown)


bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -227,11 +227,13 @@ async def receiver(self):
                         current_norm = self.normalize_text(self.current_text.strip()).replace('"', "").strip()
                         logger.info(f"Last four char - {last_four} | current text - {current_norm}")
 
-                        if current_norm.endswith(last_four):
+                        last_four_no_space = last_four.replace(" ", "")
```
prateeksachan, line 230: won't this replace all the last 4 words space?

san0808: Yes, intentionally, and only for the fallback branch. The primary branch (line 231) still uses last_four with spaces preserved via current_norm.endswith(last_four). last_four_no_space is the whitespace-stripped variant that the original fallback was already computing inline as last_four.replace(' ', ''); I just hoisted it to a named variable so we can reuse it for both the truthy guard and the endswith check. Behavior is unchanged for any non-empty last_four.

prateeksachan (approved):

PR #709, opened 2026-05-13
fix: make welcome message non-interruptible regardless of discard_pre_welcome_utterance

san0808:


bolna/agent_manager/task_manager.py
```diff
@@ -3391,9 +3391,7 @@ async def _handle_transcriber_output(self, next_task, transcriber_message, meta_
             len((transcriber_message or "").strip()),
             (transcriber_message or "")[:120],
         )
-        if not self.tools["input"].welcome_message_played() and (
-            self.discard_pre_welcome_utterance or len(self.conversation_history) > 2
-        ):
+        if not self.tools["input"].welcome_message_played():
```
puneet-bolna, line 3394: question: welcome messaged will never get interrupted ?
whatever the user says during this will get discarded ?

san0808: Yes to both. During welcome playback, any final transcript hits the early-return and gets retired with reason welcome_still_playing. No cleanup runs, so output.handle_interruption() is never called and welcome audio cannot be cut.

The during-welcome behavior is the same as discard_pre_welcome_utterance=True. The toggle still has its own purpose for the post-welcome window: it controls whether speech that started during welcome but 
finalizes after welcome ACK is also dropped (via the _speech_started_before_welcome flag a few lines below). That path is unchanged by this PR.

pavankapoor31 (approved):

bolna/agent_manager/task_manager.py
puneet-bolna: got it

puneet-bolna (approved):

PR #722, opened 2026-05-19
fix: graph agent forced tool_choice + Responses API shape

san0808:
## Summary

Two graph-agent forced-function-call defects, one chokepoint each.

1. **Responses API tool_choice shape (PRO-87, PRO-88).** Graph nodes with `function_call: <name>` produce a Chat-Completions shape `{"type":"function","function":{"name":...}}`, but the call routes through the Responses API (any GPT-5 model + the auto-route trunk patch), which rejects that shape with `unknown_parameter: tool_choice.function`. Translation now lives in `MessageFormatAdapter.chat_tool_choice_to_responses`, called from `_build_responses_create_kwargs` so both WS and HTTP SSE Responses paths are covered.
2. **Silent loop when forced-function args can't be filled (PRO-91).** `_get_tool_choice_for_node()` was pinning a forced tool even when the node prompt referenced `{var}` tokens that weren't present in `recipient_data` (e.g. test calls with no recipient context). LLM then hallucinated values, the API call failed, the stream stalled, conversation silently looped on the node for the full timeout window. New `_missing_forced_function_vars` check drops the force to `None` (auto) when required prompt vars are unresolvable, letting the LLM speak naturally from the node prompt instead. Per-turn cost is a small dict walk against state we already have.

Repro and regression cases verified locally against the patched source: empty `recipient_data` and partially populated cases drop the force; fully resolved cases still pin; nodes without `function_call` and LLMs with `trigger_function_call=False` are unaffected.

Closes PRO-87, PRO-88, PRO-91. Related: PRO-86 (already mitigated in trunk), PRO-92 (separate semantic question in Product Solutioning).

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.48"
+__version__ = "0.10.49"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_types/graph_agent.py
+++ b/bolna/agent_types/graph_agent.py
@@ -1,6 +1,7 @@
 import asyncio
 from collections import defaultdict
 import os
+import re
 import time
 from openai import OpenAI
 from dotenv import load_dotenv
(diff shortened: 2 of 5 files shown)


san0808:
Closing in favour of a fresh PR on `fix/graph-agent-fixes` (branch renamed).

PR #728, opened 2026-05-21
capturing start and cancel timestamp in case we dont get end

puneet-bolna:


bolna/synthesizer/stream_synthesizer.py
```diff
@@ -327,6 +334,18 @@ async def cleanup(self):
...
+                _cancelled_at = (
+                    round(time.time() * 1000 - _t.conversation_start_init_ts, 2)
+                    if _t is not None else None
+                )
+                self._upsert_turn_latency({
+                    "turn_id": self.current_turn_id,
+                    "sequence_id": self.current_sequence_id,
+                    "tts_start_ms": self.current_tts_start_ms,
+                    "cancelled_at_ms": _cancelled_at,
+                })
```
san0808, line 348: Guard this with `if self.current_turn_id is not None:` like the deepgram cancel block does? After a clean turn, current_turn_id/sequence_id are None — without the guard we stamp a null-id entry, and since upsert matches on sequence_id, multiple null-seq stubs overwrite each other.

puneet-bolna: discussed

san0808 (approved):

PR #729, opened 2026-05-21
force hangup if elevan labs doesnt generate audio

puneet-bolna:


pavankapoor31 (approved):

bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -68,6 +68,7 @@ def __init__(
 
         self.context_id = None
         self.ws_send_time = None
+        self.ws_recv_time = None
```
san0808: Quick question on `ws_recv_time` lifecycle: I don't see it reset anywhere between turns. `_record_turn_latency` in `stream_synthesizer.py:280` resets `ws_send_time` after each turn's EOS, but `ws_recv_time` isn't added there. Once any turn produces audio (welcome via EL, or any earlier turn in the call), `ws_recv_time` stays set for the rest of the call.

Doesn't that mean the watchdog check `if self.tools["synthesizer"].ws_recv_time is not None: return` will return without firing for any call that had prior successful audio? The common wedge case is "call worked for a few turns, then EL wedges at goodbye", and the watchdog would skip those silently because `ws_recv_time` is leftover from an earlier turn.

Should we add `self.ws_recv_time = None` to `_record_turn_latency` (or move the attribute to the base class and reset it there)?

bolna/agent_manager/task_manager.py
```diff
@@ -3430,8 +3430,35 @@ async def process_call_hangup(self):
...
+    async def _hangup_audio_delivery_timeout(self):
+        """ElevenLabs only: force end of conversation if no audio is generated within 5s of
+        the hangup text being sent. Prevents call hanging open until call_terminate when the
+        ElevenLabs WS is in a reconnect loop."""
+        # Poll up to 3s for sender to get a connection and send the text
+        deadline = time.perf_counter() + 3
+        while time.perf_counter() < deadline:
+            if self.tools["synthesizer"].ws_send_time is not None:
+                break
+            await asyncio.sleep(0.1)
```
san0808, line 3447: If the 3s poll falls through (i.e. `ws_send_time` was never set because text was never sent due to a wedged WS), do we still need the 5s wait afterward? Audio can't arrive if text didn't go out. Could we short-circuit and force-hangup immediately when the poll times out without success? That cuts the "wedged from connection 1" recovery time from 8s to 3s.

puneet-bolna: done

PR #740, opened 2026-05-23
Lid detectection

puneet-bolna:


bolna/transcriber/transcriber_pool.py
```diff
@@ -274,7 +274,7 @@ async def _handle_lid_signal(self, lang: str, confidence: float) -> None:
         to avoid the double-switch race where the transcriber flips before the
         synthesizer and prompt catch up.
         """
-        if confidence < self._LID_CONFIDENCE_THRESHOLD:
+        if confidence is not None and confidence < self._LID_CONFIDENCE_THRESHOLD:
```
san0808, line 277: This guard catches `None` correctly, but the rest of `_handle_lid_signal` still does `f"... conf={confidence:.2f} ..."` in four places (main detection log, debounce log, shadow-mode log, active-mode log). Scribe always calls back with `confidence=None` (hard-coded in `elevenlabs_scribe.py`), so any detection on a supported language passes the threshold check and then hits `f"{None:.2f}"` → `TypeError`. Was the intent to handle `None` only at the threshold check, or should the other format calls become conditional (`f"{confidence:.2f}" if confidence is not None else "n/a"`)?

bolna/lid/elevenlabs_scribe.py
```diff
@@ -0,0 +1,223 @@
...
+                        text = data.get("text", "").strip()
+                        lang = data.get("language_code") or ""
+                        logger.info(
+                            f"ElevenLabsScribeLID: committed transcript text={text!r} language_code={lang!r}"
+                        )
+                        if lang:
+                            # Normalise to short ISO-639-1 (e.g. "eng" → "en", "hin" → "hi")
+                            short = lang[:2].lower()
+                            logger.info(f"ElevenLabsScribeLID: detected {lang!r} (short={short!r})")
+                            asyncio.create_task(self.on_language(short, None))
```
san0808: `asyncio.create_task` here means any exception inside `on_language` is swallowed and only logged at task-GC time. Combined with the `:.2f` formatting on `None` in `_handle_lid_signal`, every Scribe detection on a supported language would crash silently and zero events would land in `lid_shadow_events.asr_lid_events` — while the call looks perfectly healthy. In shadow mode thats especially bad since theres no operator-visible symptom; wed ship Scribe, look at the table a week later, and find nothing.

Since detections fire at most ~every 500ms and shadow-mode `_handle_lid_signal` does no I/O (just a dict append), would `await self.on_language(short, None)` work better? Bonus: serializes events so the pools debounce counting stays consistent.

bolna/transcriber/transcriber_pool.py
puneet-bolna: done

bolna/lid/elevenlabs_scribe.py
puneet-bolna: done

bolna/transcriber/transcriber_pool.py
san0808: Looks like the fix went in on the three logs in the debounce/shadow/active branches (lines 317/336/352), but two earlier sites in the same function still crash on `None`:

* **Line 277**  the `if confidence is not None and confidence < threshold` guard that was in the previous commit got reverted back to `if confidence < self._LID_CONFIDENCE_THRESHOLD:`. That now raises `TypeError: < not supported between instances of NoneType and float` on the *first* Scribe detection, before we even reach the format string. So Scribe is still 100% broken and it just crashes one line earlier than before.
* **Line 289**  the main `logger.info(f"TranscriberPool LID: {lang} conf={confidence:.2f} (provider=...)")` still uses raw `:.2f`, so even if 277 were restored, every supported-language detection would crash here.

Could the guard at 277 be restored and the same `f{confidence:.2f} if confidence is not None else "n/a"` wrapper applied at 278 and 289?

puneet-bolna: done

san0808 (approved):

PR #745, opened 2026-05-25
fix: send vobiz stop event before disconnect to prevent phantom stream reconnect

amansrivstv:
## Summary

Vobiz `<Stream keepCallAlive="true">` can cause a phantom WebSocket reconnect on the same `run_id` after the call ends, which re-enters `process_websocket` and overwrites the transcript with empty data (see BOLNA-1083). Per [Vobiz stream events docs](https://docs.vobiz.ai/xml/stream/stream-events), send `{"event":"stop","streamId":<sid>}` over the existing WS in `disconnect_stream` before the HTTP DELETE so Vobiz tears the stream down cleanly and doesn't auto-reconnect.

Pairs with dashboard-backend PR (same branch name) which adds the empty-transcript write guard as a backstop.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.59"
+__version__ = "0.10.60"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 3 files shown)


amansrivstv:

san0808 (approved):

PR #757, opened 2026-05-30
sip-trunk: fix control-frame parsing and DTMF, harden output handler

amansrivstv:
Improvements to the sip-trunk telephony handlers.

- Input handler parses bare single-token Asterisk control frames, so MEDIA_XOFF/MEDIA_XON/QUEUE_DRAINED are honored. They were previously dropped, leaving the audio-backpressure queue unused and risking dropped frames on responses long enough to fill Asterisk's buffer.
- DTMF input accumulates digits and submits on '#' or after an inter-digit timeout (SIP_DTMF_INTERDIGIT_TIMEOUT_S, default 3s), instead of one LLM turn per keypress.
- Output handler resets backpressure state (queue_full and the local queue) on interruption so the next response sends immediately instead of stacking behind stale state. The frame-send loop is now shared between the live-send and XOFF-drain paths, drains are serialized, and a closed handler returns early.
- Module docstring aligned with actual behavior.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.67"
+__version__ = "0.10.68"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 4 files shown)


prateeksachan (approved):

amansrivstv:

PR #766, opened 2026-06-02
feat: add gpt-5 reasoning_effort and verbosity to latency telemetry

san0808:
GPT-5 `reasoning_effort` and `verbosity` are sent to the OpenAI Responses API but never carried into the per-turn latency payload, so latency cannot be sliced by either in the `voiceai.llm.*` dashboards. `reasoning_effort` is a major lever on first-token latency.

This stamps both onto the latency data the same way `service_tier`/`llm_host` already flow.

1. `llms/types.py` adds `reasoning_effort` and `verbosity` to `LatencyData`.
2. `llms/openai_base.py` populates them from `self.model_args` (the exact dict sent to OpenAI, so defaults are reflected) in the first-token branch. Covers OpenAI and Azure since both subclass `OpenAICompatibleLLM`.
3. `agent_types/graph_agent.py` surfaces the effective routing `reasoning_effort` into `routing_info`.
4. `agent_manager/task_manager.py` carries it onto the routing latency turn.

Fields flow through `model_dump()` into the turn dicts automatically. The dashboard-backend side tags these onto the Datadog metrics in a companion PR.

### DIFF
--- a/bolna/agent_types/graph_agent.py
+++ b/bolna/agent_types/graph_agent.py
@@ -525,6 +525,8 @@ class GraphAgent(BaseAgent):
             if self.routing_provider == "openai" and self.service_tier:
                 routing_kwargs["service_tier"] = self.service_tier
 
+            self._routing_reasoning_effort_used = routing_kwargs.get("reasoning_effort")
+
             response = await asyncio.to_thread(self.routing_client.chat.completions.create, **routing_kwargs)
             latency_ms = (time.perf_counter() - start_time) * 1000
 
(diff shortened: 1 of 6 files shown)


vikramsingh-bolna (approved):

bolna/agent_types/graph_agent.py
```diff
@@ -879,6 +881,7 @@ async def generate(self, message: List[dict], **kwargs) -> AsyncGenerator:
                     "routing_model": self.routing_model,
                     "routing_provider": getattr(self, "routing_provider", None),
                     "routing_latency_ms": round(routing_latency_ms, 1),
+                    "routing_reasoning_effort": getattr(self, "_routing_reasoning_effort_used", None),
```
vikramsingh-bolna, line 890: q: Since these changes set the default value to None, will the old default value still be considered?

san0808: Yes. The effort is resolved first (config value or the GPT5_ROUTING_REASONING_EFFORT env default of "minimal"), then captured into _routing_reasoning_effort_used. So gpt-5 routing emits the real effort used, never None. None is only the telemetry fallback for non-gpt-5 routing where reasoning_effort does not apply. Routing behavior is unchanged.

PR #772, opened 2026-06-03
optimize datadog latency

puneet-bolna:


bolna/transcriber/deepgram_transcriber.py
```diff
@@ -981,6 +984,13 @@ async def receiver_flux(self, ws: ClientConnection):
                             self.eager_transcript_pending = transcript
                             self.last_interim_time = time.time()
 
+                            eager_latency_ms = None
+                            if words and audio_window_end:
+                                audio_sent_at = self._find_audio_send_timestamp(audio_window_end)
+                                if audio_sent_at:
+                                    eager_latency_ms = round(timestamp_ms() - audio_sent_at, 5)
+                            self._mark_last_interim_final(latency_ms=eager_latency_ms)
```
san0808, line 992: Once this marks the last interim final, nothing clears it if the turn resumes. On the EagerEndOfTurn -> TurnResumed -> more Updates -> EndOfTurn path, this entry stays `is_final=True` while EndOfTurn marks the new tail final too, so the turn emits two `interim_latency_ms{is_final:true}` points (the stale eager latency plus the real one), skewing the FINAL distribution. Common path (eager -> EndOfTurn, no new interim) is fine since both marks hit the same entry. Worth clearing prior `is_final` flags so only one entry per turn is ever final?

puneet-bolna: done

san0808 (approved):

PR #774, opened 2026-06-04
lid llm detection enhancement

puneet-bolna:


bolna/agent_manager/task_manager.py
```diff
@@ -4208,15 +4243,97 @@ def _get_voice_name_for_label(self, label):
...
+            return
+
+        context_note = (
+            f"## Language note:\nThe caller switched to '{target}'. "
+            f'Their latest message: "{detector_transcript}". Reason: {reasoning}'
+        )
+        logger.info(
+            f"LanguageSwitcher: switching '{current}' → '{target}' (confidence={target_conf}, langs={languages}, reason={reasoning})"
+        )
+        await self.switch_language(target, triggered_by="lid_llm", context_note=context_note)
```
san0808, line 4947: This runs from the background task in `_handle_transcriber_output`, concurrent with the current turn's response. The legacy tool path awaited `wait_for_current_message()` before switching, specifically to avoid swapping the pool mid-utterance. Here, if `decide()` (~1-2s) returns while turn N's response is still synthesizing, `SynthesizerPool.switch()` cancels the in-flight `_run_generate` and routes any remaining text chunks to the new-language synth, so the agent's current sentence gets truncated or finished in the wrong voice. Should we flush the current message (or defer to the next turn boundary) before applying the switch?

bolna/agent_manager/task_manager.py
```diff
@@ -4208,15 +4243,97 @@ def _get_voice_name_for_label(self, label):
...
+        if not self.language_switcher:
+            return
+        pool = self.tools.get("transcriber")
+        if not isinstance(pool, TranscriberPool):
+            return
+        labels = pool.labels
+        if len(labels) < 2:
+            return
+
+        detector_transcript, detected_lang = pool.take_lid_transcript()
```
san0808, line 4743: The buffer is drained at the main transcriber's turn boundary, but the unbiased detector is a separate socket with its own latency. At drain time it may hold only part of this turn (rest still in flight) or trailing segments from the previous turn, giving a one-turn-late or misattributed switch. Acceptable for a heuristic, but is that the intended trade-off vs. letting the detector settle briefly first?

bolna/transcriber/transcriber_pool.py
```diff
@@ -218,9 +223,12 @@ async def _standby_keepalive(self):
     async def _start_lid_tap(self) -> None:
         """Instantiate and connect the configured LID provider."""
         try:
+            # The detector runs as an unbiased transcript source only: it buffers
+            # recognition internally (drained per-turn via take_lid_transcript).
+            # on_language is left unwired — the per-segment heuristic is not used.
             self._lid = LIDProvider.create(
                 provider=self._lid_provider_name,
-                on_language=self._handle_lid_signal,
+                on_language=None,
```
san0808: With `on_language=None` here and `on_lid_switch` no longer passed by TaskManager, `_handle_lid_signal`, `_record_lid_event`, and the active-mode branch below are now unreachable. Remove the dead shadow path in this PR, or is it intentionally kept for a follow-up?

puneet-bolna: kept intentional initally this, now removed it

san0808 (approved):

bolna/agent_manager/task_manager.py
```diff
@@ -4204,19 +4381,730 @@ def _collect_flux_lid_events(self) -> list:
...
+            # follow-up is old-language audio to the caller.
+            if followup is not None:
+                await self.__generate_switch_followup(*followup)
+        except Exception as e:
+            logger.error(f"LanguageSwitcher: handler error: {e}\n{traceback.format_exc()}")
+        finally:
+            # Discard any unconsumed speculative generation — covers every exit path
+            # (stay decisions, gate rejections, timeouts, exceptions) without littering
+            # the run with per-return cancels.
+            spec = self._spec_followup_task
```
san0808: `self._spec_followup_task` is a single shared slot, but this `finally` runs after `__generate_switch_followup`, which executes outside `language_switch_lock`. While one handler is generating the follow-up (lock released), the next per-turn decision spawned by `_spawn_language_switch_decision` can acquire the lock and overwrite `self._spec_followup_task` with its own task. When the first handler then reaches this `finally`, doesn't it cancel the second handler's in-flight speculation rather than its own? And on the `spec_target != target` return path, the first handler's own spec task is left set and gets leaked instead of cancelled. Should the cleanup capture a local handle for this invocation and cancel that, instead of reading the shared attribute?

bolna/agent_manager/task_manager.py
```diff
@@ -4330,19 +4507,730 @@ def _collect_flux_lid_events(self) -> list:
...
+        """
+        messages = self.conversation_history.get_copy()
+        target_prompt = self.multilingual_prompts.get(target_label)
+        note = self.__switch_context_note(target_label, detector_transcript)
+        target_prompt = f"{target_prompt}\n\n{note}" if target_prompt else note
+        if messages and messages[0].get("role") == "system":
+            messages[0]["content"] = target_prompt
+        else:
+            messages.insert(0, {"role": "system", "content": target_prompt})
+        messages.append({"role": "user", "content": detector_transcript})
```
san0808: On the turn-boundary path, the garbled (wrong-language) user turn was already `append_user`-ed to history in `_handle_transcriber_output` before this decision spawned, so the `get_copy()` above includes it and this line appends the detector transcript as a *second consecutive* user message. Two back-to-back user turns get rejected by agents whose LLM requires alternating roles (the spec then errors out and is discarded, so the optimization never helps on this path), and for the rest the leading garbled wrong-language text pollutes the speculative reply. Should the copy drop or replace the trailing user turn before appending, the way the real path uses `replace_last_user`?

bolna/transcriber/transcriber_pool.py
san0808: Correction to my earlier comment: this is not unreachable. In the flag-OFF flow `__setup_transcriber` wires `on_lid_switch=self.switch_language` (not None), so `_start_lid_tap` sets `on_language=self._handle_lid_signal`, and it runs per segment on every flag-off call (the default during rollout). In the default `LID_MODE=shadow` it only records telemetry (no switch); with `LID_MODE=active` it also drives the legacy auto-switch. So removing it drops the flag-off shadow telemetry and the active-mode auto-switch path, not dead code. The legacy `switch_language` tool is separate and unaffected. Fine to remove if you mean to drop the per-segment heuristic entirely, but it is reachable, so that should be a deliberate call, not a dead-code cleanup. My mistake on the original.

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py

tests/test_synthesizer_pool.py
```diff
@@ -0,0 +1,138 @@
...
+
+    same_task, active = run(scenario())
+    assert same_task  # no new task started
+    assert active == "bn"
+
+
+def test_lock_serializes_switches():
+    async def scenario():
+        pool, _ = make_pool()
+        assert isinstance(pool.switch_lock, asyncio.Lock)
```
san0808: This test fails when run: `pool.switch_lock` does not exist. After the #796 rebase the lock is `_switch_lock` (private), so this raises `AttributeError` (deterministic). It will not block merge (the suite is not run in CI), but it is a broken test, and this whole file also duplicates `tests/tests/test_synthesizer_pool_switch.py` that #796 added to master. Drop this file, or rename `switch_lock` → `_switch_lock`?

tests/tests/test_language_switcher.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@pytest.mark.asyncio
+async def test_prewarm_fires_one_llm_call_and_swallows_errors():
+    import asyncio
+
+    switcher, fake_llm = _make_switcher("ok")
+    switcher.prewarm()
+    await asyncio.sleep(0)  # let the fire-and-forget task run
+    fake_llm.generate.assert_awaited_once()
```
san0808, line 120: This assertion fails when run (deterministic, 5/5). `prewarm()` create_tasks `_warm`, which awaits `wait_for(generate(...))`; after a single `await asyncio.sleep(0)`, `generate` has been *called* (call_count=1) but not yet *awaited* (await_count=0), and `assert_awaited_once()` checks await_count → "Awaited 0 times" (reaches 1 only after ~6 yields). Not a CI gate (suite is not run in CI) and the `prewarm()` code is fine; this is just an under-pumped test. Could `prewarm()` return the task so the test can `await` it, or drain more yields here?

san0808 (approved):

tests/test_synthesizer_pool.py
puneet-bolna: resolved

tests/tests/test_language_switcher.py
puneet-bolna: resolved

PR #775, opened 2026-06-06
Added custom task webhook support

puneet-bolna:


bolna/agent_manager/task_manager.py
```diff
@@ -2836,6 +2903,26 @@ async def __execute_function_call(
...
+        # Expose live call context as substitution variables (e.g. %(call_sid)s,
+        # %(stream_sid)s, %(provider)s, %(execution_id)s, %(from_number)s) so a custom
+        # tool can drive bolna's transfer service itself. System values are authoritative
+        # (the LLM can't know stream_sid), so they override any same-named tool arg.
+        resp.update(self._build_call_context())
+
+        # If a custom tool has no url configured, fall back to the internal transfer
+        # service (same env var the transfer_call branch uses). Lets a custom transfer
+        # tool work with just a number + pre_call_webhook_url — no URL to set.
+        if not url:
```
amansrivstv: This fallback fires for every custom tool, not just transfer ones — custom_task / book_appointment all flow through here. Backend validation only requires `url` as a key (the value can be null), so a tool saved with an empty url used to come back as `ERROR CALLING API`; now it silently POSTs the call context to the transfer service and can actually transfer the call. Could we gate this on the tool being transfer-intent rather than applying it universally? (Minor: `not url` also treats `""` the same as `None`.)

bolna/agent_manager/task_manager.py
```diff
@@ -645,6 +645,73 @@ def _extract_api_call_runtime_args(resp):
...
+        tool arguments plus call context. Generic — not transfer-specific; it forwards
+        whatever arguments the LLM filled in for the tool. Never blocks or fails the
+        main tool call: it is scheduled as a background task and all errors are swallowed.
+        """
+        excluded = {"model_response", "textual_response", "tool_call_id", "resp"}
+        llm_args = {key: copy.deepcopy(value) for key, value in resp.items() if key not in excluded}
+        payload = {
+            **self._build_call_context(),
+            "tool_name": called_fun,
+            **llm_args,
```
amansrivstv: `**llm_args` spread last means an LLM-supplied arg named `call_sid` / `provider` / `stream_sid` / `tool_name` overrides the system value in the webhook payload — the opposite of the `resp.update` path below where the system value wins. So the webhook can receive a model-fabricated call_sid. Probably want the context to win here too: put `**llm_args` first, then `_build_call_context()` and `tool_name`.

bolna/agent_manager/task_manager.py
```diff
@@ -645,6 +645,73 @@ def _extract_api_call_runtime_args(resp):
...
+                            meta_info,
+                            None,
+                            LogComponent.FUNCTION_CALL,
+                            direction=LogDirection.RESPONSE,
+                            run_id=self.run_id,
+                        )
+            except Exception as exc:
+                logger.warning(f"pre_call_webhook to {webhook_url} failed (ignored): {exc}")
+
+        asyncio.create_task(send())
```
amansrivstv: This keeps no reference to the created task — the event loop only holds a weak one, so under load it can be garbage-collected before the POST finishes and the webhook is silently dropped. Worth stashing it (a set, or a `self.` attr like `execute_function_call_task` elsewhere) and discarding on completion.

amansrivstv (commented):
A few correctness notes on the pre_call_webhook path, mostly around this block running for every custom tool, not just transfer-style ones. Approved.

amansrivstv (approved):

PR #779, opened 2026-06-09
removed from exclude the silence message

puneet-bolna:


puneet-bolna:
will discuss with product and solve it properly since adding will also add cause the issue

PR #780, opened 2026-06-09
fix: send GET function-call URLs verbatim instead of re-encoding params

amansrivstv:
GET function-call tools whose URL or params are already in wire form — most commonly a percent-encoded nested URL such as an Ozonetel `appURL` — were corrupted before sending. The GET branch of `trigger_api` used `session.get(url, params=api_params)`, where yarl re-quotes the URL: it leaves `?`/`:`/`/` literal inside a nested URL value (and decodes a pre-encoded one), so the receiver truncates the nested query at the literal `?` and the callback never resolves. The same request via curl works because curl sends the bytes verbatim.

The GET branch now builds the URL through `build_get_url`: the base URL is sent verbatim (`encoded=True`) and each param value is percent-encoded exactly once with `quote(safe="")`, then appended. Plain values are unchanged, a param that is itself a URL is encoded correctly rather than left with a query-splitting literal `?`, and the wire request is byte-identical to the working curl. POST paths are untouched.

Scoped to the GET path only. An audit of all live GET function-call tools (2,635 across 1,703 agents) found 2,027 byte-identical under this change and the rest decode-equivalent (`:`/`/` → `%3A`/`%2F`, mostly Cal.com); no currently-working tool regresses.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.86"
+__version__ = "0.10.87"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/helpers/function_calling_helpers.py
+++ b/bolna/helpers/function_calling_helpers.py
@@ -1,7 +1,9 @@
 import asyncio
 import json
+from urllib.parse import quote
 
 import aiohttp
+from yarl import URL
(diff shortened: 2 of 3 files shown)


prateeksachan (approved):

amansrivstv:

PR #804, opened 2026-06-17
Add SSRF guard for user-controlled outbound URLs

SubhrajitPrusty:
API tools and pre-call webhooks fetch user-supplied URLs with no
restrictions, allowing SSRF to internal services and the cloud metadata
endpoint (169.254.169.254) -> IAM credential theft.

- validate_outbound_url(): reject non-http(s) schemes and any host that
  resolves to a non-public address (loopback/RFC-1918/link-local/metadata/
  reserved/CGNAT). Public IPs and domains pass, so domain-less clients on
  public IPs are unaffected.
- Apply in trigger_api() and the pre-call webhook (skips the operator-set
  PRE_CALL_WEBHOOK_DISPATCH_URL, which may be internal).
- allow_redirects=False on these calls so a 302 -> internal hop cannot
  bypass the pre-flight check.

### DIFF
--- a/bolna/helpers/function_calling_helpers.py
+++ b/bolna/helpers/function_calling_helpers.py
@@ -48,10 +48,13 @@ async def validate_outbound_url(url):
     rebind-after-check window remains; closing it fully needs connect-time
     pinning, which we can layer on later.)
     """
-    if not url or not isinstance(url, str):
+    if not isinstance(url, str):
+        raise SSRFError("Missing or invalid request URL")
+    url = url.strip()
+    if not url:
         raise SSRFError("Missing or invalid request URL")
 
-    parsed = urlsplit(url.strip())
+    parsed = urlsplit(url)
     scheme = parsed.scheme.lower()
     if scheme not in ALLOWED_URL_SCHEMES:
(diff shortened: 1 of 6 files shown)


bolna/helpers/function_calling_helpers.py
```diff
@@ -10,6 +12,74 @@
...
+        raise SSRFError("Request URL has no host")
+
+    try:
+        port = parsed.port
+    except ValueError:
+        raise SSRFError("Request URL has an invalid port")
+
+    loop = asyncio.get_running_loop()
+    try:
+        infos = await loop.getaddrinfo(host, port, type=socket.SOCK_STREAM)
```
san0808: This resolves the host once here, but aiohttp's connector re-resolves at connect time, so a low-TTL record that returns a public IP to this `getaddrinfo` and an internal IP to the actual connection would still get through (the rebind window the docstring notes). Could we close it now rather than defer, by pinning the validated IP for the connection (connect to the resolved IP with the original Host header), or by running `_is_disallowed_ip` inside a custom resolver / `TCPConnector` so the check happens at connect time? That would also re-validate redirect hops, so `allow_redirects=False` could be relaxed.

san0808 (approved):

PR #805, opened 2026-06-18
Add host allowlist to the outbound URL SSRF guard

san0808:
The SSRF guard added in #804 blocks any tool URL that resolves to a private address. Some deployments legitimately call an internal endpoint (for example an internal load balancer) from a tool, which the guard now blocks with no way to permit it.

This adds `BOLNA_TOOL_URL_HOST_ALLOWLIST`: a comma-separated list of trusted hostnames that bypass the block, matched case-insensitively against the URL host. It is empty by default, so behavior is unchanged unless an operator sets it. Cloud metadata, loopback, and every other private range stay blocked for non-allowlisted hosts.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.101"
+__version__ = "0.10.102"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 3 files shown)


prabalrastogi (approved):
LGTM

PR #808, opened 2026-06-19
fix: support azure provider for graph agent routing client

san0808:
Graph agent routing fails with a 404 (`Resource not found`) on every turn when the agent's `routing_provider` is `azure`. `_init_routing_client` only handled `groq` and `openai`, so an azure routing config fell through to the openai branch and reused the plain `OpenAI` client pointed at the Azure endpoint. That client posts to `{base_url}/chat/completions`, which Azure rejects (it needs the `/openai/deployments/<name>/...?api-version=...` path), so routing always returned no decision and the graph never advanced past its first node.

This adds an explicit azure branch that builds an `AzureOpenAI` client with the resource endpoint and api-version, and strips the `azure/` provider prefix from the deployment name. It also forwards `service_tier` to azure routing, matching the existing openai behaviour.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.104"
+__version__ = "0.10.105"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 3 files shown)


prabalrastogi (approved):
LGTM

PR #813, opened 2026-06-23
feat: typed input variables for graph-agent expression routing

san0808:
Expression-based routing compares a runtime variable against a literal, but input variables arrive untyped (strings from CSV/telephony, JSON-typed from the API). Comparisons can therefore run in the wrong domain: `"9" > "10"` is True lexically but `9 > 10` is False numerically, and string `"true"` never matches a boolean literal, so an edge can silently mis-route a live call.

This adds a `variable_types` map on the graph agent config (`name -> string | number | boolean`) so the evaluator coerces both sides into the declared domain before comparing, with literal inference as the fallback for undeclared values.

## Behavior

* A declared type coerces both operands (and list elements for `in`/`not_in`) into that type, applied uniformly across `eq`/`neq`/`gt`/`gte`/`lt`/`lte`/`in`/`not_in`/`contains`.
* Keys match the condition's variable path exactly (e.g. `recipient_data.age`).
* Undeclared variables fall back to inference, so existing configs are unchanged.
* A coercion that fails on a declared variable logs a warning and the condition evaluates False (fail-closed).
* Prompt rendering is untouched; values still render as their original strings.

## Usage

Types are declared once on the graph agent config; the call payload keeps sending plain variables.

```json
"variable_types": { "recipient_data.age": "number", "recipient_data.is_member": "boolean" }
```

### DIFF
--- a/bolna/helpers/expression_evaluator.py
+++ b/bolna/helpers/expression_evaluator.py
@@ -5,9 +5,12 @@ context_data for instant routing decisions.
 """
 
 import operator as op
-from typing import Any
+from typing import Any, Optional
 
-from bolna.enums import ExpressionOperator, ExpressionLogic, EdgeConditionType
+from bolna.enums import ExpressionOperator, ExpressionLogic, EdgeConditionType, VariableType
+from bolna.helpers.logger_config import configure_logger
+
+logger = configure_logger(__name__)
 
 _MISSING = object()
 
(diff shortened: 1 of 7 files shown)


pavankapoor31 (approved):

vikramsingh-bolna (approved):

bolna/helpers/expression_evaluator.py
```diff
@@ -32,10 +35,81 @@ def resolve_variable(context_data: dict, path: str) -> Any:
...
+_BOOL_TOKENS = _TRUE_TOKENS | _FALSE_TOKENS
+
+
+def _is_bool_like(value: Any) -> bool:
+    """True for a bool or a string that names one, so inference only treats genuine
+    booleans as boolean and leaves numbers/other strings to numeric comparison."""
+    return isinstance(value, bool) or (isinstance(value, str) and value.strip().lower() in _BOOL_TOKENS)
+
+
+def _to_bool(value: Any) -> bool:
```
SubhrajitPrusty, line 49: there is a util, `from dstutils.utils import strtobool`. maybe helpful here.

san0808: don't want to add a dependency for this , but will keep in mind to extend if multiple issues ban rahe hai

PR #828, opened 2026-06-29
added the support for web calling for better quality

puneet-bolna:


bolna/input_handlers/telephony_providers/freeswitch.py
```diff
@@ -0,0 +1,98 @@
...
+            return
+        await super().process_message(message)
+
+    def ingest_audio(self, data):
+        if self.conversation_recording:
+            if self.conversation_recording["metadata"]["started"] == 0:
+                self.conversation_recording["metadata"]["started"] = time.time()
+            self.conversation_recording["input"]["data"] += data
+        self.ingest_buffer += data
+        if len(self.ingest_buffer) < self.INGEST_CHUNK_BYTES:
```
san0808, line 47: The trailing sub-`INGEST_CHUNK_BYTES` (<200ms) residual in `ingest_buffer` is never flushed, so the tail of a final short utterance right before hangup can be dropped, and mid-turn the last partial chunk waits for the next audio to arrive. Minor for streaming ASR, but should `_listen` flush the remainder on disconnect?

bolna/agent_manager/task_manager.py
```diff
@@ -1410,6 +1445,9 @@ def __setup_transcriber(self):
                         if is_sip:
                             cfg["encoding"] = "mulaw"
                             cfg["sampling_rate"] = 8000
+                        elif provider in ("web_based_call", TelephonyProvider.FREESWITCH.value):
```
san0808: `"web_based_call"` is a magic string compared alongside `TelephonyProvider.FREESWITCH.value` here and again a few lines down. It is not in the enum. Would a named constant read more consistently? (Same class of nit as the hardcoded `24000` welcome rate in `__init__`.)

bolna/input_handlers/telephony_providers/freeswitch.py
puneet-bolna: done

bolna/agent_manager/task_manager.py

san0808 (approved):

san0808 (approved):

PR #831, opened 2026-07-02
added pre warm silence issue fix:

puneet-bolna:


bolna/agent_manager/task_manager.py
```diff
@@ -4660,18 +4726,73 @@ def __language_switch_enabled(self) -> bool:
...
+        synth = self.tools.get("synthesizer")
+        return not isinstance(synth, SynthesizerPool) or detected in synth.labels
+
+    async def __run_llm_after_switch_decision(self, transcriber_package, decision_task, spawn_language):
+        """Hold the main reply while the switch decision resolves. Switched → drop the
+        reply (the switch's follow-up answers in the new language); stayed, timed out,
+        or switched via the in-flight-tool branch (which produces no follow-up) → run
+        it. Shielded so cancelling the held reply never kills the decide."""
+        # Must cover the decide's own budget (settle 300ms + decide up to 8s + lock
+        hold_timeout = float(os.getenv("LANGUAGE_SWITCH_MISMATCH_HOLD_S", "5.0"))
```
san0808: Should this default be derived from the decide budget instead of being a second independent knob? LANGUAGE_SWITCH_DECIDE_TIMEOUT_S defaults to 8s (plus 300ms settle and any lock wait), so a slow decide outlives the 5s hold: the old-language reply starts playing into the silence and the confirmed switch then truncates it mid-sentence, which is the exact artifact the hold was added to remove. The comment above also cuts off mid-sentence.

bolna/lid/base.py
```diff
@@ -1,26 +1,226 @@
...
+    def _schedule_reconnect(self, source: str):
+        """Mark the socket dead and kick off one reconnect task (idempotent)."""
+        if self._stopping:
+            return
+        self._dead = True
+        if self._reconnecting:
+            return
+        self._reconnecting = True
+        logger.error(f"{self.__class__.__name__}: socket dead ({source}) — detector mute, attempting reconnect")
+        asyncio.create_task(self._reconnect())
```
san0808: Can this task outlive the call? The handle is not stored, _shutdown_connection only cancels sender/receiver, and _reconnect never re-checks _stopping after its sleep, so a stop() landing inside the reconnect window still lets start() open a fresh provider socket and spawn new sender/receiver tasks for a call that already ended. Related: if the start() retry raises, _reconnect returns with _dead still True and no live loops, so nothing can ever schedule another attempt even though the cap says attempts remain.

bolna/helpers/language_switcher.py
```diff
@@ -5,18 +5,35 @@
...
+    LANGUAGE_SWITCH_LLM_API_* wins; else the provider's standard env — ANTHROPIC_API_KEY
+    for claude, AZURE_OPENAI_* for azure/* (matches bolna/llms/azure_llm.py), else OPENAI_API_KEY.
+    """
+    key = os.getenv("LANGUAGE_SWITCH_LLM_API_KEY") or ""
+    base = os.getenv("LANGUAGE_SWITCH_LLM_API_BASE") or ""
+    version = os.getenv("LANGUAGE_SWITCH_LLM_API_VERSION") or ""
+    if model.startswith("azure/"):
+        key = key or os.getenv("AZURE_OPENAI_API_KEY") or ""
+        base = base or os.getenv("AZURE_OPENAI_ENDPOINT") or ""
+        version = version or os.getenv("AZURE_OPENAI_API_VERSION") or ""
```
san0808: azure_llm.py falls back to api_version 2024-12-01-preview when AZURE_OPENAI_API_VERSION is unset, but this resolves to an empty string (the docstring above says it matches azure_llm.py). On a pod without that env var the agent's own Azure LLM keeps working while every decide() fails, so switching goes silently inert for exactly the orgs granted the azure flag. Should this reuse azure_llm's resolution, and should a judge that fails prewarm fall back to DEFAULT_LANGUAGE_SWITCH_LLM instead of failing closed on every decide?

puneet-bolna: done

bolna/lid/base.py
puneet-bolna: done

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py
```diff
@@ -696,8 +723,8 @@ def __init__(
         # switch; new: played to cover the switch → follow-up generation gap).
         self.switch_handoff_messages = self.task_config.get("tools_config", {}).get("switch_handoff_messages") or {}
         self.agent_names = self.task_config.get("tools_config", {}).get("agent_names") or {}
-        if not self.__language_switch_enabled():
-            self.__inject_switch_language_tool()
+        # Injected for BOTH flows. In the LLM-driven flow the tool is the fast path for
+        self.__inject_switch_language_tool()
```
san0808, line 741: Master injected switch_language only when the LLM-driven flow was off; now the main agent LLM gets it in both flows, and its inline handler switches without any of the gates the Switch LLM enforces (min confidence, the short-audio substance gate) and without the hangup_triggered check that __execute_function_call applies to other tools. Is the intent that an explicit caller request via the main LLM is trustworthy enough to skip all gates? If so, should the handler at least check hangup_triggered so a tool switch cannot start synthesizing a handoff while the goodbye is queued? The comment on this line also cuts off mid-sentence.

bolna/agent_manager/task_manager.py
```diff
@@ -3140,40 +3165,51 @@ async def __execute_function_call(
...
-                    "text": handoff_text,
-                }
-                self._turn_audio_flushed.clear()
-                await self._synthesize(create_ws_data_packet(handoff_text, meta_info=meta_info_handoff))
-                await self.wait_for_current_message()
-                self.conversation_history.append_assistant(handoff_text, turn_id=turn_id, response_uid=response_uid)
-                if turn_id is not None:
-                    self._turn_msg_map[turn_id] = self.conversation_history.messages[-1]
+            # Serialize with the LID decide (which holds this lock): without it a
+            async with self.language_switch_lock:
```
san0808: This holds language_switch_lock across wait_for_current_message (up to full playback of the in-flight reply plus the 3s flush wait), handoff synthesis, and a second wait_for_current_message, so the lock can be held for many seconds. Every LID decide queues on the same lock, and a mismatch hold waiting on a queued decide can exceed its budget (the derived default covers settle + decide + 1s, not a multi-second lock wait), reopening the play-then-truncate window. Could the flip itself be the only part under the lock, with the playback waits and handoff synthesis outside it, as handle_language_switch already does?

bolna/agent_manager/task_manager.py
```diff
@@ -166,9 +170,29 @@ def build_lid_decision_record(
         "switched_to": switched_to,
         "context_note_sent": context_note,
         "context_note_sent_at": now if context_note else None,
+        # Old-language response state when this firing resolved — for a switch, whether
+        "inflight_activity": inflight_activity or {},
     }
 
 
+def trailing_utterance_text(segments, gap_seconds=1.5):
```
san0808: Is a 1.5s gap safe as the utterance boundary here? This feeds the silent-switch composed turn, the post-switch fallback, and the idle flush, and it is strictly subtractive: a caller who pauses to think for 2s mid-request ("I need tomorrow's flight status ... [2s] ... should I give you the PNR?") gets only the trailing fragment appended as their turn, and the head is unrecoverable because the buffer was drained. Master composed prefix plus tail. Should the pre-gap text be kept (or the gap widened) when the segments belong to the same turn window?

bolna/prompts.py
```diff
@@ -111,9 +111,14 @@
...
-4. CLOSELY RELATED OR ACOUSTICALLY CONFUSABLE LANGUAGES (e.g. Hindi/Marathi/Maithili/Konkani, Hindi/Urdu, Bengali/Assamese, and the Dravidian cluster Telugu/Tamil/Kannada/Malayalam): a clean LIVE transcript is WEAK evidence the caller speaks the active language (the locked recognizer decodes the sibling plausibly), and the unbiased tag itself often confuses cluster siblings — especially on short audio. Decide from distinctive function words (Marathi आहे/तुम्ही/आपण vs Hindi है/आप) rather than either signal alone.
+3a. NUMBERS, CODES, AND IDENTIFIERS ARE NOT LANGUAGE EVIDENCE. A turn that is predominantly a readout of digits, a spelled-out code, an order/tag/reference/OTP/phone/account number, or a string of Latin letters (e.g. "G P one five one two five S", "mera number 98765", "double seven three") is the caller supplying DATA in response to a prompt — English digits and Latin letters are spoken inside every Indian language. NEVER switch on a predominantly numeric/alphanumeric turn; stay in the current language. Judge only from the surrounding grammatical content, if any (rule 2).
```
san0808: LANGUAGE_SWITCH_PROMPT is no longer imported by production code (language_switcher uses LANGUAGE_SWITCH_SYSTEM_PROMPT + LANGUAGE_SWITCH_TURN_PROMPT); its only consumer is test_prompt_variants_rules_stay_in_sync, yet every rule edit in this PR had to be made twice. Could it be deleted (pointing the test at the system prompt), or composed from the two live pieces so there is one source of truth?

bolna/agent_manager/task_manager.py
```diff
@@ -1514,6 +1537,11 @@ def __setup_synthesizer(self, llm_config=None):
 
                 logger.info(f"SynthesizerPool created with labels={list(synthesizers.keys())}, active='{active_label}'")
 
+                # Pre-render every language's handoff clip on its own voice (background;
+                if self.language_switcher is not None and not self.turn_based_conversation:
+                    if self.task_config["tools_config"]["output"]["provider"] != "default":
+                        self.handoff_prewarm_task = asyncio.create_task(self.__prewarm_handoff_clips())
```
san0808, line 1557: Every call re-renders the handoff clips: handoff_audio_cache is per TaskManager and cleared at teardown, so an agent with N languages pays N one-shot TTS renders per call, concurrent with welcome-message synthesis at the moment first-word latency matters most, and billed even though most calls never switch. Could the clips be cached across calls keyed by voice + text (they are static per agent config), or rendered lazily on the first switch?

bolna/lid/soniox.py
```diff
@@ -0,0 +1,163 @@
...
+        else:
+            self._audio_format = "pcm_s16le"
+            self._input_sr = int(config.get("sampling_rate", 16000))
+        self._pending = self._empty_pending()
+
+    @staticmethod
+    def _empty_pending() -> dict:
+        return {"text": "", "lang_counts": {}, "last_lang": None, "start_ms": None, "end_ms": None}
+
+    def _build_config(self) -> dict:
```
san0808, line 52: This mirrors SonioxTranscriber almost line for line (_build_config here, the telephony to audio_format/sample_rate mapping in __init__ vs _resolve_audio_params, the ws URL in start() vs get_soniox_ws_url). A wire-protocol change will land in the transcriber and silently miss the LID tap. Worth extracting shared soniox config/url helpers that both classes call?

bolna/agent_manager/task_manager.py
```diff
@@ -5283,35 +5374,124 @@ async def __post_switch_reply_fallback(self):
...
+            return
+
+        async def render(label, synth):
+            text = self.__handoff_text_for(label)
+            if not text:
+                return
+            try:
+                clip = None
+                # Prefer a native mu-law 8000 one-shot when the provider offers it (no
+                telephony_one_shot = getattr(synth, "synthesize_telephony_clip", None)
```
san0808, line 5487: synthesize_telephony_clip is an undeclared hook only ElevenlabsSynthesizer implements; every other provider falls into __handoff_clip_to_mulaw, which byte-sniffs containers and getattr-guesses sample rates inside TaskManager. Should this be part of the BaseSynthesizer contract (a declared one-shot output spec or to_telephony method) so new providers implement it instead of extending the sniffer, with the transcode living next to pcm_to_ulaw in helpers?

bolna/agent_manager/task_manager.py
```diff
@@ -6502,6 +6678,8 @@ async def run(self):
 
                 output["progression_data"] = {
                     "call_start_epoch_ms": self.conversation_start_init_ts,
+                    # Ground-truth transcript so the dashboard renders directly, not from latency gaps.
+                    "messages": copy.deepcopy(output["messages"]),
```
san0808, line 6705: This embeds a second full copy of the transcript inside the same output payload that already carries output["messages"] at the top level, doubling transcript bytes shipped and stored per call. Does progression_data need to be self-contained, or can its consumer read the sibling key?

bolna/agent_manager/task_manager.py
```diff
@@ -6556,6 +6734,35 @@ async def run(self):
...
+                    _tid = _tt.get("turn_id")
+                    if not isinstance(_tid, int) or _tid in _ub_turns or not _tt.get("final_transcript"):
+                        continue
+                    _u_start = _tt.get("asr_turn_start_ms")
+                    if _u_start is None:
+                        _u_start = _tt.get("asr_start_ms")
+                    _u_end = _tt.get("user_speech_end_ms")
+                    if _u_end is None:
+                        _u_end = _tt.get("asr_finalized_ms")
+                    output["progression_data"]["user_bot_latencies"].append(
```
san0808, line 6776: These appended rows have sequence_id, agent_start_ms, agent_end_ms, and latency_ms all None, where user_bot_latencies previously carried only fully populated answered-turn records. Do the dashboard consumers of this JSONB tolerate None latency_ms (averages, progression rendering), or should unanswered-turn records go under a separate key?

bolna/agent_manager/task_manager.py
```diff
@@ -4660,18 +4726,80 @@ def __language_switch_enabled(self) -> bool:
...
+        land, so the main reply should wait for it."""
+        pool = self.tools.get("transcriber")
+        if not isinstance(pool, TranscriberPool):
+            return False
+        detected = pool.lid_buffer_language()
+        active_short = (self.language or "").split("-")[0].lower()
+        if not detected or detected == active_short or detected not in pool.labels:
+            return False
+        synth = self.tools.get("synthesizer")
+        return not isinstance(synth, SynthesizerPool) or detected in synth.labels
```
san0808, line 4783: Should this pre-filter on the substance gate too? It holds the reply on nothing but the latest segment language tag, so an acknowledgment-length mis-tag (the short-audio class the decide gates on later) still costs the caller the full settle + decide wait on a turn where the decision ends in gated:short_audio anyway. Reusing the buffered max-segment-seconds threshold here would skip the hold exactly where the decide cannot switch.

bolna/helpers/language_switcher.py
```diff
@@ -32,48 +51,42 @@ def __init__(self, available_labels, run_id=None, model=None):
...
             max_tokens=200,
             temperature=0.0,
             llm_key=switch_llm_key,
-            base_url=os.getenv("LANGUAGE_SWITCH_LLM_API_BASE", ""),
-            api_version=os.getenv("LANGUAGE_SWITCH_LLM_API_VERSION", ""),
+            base_url=switch_llm_base,
+            api_version=switch_llm_version,
         )
 
     def prewarm(self):
```
san0808, line 94: The prewarm pays the TLS handshake but sends only a tiny user message, so for the Anthropic default the ~1.5k-token system block with cache_control is never seeded and the first real decide of the call, the one the hold design cares about most, still pays the cold prompt-cache write. Should the prewarm send the real system block so the first decide is a cache read?

san0808 (approved):

PR #836, opened 2026-07-06
fixed the elevan labs eos issue

puneet-bolna:
Issue

On call 8909c48c-1afa-43d1-93f7-53cd7b3c03c3, the transcript showed the full LLM response but the caller heard almost none of it — only the tail of the previous turn. Both turns ended with the identical phrase ("…बात करना चाहेंगे?").

Root cause chain:

1. The ElevenLabs sender nulls context_id when a turn ends normally (close_context at end_of_llm_stream) — but that context's audio frames are still draining from ElevenLabs.
2. When the user barged in, handle_interruption found context_id = None and did nothing — the closed-but-draining context became invisible to the interruption machinery, so its frames kept flowing through the receiver.
3. One of those stale frames ended with the same phrase as the new turn's text, tripping the receiver's text-match end-of-stream heuristic (which compares against the shared mutable current_text, unscoped to context) — a false end_of_synthesizer_stream fired ~200ms into the new turn.
4. The false EOS caused on_successful_response_delivered → retire_sequence_id, so when the new turn's real audio arrived moments later, every chunk was dropped (Skipping message with sequence_id).

The transcript still showed the full text because assistant history is committed on first-chunk send, and the heard-text reconciliation (sync_history) never ran — the sequence was falsely marked successfully delivered.

The same misattribution also exists via isFinal: a stale context's late isFinal pops the new turn's meta_info from the FIFO and can retire it prematurely, even without a shared phrase.

Fix

Confined to elevenlabs_synthesizer.py:

1. current_turn_context_id — stamped in _on_push when the context is minted; unlike context_id, it survives close_context, keeping the draining context attributable.
2. handle_interruption blacklists it into context_ids_to_ignore (in addition to the live context_id), so a closed-but-draining context's frames are dropped by the existing ignore mechanism; also resets current_turn_start_time in that path so the next turn is re-detected as new.
3. Receiver EOS gate — end-of-stream (from either the text-match or isFinal branch) is suppressed when the frame's contextId differs from current_turn_context_id.

Why safe

- No hang risk (the stuck-audio_playing failure mode): suppression is only possible when a successor turn exists — whose own EOS sets the flags. With no successor, the gate always passes and behavior is byte-identical to before.
- No audio truncation: only EOS attribution is gated; audio yielding is untouched, so legitimately-draining tails (filler flows) still play.
- Zero base-class / task_manager changes — other stream synthesizers unaffected.

### DIFF
--- a/bolna/synthesizer/elevenlabs_synthesizer.py
+++ b/bolna/synthesizer/elevenlabs_synthesizer.py
@@ -97,7 +97,9 @@ class ElevenlabsSynthesizer(StreamSynthesizer):
         return audio, {"text_synthesized": text_synthesized}
 
     def _on_push(self, meta_info, text):
-        if not self.context_id:
+        # Mint only for pushes that will actually synthesize — a superseded push must not
+        # advance current_turn_context_id, or the prior turn's real isFinal gets suppressed.
+        if not self.context_id and self.should_synthesize_response(meta_info.get("sequence_id")):
             self.context_id = str(uuid.uuid4())
             self.current_turn_context_id = self.context_id
 
--- a/bolna/synthesizer/elevenlabs_synthesizer.py
+++ b/bolna/synthesizer/elevenlabs_synthesizer.py
@@ -73,6 +73,7 @@ class ElevenlabsSynthesizer(StreamSynthesizer):
         self.context_id = None
         self.context_ids_to_ignore = set()
         self._eos_context_id = None  # context the last end-of-stream was emitted for
+        self.current_turn_context_id = None  # survives close_context, unlike context_id
         self.ws_send_time = None
         self.ws_trace_id = None
         self.current_turn_ttfb = None
@@ -98,6 +99,7 @@ class ElevenlabsSynthesizer(StreamSynthesizer):
     def _on_push(self, meta_info, text):
         if not self.context_id:
             self.context_id = str(uuid.uuid4())
+            self.current_turn_context_id = self.context_id
 
     # ------------------------------------------------------------------
     # Format helper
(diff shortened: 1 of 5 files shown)


bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -269,9 +275,15 @@ async def receiver(self):
                 else:
                     logger.info("No audio data in the response")
 
+                # A stale (previous-turn) context's frames must never end the current turn's stream.
+                is_stale_context = (
+                    ctx is not None and self.current_turn_context_id is not None and ctx != self.current_turn_context_id
+                )
+                if emit_eos and is_stale_context:
```
san0808, line 284: `_on_push` advances `current_turn_context_id` to a fresh context whenever `context_id` is None, which includes the window after `sender` has closed a turn's context (`close_context`) but before that context's `isFinal` has drained. In a back-to-back agent flow with no user barge-in (eager/filler response then the real response, or a filler during a tool call), the second push mints a new context and flips `current_turn_context_id` while the first context is still owed its `isFinal`. When that `isFinal` lands, `is_stale_context` is true, so the first turn's end-of-stream is suppressed, it never yields the `b"\x00"` sentinel, and `end_of_synthesizer_stream` / `_record_turn_latency` never fire for it.

The "no hang, the successor always emits its own EOS" argument assumes the successor is a real speaking turn. But `_on_push` mints unconditionally, before `sender`'s `should_synthesize_response` gate runs, so a push for a superseded sequence also advances `current_turn_context_id` to a context that then no-ops and never emits anything, leaving the older turn's real EOS suppressed with no successor to recover it.

Would it be safer to gate suppression on the stale context actually being interrupted (present in `context_ids_to_ignore`) rather than any `ctx != current_turn_context_id`? That keeps the barge-in fix while still letting a normally-closed prior context emit its own EOS.

puneet-bolna: On gating suppression via context_ids_to_ignore instead: that check would be a
no-op at the suppression point — frames whose ctx is in context_ids_to_ignore
are already dropped at the top of the receiver loop and never reach the EOS
logic. It would effectively remove the gate, reintroducing the failure it
exists for: a normally-closed prior context's late isFinal gets attributed to
the CURRENT turn's meta via the text_queue FIFO pop, prematurely retiring it —
that's the 0ms-speaking-session silent-call bug from the incident calls, and it
doesn't require a barge-in.

For back-to-back real turns (your first scenario): with the mint gated, the
successor is by construction a speaking turn, so its EOS restores the
turn-level flags; the prior turn's loss is bounded to its latency record and
recovery-counter telemetry. Pre-fix, the same window misattributed the EOS and
silenced the successor turn — strictly worse. And a suppression deadlock can't
occur: suppression requires the successor's push to already exist, while
anything awaiting the prior turn's flush blocks before pushing it.

san0808 (approved):

PR #839, opened 2026-07-07
fixed the llm voice hangup addition in progression

puneet-bolna:


bolna/agent_manager/voicemail_handler.py
```diff
@@ -154,6 +165,12 @@ async def _background_check(self, transcriber_message: str, meta_info: dict, is_
                 self.detected = True
                 self.tm.hangup_detail = HangupReason.VOICEMAIL_DETECTED
                 await self._handle_detected()
+        except asyncio.CancelledError:
```
san0808, line 173: This records the attempt during call teardown. Is the final latency dict guaranteed to still be snapshotted after this append? If the dict is assembled before the cancel propagates through, the cancelled entry gets recorded but never persisted. Not a blocker, just worth confirming the ordering.

bolna/agent_manager/voicemail_handler.py
```diff
@@ -154,6 +165,12 @@ async def _background_check(self, transcriber_message: str, meta_info: dict, is_
...
+            # Call ended (or barge-in) while the check was still awaiting the LLM. CancelledError is a
+            # BaseException, so the `except Exception` below would miss it — record the attempt (with the
+            # elapsed time) so a cancelled check is still visible, then re-raise to honour cancellation.
+            self._record_latency(meta_info, latency_ms=round((time.time() - check_start) * 1000, 2), cancelled=True)
```
san0808: Heads up on a downstream side-effect: the cancelled record sets `latency_ms` to elapsed-until-cancel, and `dashboard-backend/utils/datadog_helpers.py` emits `voiceai.llm.response_time_ms` for every `other_latencies` item that has a `latency_ms`, with no `cancelled` guard. So cancelled checks (which aborted mid-flight, not a real completion) will skew the LLM response-time distribution downward. The main LLM generation cancel path avoids conflating the two by stamping a separate `cancelled_at_ms` rather than reusing `latency_ms`. Worth doing the same here, or guarding the datadog emit on `not item.get("cancelled")`.

bolna/agent_manager/voicemail_handler.py
puneet-bolna: done

bolna/agent_manager/voicemail_handler.py
puneet-bolna: done

san0808 (approved):

PR #842, opened 2026-07-08
fix: reset audio flag after handle_interruption in is-user-online branch

web1havv:
When the "are you still there?" prompt is synthesized and then handle_interruption() is called immediately after, Plivo discards the queued audio (clearAudio) and the mark dictionary is wiped (clear_data). This means the final-chunk mark echo from Plivo never arrives, leaving is_audio_being_played_to_user() stuck True forever.

The audio gate in __check_for_completion (line 6213) then hits `continue` on every 2s tick, permanently blocking the silence-hangup path. The call runs until Plivo's 300s carrier max-duration kills it instead of hanging up cleanly at hangup_after_silence.

Fix: explicitly reset the audio flag to False right after handle_interruption() since cleared audio can never produce a completion mark and the flag must reflect that reality.

Fixes deterministic 300s call duration for silent callers on Plivo when check_if_user_online=true and trigger_user_online_message_after < hangup_after_silence.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.132"
+__version__ = "0.10.133"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -6289,6 +6289,13 @@ class TaskManager(BaseManager):
                         await self._synthesize(create_ws_data_packet(user_online_message, meta_info=meta_info))
                     self.conversation_history.append_assistant(user_online_message, exclude_from_llm=True)
 
+                    # Explicitly reset the audio flag after synthesizing the prompt.
+                    # handle_interruption() below sends clearAudio to Plivo and wipes the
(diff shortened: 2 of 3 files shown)


puneet-bolna (approved):

baptistecristo:
@web1havv good catch on #842. Tracing a stuck `is_audio_being_played_to_user` back to `clearAudio` wiping the mark dict, and then to calls sitting until Plivo's 300s ceiling instead of hanging up, takes patience most people don't spend on a 9-line diff.

I maintain an open-source voice agent you reach from a dumbphone. You call a number, ask for the weather or a reminder, and hang up. It runs on Vapi, so the silence-and-hangup path you just fixed in bolna is one I don't own. I don't own the STT or TTS slots either, and I don't choose which country the audio crosses. The calls carry personal data, so that last part is the one that bothers me.

Dropping that dependency is open here: https://github.com/baptistecristo/AI-agents-for-dumbphones/discussions/9

I saw your GSoC entry tests for ns-3 and MBDyn. If you want a project to sink into, this one is small and the phone rings. Repo: https://github.com/baptistecristo/AI-agents-for-dumbphones

PR #843, opened 2026-07-08
fixed the force fianlise for flux

puneet-bolna:


bolna/transcriber/deepgram_transcriber.py
```diff
@@ -432,16 +431,22 @@ async def monitor_utterance_timeout(self):
...
     async def _release_stuck_flux_turn(self):
-        # speech_ended ends the user turn downstream (resets callee_speaking) without injecting a
-        # transcript or LLM call; _reset_turn_state then drops any later finalization for this turn.
-        await self.push_to_transcriber_queue(create_ws_data_packet({"type": "speech_ended"}, self.meta_info))
-        self._reset_turn_state()
+        # Deliver the buffered words (mirrors Nova's force-finalize) so the user's speech still
+        # reaches the LLM; bare speech_ended only when the stuck turn produced no text — it still
+        # resets callee_speaking downstream, and _reset_turn_state drops any later finalization.
+        if self.final_transcript.strip() or self.current_turn_interim_details:
+            await self._force_finalize_utterance()
```
san0808, line 454: When this fires right after an `EagerEndOfTurn` that already started a speculative task, does it end up with an orphaned eager task plus a duplicate turn?

`_force_finalize_utterance` emits `{type: transcript, force_finalized: True}` with no `was_eager` key, so downstream in task_manager `was_eager` is False and it takes the `_handle_transcriber_output` branch, which cancels/reassigns `self.llm_task` but never touches `self.eager_llm_task`. The eager task keeps running to completion (its audio stays BLOCKed, but any tool call in it still executes), the eager user turn appended to history is never reverted (no `turn_resumed` on this path), and the fresh llm task appends the user turn again.

This only bites eager-enabled agents that stall immediately after an eager fire, so narrow, but the failure mode (orphaned tool side-effect + duplicated history) is the same class as the barge-in double-tool bug. Worth cancelling any in-flight eager task in `_release_stuck_flux_turn` before force-finalizing, or propagating `was_eager` so downstream reuses it?

puneet-bolna: considered this case as well

san0808 (approved):

PR #846, opened 2026-07-08
fix: add utterance timeout monitor to Soniox transcriber

web1havv:
Soniox uses semantic endpoint detection which can stall indefinitely if audio stops mid-utterance (e.g. due to a degraded/dropped Plivo media stream). Unlike Deepgram and Gladia, Soniox had no monitor_utterance_timeout mechanism to recover from this.

Add _force_finalize_utterance() and monitor_utterance_timeout() to SonioxTranscriber, mirroring the existing Deepgram implementation. If no endpoint token arrives within 8 seconds of the last interim, the accumulated transcript is force-finalized and pushed to the pipeline so the call can continue instead of freezing silently.

The 8s default is intentionally larger than Deepgram's 1.2s because Soniox's semantic endpointing legitimately takes longer on natural pauses. Configurable via interim_timeout kwarg.

### DIFF
--- a/bolna/transcriber/soniox_transcriber.py
+++ b/bolna/transcriber/soniox_transcriber.py
@@ -420,6 +420,7 @@ class SonioxTranscriber(BaseTranscriber):
             raise
         except Exception as e:
             logger.error(f"Soniox utterance timeout monitor error: {e}")
+            raise
 
     async def push_to_transcriber_queue(self, data_packet):
         await self.transcriber_output_queue.put(data_packet)
--- a/bolna/transcriber/soniox_transcriber.py
+++ b/bolna/transcriber/soniox_transcriber.py
@@ -71,6 +71,11 @@ class SonioxTranscriber(BaseTranscriber):
         self.audio_frame_duration = 0.5
         self._resolve_audio_params()
 
+        # Utterance timeout — force-finalize if Soniox never sends endpoint after last interim
+        _interim_timeout = kwargs.get("interim_timeout")
+        self.interim_timeout = float(_interim_timeout) if _interim_timeout is not None else 8.0
+        self.utterance_timeout_task = None
+
         # Connection + task state
         self.websocket_connection = None
         self.connection_authenticated = False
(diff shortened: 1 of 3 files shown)


web1havv:

bolna/transcriber/soniox_transcriber.py
```diff
@@ -356,18 +361,74 @@ async def receiver(self, ws: ClientConnection):
...
+        self.meta_info["user_stop_offset_ms"] = self.user_stop_offset_ms
+        if self._last_detected_language:
+            self.meta_info["transcriber_detected_language"] = self._last_detected_language
+
+        data = {"type": "transcript", "content": transcript_to_send, "force_finalized": True}
+        logger.info(f"Soniox force-finalized transcript after timeout: {transcript_to_send}")
+        await self.push_to_transcriber_queue(create_ws_data_packet(data, self.meta_info))
+        self._reset_turn_state()
+
+    async def monitor_utterance_timeout(self):
```
puneet-bolna, line 399: looks good just one q: at other provider levels also we are following this same code?

web1havv: yes monitor_utterance_timeout is there everywhere in provider

bolna/transcriber/soniox_transcriber.py
```diff
@@ -356,18 +361,74 @@ async def receiver(self, ws: ClientConnection):
...
+                        logger.warning(
+                            f"Soniox utterance timeout: no endpoint for {elapsed:.1f}s, "
+                            f"force-finalizing turn {self.current_turn_id}"
+                        )
+                        await self._force_finalize_utterance()
+        except asyncio.CancelledError:
+            logger.info("Soniox utterance timeout monitor cancelled")
+            raise
+        except Exception as e:
+            logger.error(f"Soniox utterance timeout monitor error: {e}")
```
SubhrajitPrusty, line 420: raise on cancel but not on generic exception?

web1havv: Added raise at the end of the generic except block so any unexpected error in the monitor propagates instead of silently killing the task

SubhrajitPrusty (approved):

PR #851, opened 2026-07-14
fix: hold graph routing until a node delivers its first TTS turn (BOLNA-1582)

san0808:
## Problem

A graph node could be skipped without ever speaking its first question. When a user's affirmation is split into two ASR turns, the first fragment transitions into the next node and begins generating its question, then the second fragment arrives as a barge-in and cancels that question before any audio is heard. On the next turn the agent evaluates the (never-spoken) node's edges and routes onward, so its required question is never asked.

## Fix

Track whether the active node has delivered its first customer-facing TTS turn, and hold routing out of a speaking node until that turn completes.

* The flag resets on every transition into a node (`_advance_to_node` and the event advance).
* The task manager sets it once a full response is delivered (`on_successful_response_delivered`), which always lands before the next user turn, so it is node-correct.
* `generate()` holds routing on an LLM node whose first response has not been delivered and lets the node speak its question instead. User input arriving in that window stays in conversation history as context but does not trigger forward routing.

The initial node starts as already delivered, since its first turn is the welcome message, preserving current first-turn routing.

## Behavior note

A barge-in during a node's very first question now re-delivers the question rather than treating the barge-in as its answer, matching the requirement that routing only follows a completed first TTS turn.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -6215,6 +6215,11 @@ class TaskManager(BaseManager):
                             )
                         self.interruption_manager.on_successful_response_delivered(sequence_id)
                         self.interruption_manager.on_agent_speech_ended()
+                        if (
+                            self.__is_graph_agent()
+                            and message["meta_info"].get("message_category", "") != "is_user_online_message"
+                        ):
+                            self.tools["llm_agent"].mark_first_response_delivered()
                     # Reset asked_if_user_is_still_there flag after any message except is_user_online_message
                     if message["meta_info"].get("message_category", "") != "is_user_online_message":
                         self.asked_if_user_is_still_there = False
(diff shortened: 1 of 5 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -182,6 +182,8 @@ def build_lid_decision_record(
 HANDOFF_CLIP_CACHE: dict = {}
 HANDOFF_CLIP_CACHE_MAX = 256
 
+_NON_NODE_RESPONSE_CATEGORIES = frozenset({"is_user_online_message", "filler", "backchanneling"})
```
puneet-bolna: welcome message also should we add ?

san0808: Good call. Added agent_welcome_message, and also handoff while here ig it's the same class but riskier: a language-switch clip fully delivering on a just-entered node would mark it delivered and let the question be skipped mid-conversation. Both now excluded from the delivery signal.

puneet-bolna (approved):

PR #861, opened 2026-07-18
fix(openai-transcriber): force-finalize late finals from interims (#711)

### LINKED ISSUE #711: OpenAI realtime whisper hears speech but final transcript comes too late, so agent never replies
When using the OpenAI realtime transcriber (gpt-realtime-whisper) on Twilio calls, the system gets interim transcript chunks quickly, but the final transcript comes very late, sometimes 20–35 seconds later.

Because the backend only sends the user message to the LLM after the final transcript arrives, the assistant never replies during the actual call. Instead, it may play “Hey, are you still there?” or hang up for inactivity before the final transcript is processed.

What I observed:

OpenAI transcriber connects successfully
interim transcript tokens appear quickly
no LLM turn starts
final transcript arrives way too late
call can end before reply happens
Example symptom:
User says things like:

“Hey Brooke, what’s up?”
“What’s going on, Brooke?”
“Hey Brooke, can you hear me?”
But the agent never answers in time.

Environment ( git clone of master branch latest upstream )

vikas-prasad-cx:
## Summary
- Fixes [#711](https://github.com/bolna-ai/bolna/issues/711): OpenAI realtime whisper often delivers interim deltas quickly, but `transcription.completed` can arrive 20–35s later on Twilio. The LLM only starts on the final transcript, so the agent never replies in time.
- On utterance timeout after commit, force-finalize from accumulated interim deltas and ignore a late `completed` for the same turn (avoids duplicate LLM turns).
- Expose `speech_rms_threshold` on the `Transcriber` model so agents can tune the client-side RMS gate for noisy mulaw lines; default telephony threshold is 500 (non-telephony stays 400).
- Raise `OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S` from `0.5` → `2.0` so we wait briefly for a real completed event before falling back.

## Test plan
- [x] `pytest tests/tests/test_openai_transcriber_force_finalize.py -v` (12 passed)
- [ ] Optional live check: Twilio + `provider: openai` / `gpt-realtime-whisper` — confirm LLM responds within ~2s of end-of-utterance when `completed` is delayed
- [ ] Optional: set `"speech_rms_threshold": 250` (or higher) in agent transcriber config and confirm speech detection logs show the override

Made with [Cursor](https://cursor.com)

### DIFF
--- a/bolna/constants.py
+++ b/bolna/constants.py
@@ -11,7 +11,13 @@ WEB_BASED_CALL_PROVIDER = "web_based_call"
 WEBCALL_TTS_SAMPLE_RATE = 24000
 
 OPENAI_TRANSCRIBER_HEARTBEAT_INTERVAL_S = 5
-OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S = 0.5
+# Wait this long after commit for transcription.completed before force-finalizing
+# from accumulated interim deltas (avoids 20–35s silent waits on Twilio).
+OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S = 2.0
+# Default client-side speech gate (RMS of 24 kHz PCM16). Telephony often needs a
+# higher floor so line noise does not keep a turn open indefinitely.
+OPENAI_TRANSCRIBER_SPEECH_RMS_THRESHOLD = 400
+OPENAI_TRANSCRIBER_TELEPHONY_SPEECH_RMS_THRESHOLD = 500
 
 # ElevenLabs realtime (scribe_v2_realtime) accepts up to 50 keyterms for biasing.
 ELEVENLABS_REALTIME_MAX_KEYTERMS = 50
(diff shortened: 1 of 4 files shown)


vikas-prasad-cx:
## E2E evidence (Twilio + OpenAI whisper)

- Answered Twilio call `CAd1cbcee350308ef897e74bc83e1e034b` (outbound-api, completed)
- Branch under test with OpenAI `gpt-realtime-whisper` on Twilio mulaw, `speech_rms_threshold=500`
- Logs showed:
  - `Speech detected (RMS=1272 threshold=500.0)`
  - interim `"Hello"` → `Transcript completed for turn turn_1: Hello`
  - `first_interim_to_final_ms ≈ 676` (not the 20–35s failure mode from #711)
- Unit tests: `pytest tests/tests/test_openai_transcriber_force_finalize.py -v` → 12 passed

vikas-prasad-cx:
Hi — gentle bump, and a quick update: this branch is now rebased onto current `master` (resolved a conflict with the newer OpenAI turn-latency id logic) so it should be mergeable again.

Summary: force-finalize late/missing `transcription.completed` from interim text for Twilio/OpenAI Whisper turns (#711), with a higher telephony `speech_rms_threshold` default and unit coverage. E2E notes are earlier in the thread.

Happy to adjust further if useful for review. Thanks!

PR #866, opened 2026-07-20
Surface Sarvam stream rejections instead of ending the call silently

san0808:
Sarvam signals a rejected stream with a `{"type": "error"}` message and then closes the socket. `receiver` had no branch for that message type, and the close was swallowed by `except Exception: traceback.print_exc()`, so `connection_error` stayed `None`.

The result was a call that ended with no reason recorded anywhere: the event was logged as a benign drop rather than an error, the execution was not attributed to a transcriber failure, and the only trace was a stderr traceback with no run_id on it. Diagnosing an inbound SIP-trunk outage took a manual replay against the Sarvam endpoint to recover a message the server had already sent us.

Changes:

1. Handle `type == "error"` in `receiver`, recording the message in `connection_error`. Sarvam's own close still ends the loop, so teardown timing is unchanged.
2. Log the websocket close code and reason, and treat a non-1000 close as a connection error.
3. Route the remaining `traceback.print_exc()` calls in this module through the logger, so they carry the run_id.

Behavior change: a Sarvam-side rejection or abnormal close is now attributed as `transcriber_connection_error` rather than ending the call silently.

Follow-up to #860, which fixes the sample-rate mismatch that this would have made obvious.

Verified against the live Sarvam endpoint. Sending 8 kHz to saaras:v3 now logs `Sarvam rejected the stream: Error in Pipeline : Invalid request: 'audio.sample_rate' must be 16000` and the emitted packet carries that message, at the same 254 ms teardown as before. A correctly configured 16 kHz stream is unaffected and still returns transcripts.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.144"
+__version__ = "0.10.145"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/transcriber/sarvam_transcriber.py
+++ b/bolna/transcriber/sarvam_transcriber.py
@@ -11,7 +11,7 @@ from dotenv import load_dotenv
 import aiohttp
 import websockets
 from websockets.asyncio.client import ClientConnection
-from websockets.exceptions import InvalidHandshake
+from websockets.exceptions import ConnectionClosed, InvalidHandshake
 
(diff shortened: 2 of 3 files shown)


san0808:
Folded into #860 as an extra commit, since the two changes cover the same failure.

PR #868, opened 2026-07-20
Enhance/webcall

puneet-bolna:
### 1. Scribe (ElevenLabs) turn latency: `interim_timeout` 5.0s → 1.2s
`scribe_v2_realtime`'s server-side VAD commit frequently never fires on short Hindi utterances ("हिंदी में", "ठीक है"), so bolna's force-finalize fallback was the de-facto endpointing — and it waited a full 5 seconds. That stall *was* the reported 3.5–6s webcall turn latency (verified across runs 23ccefe1, a03e62bf, 7c844501: `Interim timeout: No finalization for 5.0s`). Now 1.2s, matching the deepgram transcriber's default. Safe: the timer resets on every partial and scribe streams partials ~1/s while the user speaks, so it fires only after real silence + a missed commit. Worst case drops from ~5–6s to ~1.2–2.2s.

### 2. Force-finalize duplicate-turn race fix
With the shorter timeout, scribe's *late* `committed_transcript` for an already-force-finalized utterance could arrive afterwards and pass the `is_transcript_sent_for_processing` guard (the reset left it False) — pushing the same utterance as a second turn: bot answers a fragment, then re-answers the full sentence. `_force_finalize_utterance` now closes the gate (`True`) after pushing; the next utterance's first partial re-opens it, so only the stale commit is swallowed. Mirrors the deepgram transcriber's post-force-finalize state. The no-transcript early-return keeps the gate open. Tests: `tests/test_elevenlabs_force_finalize.py`.

### 3. Welcome audio source-rate support (pairs with dashboard-backend)
`welcome_pcm_upsampled(welcome_b64, target, source_sample_rate=8000)` — passthrough when source == target, resample otherwise; TaskManager reads `welcome_message_audio_sample_rate` from the ws payload (missing → 8000, byte-identical for legacy payloads). The backend now sends 24kHz welcomes for web calls so the first turn matches the full-band TTS instead of sounding like a muffled telephone voice; telephony always receives 8000 (mulaw path untouched). Call-site guard skips the resample+LRU entirely for already-24k welcomes. Tests: `tests/test_welcome_pcm_upsampled.py`.

### Testing
20 tests across the transcriber/welcome suites; full suite green except a pre-existing broken file (`test_event_injection_e2e.py`, fails on master too).

### Deploy notes
Ship with dashboard-backend `enhance/webcall` (it produces the welcome rate field and the 24k welcome objects). Deployed and verified on topaz-ws-mumbai.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bolna/transcriber/elevenlabs_transcriber.py
+++ b/bolna/transcriber/elevenlabs_transcriber.py
@@ -108,9 +108,11 @@ class ElevenLabsTranscriber(BaseTranscriber):
         # Latency tracking
         self.last_audio_send_time = None
 
-        # Timeout tracking for stuck utterances
+        # Timeout tracking for stuck utterances. 2.0s ≈ 2× scribe's partial cadence
+        # (~1s between partials in prod), so a slightly late partial doesn't trigger a
+        # mid-utterance force-finalize, while a genuinely stuck commit is cut at ~2s not 5s.
         self.last_interim_time = None
-        self.interim_timeout = kwargs.get("interim_timeout", 1.2)
+        self.interim_timeout = kwargs.get("interim_timeout", 2.0)
         self.utterance_timeout_task = None
 
     def get_elevenlabs_ws_url(self):
--- a/bolna/transcriber/elevenlabs_transcriber.py
+++ b/bolna/transcriber/elevenlabs_transcriber.py
@@ -243,6 +243,11 @@ class ElevenLabsTranscriber(BaseTranscriber):
         logger.info(f"Force-finalized transcript after timeout: {transcript_to_send}")
         await self.push_to_transcriber_queue(create_ws_data_packet(data, self.meta_info))
         self._reset_turn_state()
+        # This utterance was already pushed — suppress ElevenLabs' own (late) commit for it,
+        # or the same turn gets processed twice (fragment now + full text seconds later).
+        # The next utterance's first partial flips this back to False, so only the stale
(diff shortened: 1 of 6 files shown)


bolna/transcriber/elevenlabs_transcriber.py
```diff
@@ -110,7 +110,7 @@ def __init__(
 
         # Timeout tracking for stuck utterances
         self.last_interim_time = None
-        self.interim_timeout = kwargs.get("interim_timeout", 5.0)
+        self.interim_timeout = kwargs.get("interim_timeout", 1.2)
```
SubhrajitPrusty: this will change it for all defaults, is this valid?

puneet-bolna: yes we have force finalise timeout of 1.2s for all others 
cc @san0808

san0808: @puneet-bolna this 1.2 secs was based on stanbdard time we observed for deepgram and by default it was 5 secs too first and then we saw ki for p95 the time taken is max 1 secs so we then changed it to 1.2

did u check the same for elevenlabs ?

puneet-bolna: Screenshot 2026-07-22 at 10 55 23 AM

this is the data for safer side should we do 2sec or 1.5 sec?

san0808 (approved):

PR #872, opened 2026-07-22
 fix eos stream

puneet-bolna:


bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -267,23 +274,20 @@ async def receiver(self):
...
+                        spoken_norm = self.normalize_text(self.eos_accum_text.strip()).replace('"', "").strip()
                         # Strip whitespace before compare: ElevenLabs alignment splits
                         # "first-time" into "first- time", breaking endswith.
-                        last_four_cmp = re.sub(r"\s+", "", last_four)
                         current_cmp = re.sub(r"\s+", "", current_norm)
-                        if last_four_cmp and has_alnum and current_cmp.endswith(last_four_cmp):
+                        spoken_cmp = re.sub(r"\s+", "", spoken_norm)
+                        logger.info(f"EOS check spoken_len={len(spoken_cmp)} current_len={len(current_cmp)}")
+                        # End the stream only once the WHOLE turn text has been spoken, not when a
+                        # truncated frame fragment (e.g. "s.") coincidentally suffixes it (87da790e).
```
san0808, line 293: If the final segment text also occurs earlier in the same turn at a frame boundary (short repeated closers, e.g. a turn that says "Sure." twice with final push "Sure."), spoken_cmp matches one segment early. A premature match also sets _eos_context_id, so the context's genuine isFinal gets deduped afterwards. Is that acceptable exposure here, or should a later real isFinal still be let through when the text-match fired before the last audio frame?

san0808 (approved):
just one issue which is not due to ur pr but something to think of

bolna/synthesizer/elevenlabs_synthesizer.py
puneet-bolna: Added a length floor to the ElevenLabs end-of-stream text-match so it can't fire on a repeated closer.

bolna/synthesizer/elevenlabs_synthesizer.py, in the receiver() elif self.last_text_sent: block.

# Require ~the whole turn spoken before trusting the suffix match, else a
# repeated closer ("Sure." twice, final push "Sure.") matches one segment early.
spoken_enough = (
    self.current_sequence_chars <= 0
    or len(self.eos_accum_text) >= 0.9 * self.current_sequence_chars
)
...
if current_cmp and spoken_enough and spoken_cmp.endswith(current_cmp):
    emit_eos = True

Why this and not "let a later isFinal through"

- The suffix match compares against current_text, which is only the last push (stream_synthesizer.py:157), not the whole turn. So on a multi-push turn where the final segment repeats earlier at a frame boundary ("Sure." … "Sure."), spoken_cmp ends with current_cmp after the first closer → premature EOS. Confirmed exposure.
- Un-deduping isFinal can't recover it. The premature b"\x00" is emitted immediately, retires the sequence, and from then is_valid_sequence(sequence_id) is False (task_manager.py:5879) so every later frame hits Skipping message with sequence_id (5916). By the time isFinal arrives the audio is already dropped — a later isFinal would only emit a redundant terminal signal, recovering nothing. The lever is not emitting the premature EOS, which is what the floor does.

san0808: Confirmed the trace: the premature end_of_synthesizer_stream reaches __process_output_loop, on_successful_response_delivered retires the sequence, and __listen_synthesizer drops every later frame, so blocking the emission is the right lever and the floor does that. The genuine final frame lands at ~100% of seq chars, so the correct case is unaffected.

One small leftover: current_sequence_chars is only reset in _record_turn_latency, which never runs for an interrupted turn (handle_interruption nulls current_turn_start_time and the suppressed EOS skips it). The leftover chars inflate the next turn's denominator, so the turn after a barge-in can have the text-match silently disabled and fall back to isFinal. Safe direction, fine to ship as is; a current_sequence_chars = 0 reset in handle_interruption would tidy it. Resolving.

PR #886, opened 2026-07-25
fixing the false inactivity timeout detection:

puneet-bolna:


MujeerAhmed (approved):

MujeerAhmed:
Pushed a follow-up commit (`2e31f59f`) to this branch — it closes a residual gap the timestamp fix alone doesn't cover.

**Repro (staging, run `1170a1c5`, this branch deployed):** the INACTIVITY hangup was correctly avoided ✅, but the agent then fired `trigger_user_online_message_after` ("क्या आप अभी भी लाइन पर हैं?") **mid-utterance** and cleared its still-playing closing line.

**Why `compute_last_ai_audio_timestamp` can't catch it:** the closing line was a **single ~16s TTS chunk** (`mark 5274033a`, `duration 16.07`, `acked:false / cleared_on_interrupt:true`). A single chunk emits one pre-mark (start) + one content-ack (end) — so for the whole 16s there are no intermediate marks to refresh `get_current_mark_started_time()`. At the check the clock read 7.92s, `is_audio_being_played` was already cleared at synth stream-end, and 7.92s > the 6s nudge threshold → nudge fired and interrupted/cleared the still-playing audio.

**Added:** `InterruptionManager.is_agent_audio_pending_playback()` — True when the latest turn has `agent_start_s` but no `agent_end_s` (final-chunk ack not yet in). Folded into the completion watchdog's audio gate so neither the nudge nor `hang_conversation_after` fires while the caller is still hearing the agent. The stall backstop is intentionally left on the raw flag (+ your fresh timestamp), so a genuinely stuck stream is still cleaned up after `STALL_HANGUP_FLOOR_S` — no permanent latch. Unit test added.

Deeper trigger worth a separate look: capping max TTS chunk duration would keep marks flowing every ~2s and prevent this class of issue at the source.

SubhrajitPrusty (approved):

PR #887, opened 2026-07-26
fix: base the inactivity watchdog on a queued-audio playout estimate

san0808:
Follow-up to #886. Same bug, one mechanism instead of two, and the configured timeout is honoured again.

#886 fixes the false hangup with two changes: a refreshed AI-silence clock, and a gate that treats "this turn's final chunk has not been ACKed" as busy. On live reproductions the clock half is not sufficient on its own. It is driven by `update_start_ts`, which stops advancing once the provider stops emitting marks, and on one call that happened 7.8s before the agent actually finished speaking. At that point the 6s online-check threshold is reachable mid-utterance, so only the gate prevented the prompt firing over the closing question. The clock is still wrong; the gate compensates for it. Anything else keyed on that clock, `repeat_after_silence` and the stall backstop, is reading the same wrong value.

This measures the thing directly instead. `MarkEventMetaData` already records a per-chunk duration, so it now accumulates the wall-clock end of everything handed to the provider, and drops it in `clear_data()` when the provider is told to discard its queue. `compute_last_ai_audio_timestamp` clamps that to now, so measured silence reads 0 while the caller can still hear the agent and starts counting the moment playback ends. Replayed against real calls the estimate lands within 0.02s and 0.36s of the provider's own final-chunk ACK.

Because every watchdog branch is gated on that clock, the extra audio term in the gate is no longer needed and comes out, along with `is_agent_audio_pending_playback`. That also removes a latch: the gate is only ever cleared by an ACK, and a barge-in destroys the pending marks, so an interrupted turn that is the last agent turn of a call leaves the watchdog blocked for the rest of it, with no inactivity hangup, no online check and no repeat-after-silence, only the 20s stall backstop.

The estimate cannot latch in the same way. It is a deadline, so it expires by the clock passing it.

Two details. The online-check prompt is excluded from the estimate so it does not postpone the inactivity hangup, matching the exclusion it already has from `final_chunk_played_observable` and keeping `hangup_after_silence` measured from the last real agent speech as configured. Web-call marks now carry a duration too, so the same estimate applies there rather than only on telephony; the mark wire format the browser sees is unchanged.

Validated on live calls against a build carrying this change. The failure window reproduced at 24.5s and 28.7s of audio still playing after the flag was cleared, held correctly in both, with the online check and hangup landing at 7.2s and 11.2s measured from real playback end.


bolna/output_handlers/default.py
```diff
@@ -3,15 +3,23 @@
...
 class DefaultOutputHandler:
     def __init__(
-        self, io_provider="default", websocket=None, queue=None, is_web_based_call=False, mark_event_meta_data=None
+        self,
+        io_provider="default",
+        websocket=None,
+        queue=None,
+        is_web_based_call=False,
+        mark_event_meta_data=None,
+        sampling_rate=24000,
```
puneet-bolna: default we should keep 8k?

san0808: No, 8k would be wrong here. This is the web/dashboard handler and web plays raw PCM at 24k (`use_mulaw=False`, and `task_manager` sets the synth rate to 24000 for non-telephony). Telephony never reads this: `TelephonyOutputHandler` overrides `handle()` entirely and computes its own duration against a hardcoded 8000, so the inherited value is unused there.

Fair catch on the magic number though. Switched the default to the existing `WEBCALL_TTS_SAMPLE_RATE`, which already carries the "telephony stays 8k mulaw" note, so the 24k is self-documenting instead of looking arbitrary. b5661bb0

puneet-bolna (approved):

PR #888, opened 2026-07-27
fixed repititve llm call

puneet-bolna:
Summary

Root cause (reproduced empirically): A tool-call turn is recorded as plain text first with tool_calls attached after (commit-then-attach). When the staged text is dropped before commit, attach_tool_calls_to_turn leaves a uid-less placeholder holding the tool_calls. The heard-audio materializer (sync_history → upsert_assistant_for_response) then can't match it by response_uid and appended a duplicate plain-text assistant between the tool-call turn and its tool result. _sanitize_tool_messages (run in get_copy() before every LLM request) then sees the broken pairing, strips the tool_calls and drops the tool result — so the follow-up generation has no record the tool ran and re-issues it. That's the double reschedule_call + trailing non_purchase_reason.

The fix (conversation_history.py, upsert_assistant_for_response): when not found by response_uid, adopt the uid-less placeholder for the same turn_id in place (fill content + uid, preserve tool_calls) instead of duplicating. ~10 lines, one method, data layer. It only touches uid-less placeholders, so a legit distinct response already owning a response_uid in the same turn is never clobbered (covered by a test).

Verification:
- scratchpad/repro_blt018.py — shows the strip pre-fix (tool_calls survive? False) → preserved post-fix (True); clean ordering unaffected.
- New tests/tests/test_conversation_history_tool_call_preservation.py (4 cases) — 2 fail pre-fix / all pass post-fix, proving it's a real regression guard.
- 24 pass across the new + related history/task-manager suites; no regressions.
- INIT.md updated under ## Bugs Found & Resolved.


bolna/agent_manager/task_manager.py
```diff
@@ -2927,6 +2931,18 @@ async def _process_conversation_formulaic_task(self, message, sequence, meta_inf
                         )
                     )
 
+    @staticmethod
+    def _tool_call_signature(called_fun, resp):
+        """Stable key for an API tool invocation: function name + resolved args, ignoring the
+        OpenAI plumbing (model_response/tool_call_id/textual_response). Two calls with the same
+        signature produce the same side effect, so the second is a safe no-op."""
+        ignore = {"model_response", "tool_call_id", "textual_response"}
+        args = {k: v for k, v in resp.items() if k not in ignore}
```
san0808: **Blocking: the signature only covers `**resp`, which makes some tools once-per-call.**

`url` / `method` / `param` are named parameters of `__execute_function_call`, so they never reach `resp`. For a tool whose spec declares no properties, nothing variable is left after the filter. Verified against a real `FunctionCallPayload`:

```
zero-arg resp keys : ['model_response', 'resp', 'textual_response', 'tool_call_id']
signature          : custom_task_check_stock|{"resp": null}
zero-arg dedupes?  : True
same-args re-check dedupes?: True
```

So any `check_status` / `fetch_cart` / `get_current_offers` style tool silently no-ops from its second invocation onward for the rest of the call, and read tools re-called with identical args are blocked too.

The underlying issue is that this is an unconditional global policy. A tool knows whether it is safe to repeat; the task manager doesn't. Suggest an opt-in flag in `api_tools.tools_params` (`idempotent: true` / `once_per_call: true`) and only consulting the set for those tools and that also removes the need to reason about every retrieval tool in the fleet.

(Also: `"resp"` should be in `ignore` for consistency with `fire_pre_call_webhook`'s `excluded` at `:911` it's why `{"resp": null}` appears in every signature.)

bolna/agent_manager/task_manager.py
```diff
@@ -3376,6 +3392,27 @@ async def __execute_function_call(
             )
             return
 
+        # Idempotency guard (BLT-018): a burst of short user acknowledgements during call
+        # closing spawns fresh LLM turns that re-emit the same action tool with identical args
+        # (e.g. reschedule_call), re-hitting the external API. Same pattern as the transfer_call
+        # guard above - skip the re-run and record an "already done" result so the LLM stops
+        # re-triggering. Keyed on tool name + args, so a retrieval tool called with NEW args
```
san0808: This reads as reassurance that retrieval tools are unaffected, but only tools called with **new** args are safe. A retrieval tool re-called with the **same** args is deduped and gets `{"status": "success", "message": "Already done earlier..."}` with no data the `check_availability_of_slots` for the same date, `get_order_status` for the same id, "can you check that again?". Worth rewording so the boundary isn't overstated (or, better, narrowing the guard so the claim becomes true).

bolna/agent_manager/task_manager.py
```diff
@@ -3376,6 +3392,27 @@ async def __execute_function_call(
...
+            self.conversation_history.append_tool_result(
+                resp.get("tool_call_id", ""),
+                json.dumps(
+                    {
+                        "status": "success",
+                        "message": "Already done earlier in this call. Do not call this tool again; continue the conversation.",
+                    }
+                ),
+            )
+            return
```
san0808, line 3432: Two problems with returning here.

**`check_if_user_online` stays `False` for the rest of the call.** `:2949` sets it `False` on entry, and only the normal path (`:3530`) and `switch_language` (`:3372`) restore it. The other early returns get away with it because the call is ending and this one continues the conversation, so after a single deduped tool call the agent never asks "are you still there?" again (`:6527`) and just sits silent until the inactivity hangup fires.

**No follow-up generation → dead air.** The message tells the LLM to "continue the conversation", but the method returns without calling `__do_llm_generation`, so the LLM never sees the tool result this turn. The turn ends right after a spoken filler like "one moment please", with nothing following it. `switch_language` handles the analogous case at `:3379-3383` by feeding the result back with `should_trigger_function_call=True`.

Something like:

```python
self.check_if_user_online = self.conversation_config.get("check_if_user_online", True)
messages = self.conversation_history.get_copy()
followup_meta_info = self._spawn_followup_meta_info(meta_info)
await self.__do_llm_generation(
    messages, followup_meta_info, next_step, should_trigger_function_call=True
)
self.execute_function_call_task = None
return
```

bolna/agent_manager/task_manager.py
```diff
@@ -3466,6 +3503,8 @@ async def __execute_function_call(
         textual_response = resp.get("textual_response", None)
         self.conversation_history.attach_tool_calls_to_turn(turn_id, resp["model_response"])
         self.conversation_history.append_tool_result(resp.get("tool_call_id", ""), function_response)
+        # Mark this exact call executed so an identical re-emission on a later turn is skipped.
+        self.executed_tool_call_signatures.add(call_signature)
```
san0808: **Blocking: a failed API call is recorded as executed here.**

`trigger_api` never raises and `asyncio.TimeoutError`, `SSRFError` and the bare `except Exception` all return a dict with an error body (`function_calling_helpers.py:273-317`), and non-2xx statuses return normally too. Only `CancelledError` propagates. So this line runs after a failure just as it does after a success.

Failure scenario: `book_appointment` times out after 10s → LLM sees `"ERROR CALLING API: Request ... timed out"` → says "let me try that again" → re-emits identical args → the guard above returns `{"status": "success", "message": "Already done earlier in this call."}`. The agent now tells the caller their appointment is booked when nothing was booked.

Suggest recording only on a real success:

```python
if 200 <= (response.get("status_code") or 0) < 300 and not response.get("error"):
    self.executed_tool_call_signatures.add(call_signature)
```

Related, in the other direction: because this is only reached *after* `await trigger_api`, a barge-in that cancels the turn mid-request re-raises at `:3462-3464` without recording, so a delivered-but-cancelled request isn't deduped. That's exactly the "burst of acks" timing this PR targets, so it may be worth recording optimistically before the request and clearing on a confirmed failure instead.

tests/tests/test_tool_call_idempotency.py
```diff
@@ -0,0 +1,65 @@
...
+Fix: __execute_function_call skips a repeat whose signature (tool name + resolved args) was already
+executed this call, recording an "already done" result instead (same shape as the transfer_call guard).
+These tests pin _tool_call_signature: identical calls collapse (deduped), genuinely different calls do
+not (still run), and OpenAI plumbing keys never affect the identity.
+"""
+
+from bolna.agent_manager.task_manager import TaskManager
+
+
+sig = TaskManager._tool_call_signature  # static method, call directly
```
san0808: All six cases exercise `_tool_call_signature` as a pure function and they pin the key format, but nothing covers the guard itself. The bugs I flagged all live in the untested part:

- a second identical call actually skips `trigger_api` and appends the tool result
- a **failed** first call still permits a retry (currently it doesn't)
- a zero-argument tool isn't permanently locked out
- `check_if_user_online` survives the skip path

An `__execute_function_call` test with `trigger_api` patched would cover all four and would have caught the failure-retry case.

tests/tests/test_tool_call_idempotency.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_different_args_are_distinct_calls():
+    # A different requested time is a genuinely different reschedule -> must NOT be deduped.
+    assert sig("custom_task_reschedule_call", _reschedule("later")) != sig(
+        "custom_task_reschedule_call", _reschedule("tomorrow 5pm")
+    )
+
+
+def test_non_purchase_reason_with_different_reasons_not_deduped():
```
san0808: Worth double-checking this boundary against the ticket. BOLNA-2017 lists "the `non_purchase_reason` tool call right after the second reschedule tool call" as part of the reported problem, but this test locks in that a re-emitted `non_purchase_reason` with different reason text still runs. If the spurious trailing call is in scope, this guard won't stop it and the test documents that as intended.

bolna/agent_manager/task_manager.py
```diff
@@ -2927,6 +2931,17 @@ async def _process_conversation_formulaic_task(self, message, sequence, meta_inf
                         )
                     )
 
+    def _is_once_per_call(self, called_fun):
+        """A tool the agent opted into firing at most once per call (tools_params[...].idempotent)."""
+        return bool(self.kwargs.get("api_tools", {}).get("tools_params", {}).get(called_fun, {}).get("idempotent"))
```
san0808: **Blocking: this always returns `False` in production, because the flag doesn't survive config validation.**

`APIParams` (`bolna/llms/types.py:8`) doesn't declare `idempotent` and doesn't set `extra="allow"` the way `FunctionCallPayload` does right below it, so pydantic v2's default `extra='ignore'` drops it:

```
APIParams extra policy      : ignore (pydantic v2 default)
idempotent survives validate: False
model_dump keys             : ['api_token', 'headers', 'method', 'nodes', 'param',
                               'pre_call_message', 'pre_call_webhook_param',
                               'pre_call_webhook_url', 'scope', 'url']
```

dashboard-backend validates agent configs against exactly this class (`models.py:118` -> `from bolna.models import *` -> `bolna/models.py:574`), so `idempotent: true` in an agent's `tools_params` never reaches `self.kwargs`. `_is_once_per_call` returns `False` for every tool, the guard never fires, and BLT-018 reproduces unchanged for the reporting agent.

```python
class APIParams(BaseModel):
    ...
    pre_call_webhook_param: Optional[Union[str, dict]] = None
    # Fire this tool at most once per call; a re-emission is skipped (BLT-018).
    idempotent: Optional[bool] = False
```

Please also thread it through the dashboard-backend tool-config schema and confirm with a real agent payload end to end. The unit tests build `tm.kwargs` by hand and never cross the validation boundary, so they pass either way, which is why this wasn't caught.

bolna/agent_manager/task_manager.py
```diff
@@ -3376,10 +3391,39 @@ async def __execute_function_call(
...
+
+        # Idempotency guard (BLT-018), OPT-IN per tool via tools_params[...].idempotent. A burst of
+        # short user acknowledgements during call closing spawns fresh LLM turns that re-emit the
+        # same action tool, re-hitting the external API - the same failure the transfer_call guard
+        # above prevents. Only once-per-call tools are guarded, so retrieval tools (product_search,
+        # check_availability, zero-arg status checks) that may legitimately repeat are untouched.
+        # After a tool has succeeded once this call, skip the re-run and feed an "already done"
+        # result back through a normal follow-up generation so the agent keeps talking (never a
+        # dead-air return). The matching pre-call filler is suppressed in __do_llm_generation.
+        if self._is_once_per_call(called_fun) and called_fun in self.executed_once_per_call_tools:
```
san0808, line 3414: This is keyed on tool name alone, so a re-emission with **different arguments** is also skipped. That's a plausible turn on the very tool in this ticket:

> "Actually, can you make it 5pm tomorrow instead of this evening?"

The second `reschedule_call` never fires, the caller is told "Already done earlier in this call", and the booking stays at the original time. `test_once_per_call_dedupes_even_with_different_args` pins this as intended, so I want to check it's deliberate rather than a side effect of simplifying away the old signature key.

Two coherent options:

1. Keep it name-keyed, but rename the config flag to `once_per_call` so it matches `executed_once_per_call_tools`, `_is_once_per_call`, and every comment in this PR. "Idempotent" conventionally means *same args, same effect, safe to repeat*, so the current name invites exactly the misconfiguration above.
2. Key on name plus resolved args. Now that the guard is opt-in, most of the objections to signature-keying from the first pass go away, since the zero-arg problem only bites tools that opted in.

Either works as long as the name matches the behaviour. I'd lean toward (2) for `reschedule_call` specifically.

bolna/agent_manager/task_manager.py
```diff
@@ -3376,10 +3391,39 @@ async def __execute_function_call(
...
+            )
+            self.conversation_history.attach_tool_calls_to_turn(turn_id, resp["model_response"])
+            self.conversation_history.append_tool_result(resp.get("tool_call_id", ""), already_done)
+            convert_to_request_log(
+                already_done, meta_info, None, "function_call", direction="response", run_id=self.run_id
+            )
+            self.check_if_user_online = self.conversation_config.get("check_if_user_online", True)
+            messages = self.conversation_history.get_copy()
+            followup_meta_info = self._spawn_followup_meta_info(meta_info)
+            await self.__do_llm_generation(messages, followup_meta_info, next_step, should_trigger_function_call=True)
```
san0808, line 3430: **Recursion risk.** `should_trigger_function_call=True` means this follow-up can emit tool calls, including the same once-per-call tool — which lands back in the skip branch above and calls `__do_llm_generation` again. There's no depth guard anywhere in `task_manager.py` (I grepped), so nothing structurally bounds it.

The normal tool path has the same recursive shape, but each iteration hits a real API and returns new data, so the model's input changes. Here it's a fixed point: identical history in, identical `"Already done earlier in this call"` out — so a model that ignores the instruction once will likely ignore it every time, growing the stack and history until the call drops.

Cheapest fix is to match the `end_call` no-textual path and let the model only talk:

```python
await self.__do_llm_generation(
    messages, followup_meta_info, next_step, should_trigger_function_call=False
)
```

That costs the ability to call a *different* tool in the same turn, which seems an acceptable trade for a branch whose whole purpose is "stop calling tools and continue the conversation". If you'd rather keep `True`, a per-turn skip counter that bails after 2 would do.

tests/tests/test_tool_call_idempotency.py
```diff
@@ -0,0 +1,154 @@
...
+    tm = _make_task_manager(idempotent=True)
+    trigger = AsyncMock(return_value=OK_RESPONSE)
+    with _patchers(trigger):
+        await _run_tool(tm, mentioned_time="later")
+        await _run_tool(tm, mentioned_time="tomorrow 5pm")
+    assert trigger.await_count == 1
+
+
+@pytest.mark.asyncio
+async def test_failed_once_per_call_permits_retry():
```
san0808, line 134: Good test — this is the case that would have caught the old bug. Worth adding a sibling that asserts a **non-2xx** first call also permits a retry (`{"status_code": 500, "error": None}`), since that's a separate branch of `_tool_execution_succeeded` and the more common failure in practice than a timeout.

bolna/agent_manager/task_manager.py
```diff
@@ -2937,6 +2941,17 @@ async def _process_conversation_formulaic_task(self, message, sequence, meta_inf
                         )
                     )
 
+    def _is_once_per_call(self, called_fun):
+        """A tool the agent opted into firing at most once per call (tools_params[...].idempotent)."""
```
san0808, line 2945: The code reads `once_per_call` now, but three comments still name the old key: this docstring, `:343`, and `:3406` ("OPT-IN per tool via `tools_params[...].idempotent`").

Since `APIParams` uses `extra='ignore'`, anyone who follows these and sets `idempotent: true` gets a silent no-op, which is the exact failure this revision just fixed. Worth updating all three to `tools_params[...].once_per_call`.

PR #891, opened 2026-07-29
fix: make the reasoning summary request opt-in so it stops delaying the first token

san0808:
Every gpt-5 request sets `reasoning.summary = "auto"` unconditionally. The summary is emitted before the answer text, so on a live call the caller waits through it before hearing anything.

Measured at `effort=low` with a ~14K token prompt, holding everything but this field constant:

```
Azure  ptu-gpt-5.4-mini, HTTP SSE     summary=auto 2856 ms   omitted  766 ms
OpenAI gpt-5.4-mini, HTTP SSE         summary=auto 2034 ms   omitted 1128 ms
OpenAI gpt-5.4-mini, WebSocket        summary=auto 1539 ms   omitted 1043 ms
```

Reasoning token counts are unchanged either way (26 to 48 in both arms), so the model is not thinking more. The cost is emitting the summary first, and it scales with how much reasoning there is to summarise: gpt-5.4-mini and gpt-5.4 reason at `low` and show the penalty, while gpt-5.4-nano, gpt-5.5 and gpt-5.6-sol produce no reasoning tokens at `low` and show no measurable difference.

`reasoning_summary` now defaults to unset, which omits the field. Passing `reasoning_summary: "auto"` in the agent's llm config restores the previous behaviour, wired the same way as `reasoning_effort`. Both request paths go through `_build_responses_create_kwargs`, so this covers Azure and OpenAI, streaming and non-streaming.

Nothing functional reads the summary. It surfaces only as `reasoning_content` on execution log rows, and no `reasoning_content` appears in prod logs over the last six hours while `reasoning_tokens` appears 239 times. Reasoning token counts are unaffected, since `output_tokens_details.reasoning_tokens` is reported regardless of this setting.

The OpenAI figures include roughly 900 ms of network round trip from the machine running the measurement, so the relative penalty holds but the absolute floor is not representative.

### DIFF
--- a/bolna/llms/openai_base.py
+++ b/bolna/llms/openai_base.py
@@ -58,6 +58,10 @@ class OpenAICompatibleLLM(BaseLLM):
     _request_log_model = None
     _model_family = None
 
+    # A reasoning summary streams ahead of the answer text, so it is opt-in: requesting one costs
+    # time to first spoken token on a live call. Set from config by subclasses; None omits it.
+    reasoning_summary = None
+
     @property
     def request_log_model(self):
         """Provider-qualified model name, so request logs key the same way the task manager records."""
@@ -466,7 +470,8 @@ class OpenAICompatibleLLM(BaseLLM):
             reasoning_config = {}
             if reasoning_effort:
                 reasoning_config["effort"] = reasoning_effort
-            reasoning_config["summary"] = "auto"
+            if self.reasoning_summary:
+                reasoning_config["summary"] = self.reasoning_summary
             create_kwargs["reasoning"] = reasoning_config
             verbosity = self.model_args.get("verbosity")
             if verbosity:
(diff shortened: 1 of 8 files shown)


bolna/llms/openai_base.py
```diff
@@ -466,7 +470,8 @@ def _build_responses_create_kwargs(
             reasoning_config = {}
             if reasoning_effort:
                 reasoning_config["effort"] = reasoning_effort
-            reasoning_config["summary"] = "auto"
```
vikramsingh-bolna, line 469: earlier it was auto for all now if reasoning summary is missing then it will be blank, is it intended?

bolna/llms/openai_base.py
```diff
@@ -701,7 +706,8 @@ async def _generate_responses(self, messages, request_json=False, ret_metadata=F
             reasoning_effort = self.model_args.get("reasoning_effort")
             if reasoning_effort:
                 reasoning_config["effort"] = reasoning_effort
-            reasoning_config["summary"] = "auto"
+            if self.reasoning_summary:
```
vikramsingh-bolna, line 709: same as above

bolna/llms/openai_base.py
san0808: yes intended as this just enables reasoning models to summarize and store in reasoning_content that if people want can use later but the issue as mentioned in pr body this was never refered before and introduced unwanted latency and we donot use this at all durive live multi turn conv

note reasoning still works and happens during the call bas for debugging purpose for now we won't get content used which is good as delta is 3x and we also have option tomorrow if we want we can pass and be able to set for certain calls if needed and by default it is none

vikramsingh-bolna (approved):

PR #894, opened 2026-07-30
feat(cb): report the ASR/TTS model on provider health records

san0808:
ASR/TTS health records carried `model=None`, so the circuit breaker cannot see which model is degrading and per-model thresholds have nothing to key on. TTS first-byte latency varies up to 5.7x between models on the same provider (elevenlabs flash p95 131ms vs multilingual_v2 747ms).

Reports the live component model on the per-turn success hook and at all six `TranscriberError`/`SynthesizerError` raise sites. Both sides have to carry it: `_end_call_on_component_error` forwards `getattr(error, "model", None)`, so if only successes reported a model then successes and failures would land on different breaker members and the failure member would read 100% error indefinitely.

`_component_model` reads the live pool member, so a mid-call multilingual switch reports the model actually in use. Azure ASR has no model concept (it keys on recognition language) and reports None.

Observe-only, nothing reads breaker state for routing.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.166"
+__version__ = "0.10.167"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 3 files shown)


web1havv (approved):
LGTM

PR #895, opened 2026-07-30
fix: name TTS connect failures and bound the reconnect budget by wall clock

san0808:
Provider WebSocket connect failures were unreadable in the logs, and the retry budget gave up at the wrong times.

## Connect failures logged nothing

`logger.error(f"Failed to connect to X: {e}")` printed an empty string for a large share of real failures, because bare socket and TLS errors carry no args: `str(ConnectionResetError())` and `str(TimeoutError())` are both `""`. Switched to `{e!r}` plus `exc_info=True` on the generic handler, so the type, errno and origin reach the log:

```
Failed to connect to ElevenLabs: ConnectionRefusedError(61, "Connect call failed ...")
```

This is the point of the PR. Failures that produced a blank line were most likely a bare `ConnectionResetError()` raised by the TLS layer on EOF during the handshake, which fits what was observed: fast failures, no HTTP status. That cannot be confirmed until the type is actually logged, which is what this change enables.

## One connect deadline, owned by the library

`establish_connection` wrapped `websockets.connect` in `asyncio.wait_for(timeout=10.0)` while the library ran its own `open_timeout`, also defaulting to 10s. Two timers racing for the same job. Per review feedback the wrapper is gone and `open_timeout` owns the deadline, with the value in one shared `CONNECT_TIMEOUT_SECONDS` rather than repeated per provider. It covers DNS, the TCP connect and the opening handshake, and the library closes its own half-opened transport rather than relying on cancellation landing mid-handshake.

Timeout handlers now catch `(asyncio.TimeoutError, TimeoutError)`, because the class differs by version: on py3.10 `asyncio.TimeoutError` is a separate class from the builtin, and from 3.11 they are aliases. `open_timeout` itself raises `asyncio.TimeoutError`, which the original handler already caught.

## Retry budget

`MAX_CONNECTION_FAILURES = 3` with a flat `sleep(1)` and no backoff was wrong in both directions: a provider refusing instantly burned all three attempts in about 4s, and one hanging the connect held roughly 33s of dead air. Replaced with an attempt cap plus a wall-clock budget and jittered exponential backoff, where the budget stops early if another attempt as slow as the last would run past it.

| Per-attempt failure | Attempts | Dead air | Bound that fires |
|---|---|---|---|
| 0.3s | 6 | 6.1s | attempt cap |
| 1.0s | 6 | 10.0s | attempt cap |
| 2.0s | 5 | 12.3s | time budget |
| 10.0s | 1 | 10.0s | time budget |

Both bounds are load-bearing: the cap keeps fast refusals from hammering a struggling provider, the budget caps dead air when attempts are slow. Worst case is about 12s against 33s before, and fast failures get twice the chances to recover.

Also logs connect duration on every attempt, which gives the data to lower the connect deadline later without guessing, and the loop exits on `conversation_ended` instead of reconnecting into a finished call. The `connection_error` string is unchanged so existing log alerts keep matching.

`pixa_synthesizer.py` keeps its own copy of the retry loop because it extends `BaseSynthesizer` rather than `StreamSynthesizer`, so it does not inherit the budget change. It does get the connect and diagnostics fixes. Moving it onto `StreamSynthesizer` is a separate change.

### DIFF
--- a/bolna/synthesizer/cartesia_synthesizer.py
+++ b/bolna/synthesizer/cartesia_synthesizer.py
@@ -234,7 +234,7 @@ class CartesiaSynthesizer(StreamSynthesizer):
         try:
             start_time = time.perf_counter()
             websocket = await asyncio.wait_for(
-                websockets.connect(self.ws_url, ssl=get_ssl_context(self.ws_url)), timeout=10.0
+                websockets.connect(self.ws_url, ssl=get_ssl_context(self.ws_url), open_timeout=None), timeout=10.0
             )
             if not self.connection_time:
                 self.connection_time = round((time.perf_counter() - start_time) * 1000)
@@ -246,7 +246,7 @@ class CartesiaSynthesizer(StreamSynthesizer):
             else:
                 logger.info(f"Cartesia WebSocket connected connection_time={self.connection_time}ms")
             return websocket
-        except asyncio.TimeoutError:
+        except (asyncio.TimeoutError, TimeoutError):
             logger.error("Timeout while connecting to Cartesia websocket")
             return None
         except InvalidHandshake as e:
@@ -260,7 +260,7 @@ class CartesiaSynthesizer(StreamSynthesizer):
             self.connection_error = str(e)
             return None
         except Exception as e:
-            logger.error(f"Failed to connect to Cartesia: {e}")
+            logger.error(f"Failed to connect to Cartesia: {e!r}", exc_info=True)
             return None
 
     # ------------------------------------------------------------------
--- a/bolna/synthesizer/cartesia_synthesizer.py
+++ b/bolna/synthesizer/cartesia_synthesizer.py
@@ -9,7 +9,7 @@ import base64
 import websockets
 from websockets.exceptions import InvalidHandshake
 
-from .stream_synthesizer import StreamSynthesizer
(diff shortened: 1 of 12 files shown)


bolna/synthesizer/cartesia_synthesizer.py
```diff
@@ -234,7 +234,7 @@ async def establish_connection(self):
         try:
             start_time = time.perf_counter()
             websocket = await asyncio.wait_for(
-                websockets.connect(self.ws_url, ssl=get_ssl_context(self.ws_url)), timeout=10.0
```
vikramsingh-bolna, line 237: can we use this without asyncio.wait_for and using `open_timeout` ?

san0808: yes checked it out and ur idea made sense [7951d32](https://github.com/bolna-ai/bolna/pull/895/commits/7951d322d9390daf6e7f720e1997f9729e380794)

vikramsingh-bolna (approved):

bolna/synthesizer/cartesia_synthesizer.py
vikramsingh-bolna: test kr lena kuch break to nhi krega also internally yah library timeout me kon sa error deti h wo same hi h.

PR #915, opened 2026-07-31
feat: add Maya TTS synthesizer support

saicherry93479:
MayaSynthesizer adds Maya Research's Maya 2 Native as a TTS provider — eleven Indian languages including Indian English, with voices Ananya and Arjun, each covering all eleven. A conversation runs on one persistent WebSocket; a one-shot HTTP call covers prewarm clips.

- Sentence segmentation happens on Maya's side, so LLM fragments go out as they arrive with no client-side chunking
- Output is 24 kHz throughout — downsampled and mu-law encoded for telephony, left native for web
- ASR codes like `hi-IN` are trimmed to `hi`, which also lets language switch mid-call without reconnecting
- Interrupt sends a clear frame, then drops frames until `cancelled` lands
- A whitespace primer is added when the final push is empty, otherwise the flush returns nothing and the turn never closes
- Handoff and prewarm clips go over HTTP, with a direct mu-law path that skips the transcode
- Registered in enums, providers, models (MayaConfig) and README
- Unit tests included

### DIFF
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -112,11 +112,11 @@ class PixaConfig(BaseModel):
 class MayaConfig(BaseModel):
     # "Ananya" (female) or "Arjun" (male) — the only two voices, both speak every language.
     # Case-sensitive: Maya rejects "ananya" with a 400.
-    voice: str = "Ananya"
-    model: Optional[str] = "Maya 2 Native"
-    # One of hi/bn/gu/kn/ml/mr/or/pa/ta/te/en ("en" is Indian English). Region-qualified
-    # codes ("en-IN") are reduced to the primary subtag. Set null for code-switching text,
-    # so each script keeps its own pronunciation rules instead of being forced to one.
+    voice_id: str
+    voice: str
+    model: str
+    # One of hi/bn/gu/kn/ml/mr/or/pa/ta/te/en/auto. "en" is Indian English, "auto" lets Maya
+    # detect per utterance. Region-qualified codes ("en-IN") reduce to the primary subtag.
     language: Optional[str] = "en"
 
 
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -109,6 +109,17 @@ class PixaConfig(BaseModel):
     repetition_penalty: Optional[float] = 1.3
 
 
+class MayaConfig(BaseModel):
+    # "Ananya" (female) or "Arjun" (male) — the only two voices, both speak every language.
+    # Case-sensitive: Maya rejects "ananya" with a 400.
+    voice: str = "Ananya"
+    model: Optional[str] = "Maya 2 Native"
+    # One of hi/bn/gu/kn/ml/mr/or/pa/ta/te/en ("en" is Indian English). Region-qualified
+    # codes ("en-IN") are reduced to the primary subtag. Set null for code-switching text,
+    # so each script keeps its own pronunciation rules instead of being forced to one.
+    language: Optional[str] = "en"
+
+
 class AzureConfig(BaseModel):
     voice: str
     model: str
@@ -156,6 +167,7 @@ class Synthesizer(BaseModel):
         CartesiaConfig,
         DeepgramConfig,
         OpenAIConfig,
+        MayaConfig,
     ] = Field(union_mode="smart")
     stream: bool = False
     buffer_size: Optional[int] = 40  # 40 characters in a buffer
@@ -199,6 +211,9 @@ class Synthesizer(BaseModel):
         elif provider == "rime":
             if isinstance(config, dict):
                 values["provider_config"] = RimeConfig(**config)
+        elif provider == "maya":
+            if isinstance(config, dict):
+                values["provider_config"] = MayaConfig(**config)
 
         return values
 
--- a/bolna/synthesizer/maya_synthesizer.py
+++ b/bolna/synthesizer/maya_synthesizer.py
@@ -195,13 +195,15 @@ class MayaSynthesizer(StreamSynthesizer):
 
             await self._wait_for_ws()
 
+            # The wait above can span a barge-in, which retires this sequence without
+            # cancelling the task. Re-check before anything reaches the socket, including the
+            # flush branch: priming opens an utterance, so Maya would answer it with `end`.
+            if not self.should_synthesize_response(sequence_id):
+                logger.info(f"Not synthesizing (inner): sequence_id {sequence_id} not current")
+                await self.flush_synthesizer_stream()
+                return
+
             if text != "":
-                # The wait above can span a barge-in, which retires this sequence without
-                # cancelling the task. Re-check before anything reaches the socket.
-                if not self.should_synthesize_response(sequence_id):
-                    logger.info(f"Not synthesizing (inner): sequence_id {sequence_id} not current")
-                    await self.flush_synthesizer_stream()
-                    return
                 try:
                     if self.ws_send_time is None:
                         self.ws_send_time = time.perf_counter()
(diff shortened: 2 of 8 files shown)


bolna/synthesizer/maya_synthesizer.py
```diff
@@ -0,0 +1,394 @@
...
+from bolna.memory.cache.inmemory_scalar_cache import InmemoryScalarCache
+
+logger = configure_logger(__name__)
+
+MAYA_NATIVE_SAMPLE_RATE = 24000
+MULAW_SAMPLE_RATE = 8000
+
+MAYA_VOICES = {"Ananya", "Arjun"}
+# "en" is Indian English. Omit the field entirely for code-switching text.
+MAYA_LANGUAGES = {"hi", "bn", "gu", "kn", "ml", "mr", "or", "pa", "ta", "te", "en"}
```
san0808: `language="auto"` raises here, so Maya's per-utterance detection cannot be configured at all. The raise happens when the synthesizer is constructed at call time, so the call fails with no audio rather than failing at save time. `set_target_language("auto")` gets dropped too.

Maya does support it. Its own rejection frame advertises it: `available_languages: ["hi","bn",...,"en","auto"]`, and I confirmed against the live API that it renders for both romanized and mixed-script text.

```suggestion
MAYA_LANGUAGES = {"hi", "bn", "gu", "kn", "ml", "mr", "or", "pa", "ta", "te", "en", "auto"}
```

also would be better if we can move maybe `MAYA_VOICES` and `MAYA_LANGUAGES` to bolna/constants.py ?

bolna/models.py
```diff
@@ -109,6 +109,17 @@ class PixaConfig(BaseModel):
...
 
+class MayaConfig(BaseModel):
+    # "Ananya" (female) or "Arjun" (male) — the only two voices, both speak every language.
+    # Case-sensitive: Maya rejects "ananya" with a 400.
+    voice: str = "Ananya"
+    model: Optional[str] = "Maya 2 Native"
+    # One of hi/bn/gu/kn/ml/mr/or/pa/ta/te/en ("en" is Indian English). Region-qualified
+    # codes ("en-IN") are reduced to the primary subtag. Set null for code-switching text,
+    # so each script keeps its own pronunciation rules instead of being forced to one.
+    language: Optional[str] = "en"
```
san0808: `voice_id` is dropped here. The synthesizer takes it as an alias for `voice`, but this model has no such field:

```
Synthesizer(provider='maya', provider_config={'voice_id':'Arjun', ...}).provider_config
-> voice='Ananya' model='Maya 2 Native' language='en-IN'
```

The dashboard writes `voice_id` for every other provider, so an agent set to Arjun would speak as Ananya with nothing logged. Sarvam, Cartesia, Rime, Smallest, Pixa and Deepgram all require both `voice_id` and `voice`. Match that shape?

```suggestion
    voice_id: str
    voice: str
    model: str
    # One of hi/bn/gu/kn/ml/mr/or/pa/ta/te/en/auto. "en" is Indian English, "auto" lets Maya
    # detect per utterance. Region-qualified codes ("en-IN") reduce to the primary subtag.
    language: Optional[str] = "en"
```

bolna/synthesizer/maya_synthesizer.py
```diff
@@ -0,0 +1,394 @@
...
+        try:
+            if self.conversation_ended:
+                return
+            if not self.should_synthesize_response(sequence_id):
+                logger.info(f"Not synthesizing: sequence_id {sequence_id} not current")
+                return
+
+            await self._wait_for_ws()
+
+            if text != "":
```
san0808, line 206: The sender can sit in `_wait_for_ws()` across a barge-in, and nothing cancels it: `__cleanup_downstream_tasks` calls `handle_interruption()` but leaves the sender task running. It then reaches the `end_of_llm_stream` branch and resets `_discard_audio` to `False`, sends the primer and flush, and Maya answers with `end`, so a retired sequence emits an end-of-stream sentinel.

Deepgram, ElevenLabs and Rime all re-check `should_synthesize_response` after the wait for this reason. Add the same check here, and leave `_discard_audio` alone in the `end_of_llm_stream` branch?

```suggestion
            await self._wait_for_ws()

            if text != "":
                if not self.should_synthesize_response(sequence_id):
                    logger.info(f"Not synthesizing (inner): sequence_id {sequence_id} not current")
                    await self.flush_synthesizer_stream()
                    return
```

bolna/synthesizer/maya_synthesizer.py
```diff
@@ -0,0 +1,401 @@
...
+    # ------------------------------------------------------------------
+
+    def _config_message(self):
+        payload = {"type": "config", "voice": self.voice, "model": self.model}
+        if self.language:
+            payload["language"] = self.language
+        return payload
+
+    def form_payload(self, text):
+        """A fragment. `flush` is Maya's only terminator, so text frames are never final."""
```
san0808, line 139: Maya generates nothing until it receives the flush, so with `flush: False` on every text frame the TTS does not start until the LLM has finished the whole response.

Measured on the live socket: I sent one complete sentence and held the flush. Zero audio for 4s. First audio arrived 80ms after the flush and covered everything.

```
sent sentence 1 (no flush), listening 4s...
  audio before flush: 0B
  after flush: first audio +0.08s, 134400B (2.80s of speech)
```

ElevenLabs, Cartesia, Deepgram and Rime all overlap synthesis with the LLM stream, so this costs roughly the LLM's remaining stream time in extra time-to-first-audio on every turn. The PR description and commit message both say Maya segments server-side and "starts speaking at the first one", which does not match what the socket does.

Is there an incremental mode in their API that we are missing? If not, per-sentence flush using `text_chunker` would fix the latency, but each flush draws its own `end`, so `receiver` would need to forward only the last one as the end-of-stream sentinel. Worth measuring time-to-first-audio on a live call before merging either way.

bolna/synthesizer/maya_synthesizer.py
```diff
@@ -0,0 +1,401 @@
...
+
+    async def handle_interruption(self):
+        """Already-generated frames keep arriving until Maya answers with `cancelled`, so
+        audio is dropped over that window."""
+        try:
+            if not self._is_ws_connected():
+                return
+            self._discard_audio = True
+            self._turn_has_text = False
+            await self._send_json({"type": "clear"})
```
san0808, line 176: `clear` cancels generation but does not purge text that has not been flushed yet, so an abandoned response is spoken at the start of the next reply. Measured on the live socket:

```
long alone (flushed)                        4.38s
short alone (flushed)                       0.68s
long unflushed, then short + flush          4.88s   <- old text spoken
long unflushed, CLEAR, then short + flush   4.90s   <- clear made no difference
```

Two ways to reach it. A barge-in landing while the LLM is still streaming, and the `speech_final` path in `task_manager` that cancels the LLM task and calls `invalidate_pending_responses()` without touching the synthesizer at all, so no clear is even sent. The caller hears the interrupted answer prefixed onto the next one, and because the audio carries the new turn's `meta_info`, nothing in the transcript or logs shows it.

The new guard in `sender` widens this slightly: returning before the flush is right for not emitting a stray end-of-stream, but it leaves the buffer dirty.

A flush does drain it, verified: after `text, clear, flush` the drain produced 4.38s of audio plus `end`, and the next turn was back to 0.64s. `_turn_has_text` is already the "buffer is dirty" signal, so `handle_interruption` has what it needs, though a drain would also need `receiver` to swallow that `end`. Your call on the shape, but I do not think this can ship as is.

san0808 (approved):
LGTM

PR #919, opened 2026-08-03
Lid/v2

puneet-bolna:
LID v2: judge hardening, Bedrock migration, and switch-flow race fixes
This PR hardens the LLM-driven language-switch flow end to end — judge model, decision gates, playback timing, detector observability — based on ~40 QA calls root-caused on topaz. All changes are already running on topaz (bolna 0.10.173) and validated against live traffic.

Judge on Bedrock (latency)
bedrock/ model support in LanguageSwitcher: IAM-role auth (no API key), prompt caching via cache_control → cachePoint (scoped to Claude ids), and an in-code region default (ap-south-1, AWS_REGION overrides) so boxes need no aws config.
Runtime fallback: Bedrock permission/throttle errors only surface at invoke time — after 2 consecutive decide failures the judge swaps to the Anthropic-key default mid-call. A parsed null reply is a valid decision, not a failure, and resets the streak.
Measured (100-request benchmark, real prompts, caching on): p50 1.2 s / p99 2.25 s vs Anthropic-direct p50 1.5 s / p99 5.8 s (which was brushing the 6 s decide timeout).
Decision-quality gates
Substance gate measures foreign segments (per-target evidence), not the buffer-lifetime max — a stale active-language segment can no longer carry a short mis-tagged fragment past the gate. Gate at 0.8 s.
RECENT TURNS integrity: durations come only from segments actually tagged with the detected language (no borrowed-duration en(2.5) built from hi audio); entries carry →xx markers on switch firings so pre-switch drift reads as stale; rule 3a (digits/codes) now has absolute precedence over rule 8 drift.
Stale-decision guard keeps the detector buffer (drops only the decision) — a caller's explicit "Can you speak in English?" is no longer deleted mid-switch.
Timing/race fixes
Idle-flush suppression: the watcher no longer fires mid-utterance — deferred while main-ASR interims show the caller speaking, with a 2.5 s stale-flag escape (buffer age doubles as the activity clock). Kills the one-utterance→two-decides slicing behind most "switch came a turn late" reports.
Playback gate: armed from drained evidence inside the decide (spawn-time arming raced the idle-flush drain and let old-language replies play), stale done-task gates retired with telemetry, and on a switch the gate now holds until the sequence is invalidated — closing a 50 ms window where held old-language audio could ship between release and truncate.
Eager (Flux) path passes eager_meta_info so gates arm with the sequence the reply actually plays under.
Detector (unbiased ASR)
sarvam_model config override (saaras:v4 via the dashboard feature flag; default v3 unchanged).
detector_health telemetry: a silent detector previously wrote zero records and looked identical to a quiet call — now a per-call record (chunks fed, segments received, unknown frames) is emitted when the tap produced nothing while the caller spoke, and it's flushed before the task_output snapshot so it actually persists.
Undrained buffer age-bounded (30 s) so skipped generations can't lend duration/text to later turns; Sarvam VAD events frames excluded from unknown-frame accounting.
Prompt
Rules 4a (unstable tags vs stable LIVE script), repeated-bare-name explicit requests, embedded-vs-whole-turn code-mixing test, LIVE-unavailable marker for idle flushes, and the precedence/staleness sentences above.
Language directives unified across simple and graph agents (carve-out for proper nouns/codes; one-shot restate lives only in the switch-time note).
Testing
~90 new tests across 15 files (gate lifecycle race, hedge ownership, fallback matrix, watcher suppression branches, buffer eviction, prompt pins). Full LID suite: 290 passing. Live verification on topaz per change: gate telemetry (decided/expired + held_ms), decide latency, and detector health are all queryable in lid_shadow_events.

Rollout notes
No new env vars. Bedrock enables via existing LANGUAGE_SWITCH_LLM; per-agent tools_config overrides (Azure flag) still win. Prod needs only the Bedrock invoke IAM policy.
Legacy flow untouched; the switch_language tool injects only when the new flow is off or the judge has no credentials.


bolna/agent_manager/task_manager.py
```diff
@@ -4946,57 +5096,25 @@ def __detector_language_mismatch(self) -> bool:
...
         if not isinstance(pool, TranscriberPool):
             return False
-        detected = pool.lid_buffer_language()
         active_short = (self.language or "").split("-")[0].lower()
-        if not detected or detected == active_short or detected not in pool.labels:
+        # Read the WHOLE buffer, exactly as the idle watcher does. Reading only the newest tag
+        # made a foreign turn whose tail fragment is mis-tagged as the active language look like
+        # no mismatch at all, so its audio was never gated and got truncated mid-sentence.
+        _, foreign_langs = self.__buffered_language_evidence(pool, active_short)
+        detected = next((lang for lang in foreign_langs if lang in pool.labels), None)
```
san0808, line 5244: The foreign tag can now come from any segment, but the substance check below is still `lid_buffer_max_segment_seconds()`, which is the max over *all* segments. So a 3s active-language turn carrying a 0.15s fragment mis-tagged as another language passes both checks and arms the playback gate, and rule 6 then correctly answers "stay". Is that intended? Previously only the newest tag counted, so a turn ending in the active language armed nothing.

On a code-mixed turn that opens or middles on a borrowed English word and ends in the active language, the caller would now pay the difference between first-audio-ready and decide-complete, roughly 0.7-2.1s of extra silence, on a turn where nothing changes. Should the *foreign* segment itself have to clear `min_segment_s`, the way `__detector_corroborates` already scans per segment? Its docstring makes exactly this argument about not combining aggregates that describe different segments.

bolna/agent_manager/task_manager.py
```diff
@@ -4924,6 +4912,163 @@ def __language_switch_enabled(self) -> bool:
...
+    @staticmethod
+    def __recent_detected_turns(pool, limit: int = 4) -> list:
+        """(detected_language, longest_segment_s) for the last few Switch-LLM firings, oldest
+        first — the cross-turn evidence the judge needs to spot sustained drift. Read from the
+        telemetry we already append per firing, so it costs nothing extra to collect."""
+        # Filter THEN slice: handoff and legacy records share this list, so slicing first let
+        # them displace real turns — right after a switch the tail is all handoff records and the
+        # judge got "(none)" exactly when drift evidence matters most.
+        turns = [
+            (r.get("detected_language"), r.get("buffered_max_segment_s") or 0.0)
```
san0808: Two things about what this feeds rule 8:

`detected_language` is recorded on `stay` and `gated:*` firings too, so a turn where the judge correctly decided "caller is speaking B, but this is transliteration / digits / one borrowed word, stay" comes back as evidence of B. Rule 8 then says two or more substantive B turns in a row means the agent got it wrong earlier and it should prefer switching. Does that let a consistent judge-level error corroborate itself into a switch, which is the class the LIVE marker was added to prevent?

And `buffered_max_segment_s` is the buffer-wide max, not the duration of the segments tagged with that language, so the same one-borrowed-word turn is reported as around 2s and reads as "real speech" under rule 8's duration guard. Would carrying `outcome`/`target_language` per entry, and taking the duration from the detected language's own segments, keep the guard meaningful?

bolna/agent_manager/task_manager.py
```diff
@@ -5399,7 +5537,24 @@ def as_float(value):
 
         if any(activity.values()):
             logger.info(f"LanguageSwitcher: truncating in-flight old-language response before switch ({activity})")
+            # Truncating wipes the mark dict, so the final-chunk ack that would clear this flag
+            # never arrives — clear it here as the barge-in path does, or it latches True and
+            # blocks the silence prompt and the stall backstop for the rest of the call.
+            if "input" in self.tools:
+                self.tools["input"].update_is_audio_being_played(False)
```
san0808, line 5754: Should the playback gate be released explicitly here, before `__cleanup_downstream_tasks()`?

As it stands the only release sites are inside `__lid_playback_gate_holds`, whose only caller is the output loop under `status == "SEND"`. Cleanup invalidates the pending sequence, swaps `buffered_output_queue` and cancels/restarts `output_task`, and the follow-up gets a fresh `sequence_id`, so that gate is never re-polled: it stays armed until the next `__arm_lid_playback_gate` silently overwrites it, and no `playback_gate` record is written.

That means the confirmed-switch case, the drop half of the played/dropped ratio the release docstring argues for, is exactly the one with no telemetry. The records that do land would be almost entirely `outcome=decided` from turns that stayed.

bolna/helpers/language_switcher.py
```diff
@@ -108,45 +115,128 @@ async def _warm():
...
+        1.3-2.7s, observed tail 5.9s), not a slow model — a fresh request usually beats the
+        straggler. Bounds caller-visible silence without raising the decide timeout, and both
+        requests read the same cached prefix. 0 disables (single request)."""
+        hedge_after_s = float(os.getenv("LANGUAGE_SWITCH_HEDGE_AFTER_S", str(DEFAULT_HEDGE_AFTER_S)))
+        self.hedge_won = False  # per-decide; without the reset it stays True for the rest of the call
+
+        async def attempt():
+            return self._parse_json(await self._llm.generate(messages))
+
+        first = asyncio.create_task(attempt())
```
san0808: `first` is created here, but the `try`/`finally` that cancels the losers starts below, after the hedge wait. If `decide()` is cancelled during `return await first` or during `asyncio.wait({first}, timeout=hedge_after_s)`, for instance on call teardown cancelling the `handle_language_switch` task, `first` keeps running unowned and its exception is never retrieved. Worth moving the task creation inside the `try` so the existing `finally` covers that window?

bolna/lid/base.py
```diff
@@ -211,13 +211,21 @@ def buffer_language(self):
...
+        deliberately — it is the highest-precision signal the detector produces (N same-language
+        confirmations), so it stays available if the fast path is ever re-enabled.
+        """
         return self._buffer_lang_streak
 
     def buffer_language_confidence(self):
         """Language probability of the current buffer_language (peek, no drain).
 
-        None when the provider does not return a language score (e.g. Soniox).
+        Sarvam reports a per-utterance score; Soniox has none, so it derives one from the
```
san0808: This says Soniox derives a score from the winning language's token share, but `soniox.py` in this same PR deliberately hardcodes `prob = None` with a comment explaining why it must not. Leftover from an earlier iteration of the branch? As written it will convince the next reader that detector corroboration works on Soniox.

bolna/agent_manager/task_manager.py
```diff
@@ -4935,9 +5080,14 @@ def _spawn_language_switch_decision(self, transcriber_message: str, meta_info: d
             return None
         snapshot = dict(meta_info)
         self._last_turn_meta_info = snapshot
-        return asyncio.create_task(
+        decision_task = asyncio.create_task(
             self.handle_language_switch(transcriber_message, snapshot, spawn_language=self.language)
         )
+        # Arm here, not at the call sites: the eager (Flux) path spawns the decision too, and
+        # arming only at the turn boundary left every eager turn playing the old-language reply.
+        if self.__detector_language_mismatch():
```
san0808, line 5228: Arming here rather than at the call sites makes sense. Given that, does the eager branch's mismatch skip ("Skipping speculative LLM: detector language mismatch - reply holds for the switch decision") still earn its keep? The reply no longer holds and eager audio would be gated anyway, so the skip now just costs eager latency on mismatch turns, including the ordinary code-mixed ones the widened check picks up. The comment is stale either way.

bolna/agent_manager/task_manager.py
```diff
@@ -789,8 +792,14 @@ def __init__(
...
         self.agent_names = self.task_config.get("tools_config", {}).get("agent_names") or {}
-        # Injected for BOTH flows — in the LLM-driven flow it's the fast path for explicit requests.
-        self.__inject_switch_language_tool()
+        # LEGACY FLOW ONLY, matching the design comment above: with the Switch LLM enabled the
+        # judge is the single switching authority. Injecting the tool alongside it made the main
+        # LLM a second, competing switcher deciding from main-ASR text — which mis-scripts foreign
+        # speech precisely when switching matters (QA 5765dd9f: tool switched to 'ta' from
+        # Tamil-rendered text while the unbiased detector heard 'te'), and its silent history-side
+        # races produced unexplained "Already speaking in X" tool responses in every QA round.
+        if not self.__language_switch_enabled():
```
san0808: This gates on the config flag rather than on a judge that actually works. `LanguageSwitcher.__init__` logs an error and constructs anyway when no key resolves, so a flagged agent with a bad key or a provider outage now has no switch path at all, including for an explicit "speak English please". Should the tool be re-injected when no key resolved, or after N consecutive decide failures?

bolna/agent_manager/task_manager.py
```diff
@@ -5115,25 +5220,24 @@ async def __lid_idle_watcher(self):
...
-                # cost of one extra LLM call, not a wrong switch.
-                streak_fired = mismatch and pool.lid_buffer_language_streak() >= mismatch_streak_fire
-                if streak_fired:
-                    threshold = 0.0
-                if age >= threshold:
+                # An all-active-language buffer can only produce "stay", so firing would spend
+                # ~1.5-2s of decide (and the lock it holds) on a foregone conclusion. Skipping
+                # deliberately does NOT drain: if the main ASR later delivers this turn, the
+                # turn-boundary decide still sees the full transcript.
+                nothing_to_decide = saw_tags and not mismatch
```
san0808, line 5391: The skip deliberately does not drain, and `_buffer_max_segment_s` is a monotonic max that only resets on a drain. So while a non-empty all-active-language buffer sits there with no main turn arriving, this logs every 2s (`remaining = 2.0`, the event wait times out, `age` is still past threshold) and the substance gate stays satisfied by the stale lifetime max. Should the skip drain once past some age, or at least stop re-logging?

bolna/agent_types/graph_agent.py
```diff
@@ -998,16 +999,40 @@ def get_node_by_id(self, node_id: str) -> Optional[dict]:
...
+            lang_name = LANGUAGE_NAMES.get(detected_lang, detected_lang)
+            directive = (
+                f"\n\nLANGUAGE GUIDELINES\n\nPlease make sure to generate replies ONLY in "
+                f"{lang_name} ('{detected_lang}') — every reply must be in {lang_name}, "
+                f"regardless of the language this prompt or the earlier conversation is written in. "
+                f"This overrides EVERY other language setting in this prompt: preferred-language "
+                f"variables, per-language scripted questions or sample responses, and instructions "
+                f"to speak 'as per' any language preference. When this prompt provides the same "
+                f"script in multiple languages, use the {lang_name} version; when it only exists "
+                f"in another language, translate it into {lang_name} and keep the meaning exact."
```
san0808: "when it only exists in another language, translate it into X and keep the meaning exact" now reaches every multilingual graph agent, since neither this nor `switch_language`'s unconditional `__apply_language_directive` is gated on `llm_language_switch`. For agents whose node prompts deliberately script mixed-language content (addresses read in English, brand names, scripted legal lines) this instructs translation. Should the directive carve those out, given the switch prompt's own rule 3a treats digits and Latin letters as language-neutral?

puneet-bolna: resolved

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/lid/base.py
puneet-bolna: resolved

bolna/helpers/language_switcher.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py
```diff
@@ -5002,9 +5170,14 @@ def _spawn_language_switch_decision(self, transcriber_message: str, meta_info: d
...
         snapshot = dict(meta_info)
         self._last_turn_meta_info = snapshot
-        return asyncio.create_task(
+        decision_task = asyncio.create_task(
             self.handle_language_switch(transcriber_message, snapshot, spawn_language=self.language)
         )
+        # Arm here, not at the call sites: the eager (Flux) path spawns the decision too, and
+        # arming only at the turn boundary left every eager turn playing the old-language reply.
+        if self.__detector_language_mismatch():
+            self.__arm_lid_playback_gate(snapshot.get("sequence_id"), decision_task)
```
san0808, line 5229: Following up on the eager skip removal, because this arming does not actually reach the eager path and I think removing the skip left it exposed.

The eager EndOfTurn branch calls `_spawn_language_switch_decision(transcriber_message, meta_info)` with the loop-scoped `meta_info = message["meta_info"]`, and that branch (unlike the `else` two lines below it) never runs it through `__get_updated_meta_info`. Raw transcriber meta_info has no `sequence_id`: the input handlers build it as `{"io", "type", "sequence"}` and the transcriber only adds `request_id`, `transcriber_duration`, `user_stop_offset_ms`, `user_stop_ts_wall`. So `snapshot.get("sequence_id")` is None and `__arm_lid_playback_gate` returns on its first line. The eager reply's audio meanwhile carries `self.eager_meta_info["sequence_id"]`, the id `revalidate_sequence_id` re-adds a few lines earlier.

Net effect: the gate is never armed on eager turns, which is the case this comment says the move into `_spawn_language_switch_decision` was meant to cover. That was harmless while the eager branch skipped generation on mismatch, since no eager reply existed. Now the eager reply is generated, sits in the output loop's WAIT branch while `callee_speaking` is True, and is released the moment `on_user_speech_ended` fires, in the old language, then truncated by the switch.

Passing `self.eager_meta_info` at that call site should be enough: it is still populated there and cleared two lines later, and there is no real suspension point between `on_user_speech_ended` and the spawn (`_report_component_health` is fire-and-forget, `collect_transcript` returns or spawns a task, `_maybe_update_tts_language` returns early unless the synth exposes `set_target_language`), so a correctly keyed gate does hold the queued audio.

Only bites agents with `eager_eot_threshold` set, which is Deepgram Flux only, so it is config gated rather than every LID call.

bolna/agent_manager/task_manager.py
san0808: The duration half is fixed, thanks. The other half is still open: `__recent_detected_turns` filters on `flow` and `detected_language` only, so `stay` and `gated:*` firings still reach rule 8. A turn the judge deliberately refused as transliteration, digits or a single borrowed word still comes back as evidence of that language, and rule 8 reads two of them in a row as "the agent got it wrong earlier, prefer switching". Carrying `outcome` or `target_language` per entry would let the rule tell a refusal apart from a missed switch.

bolna/agent_manager/task_manager.py
san0808: `skip_logged` handles the log repeat. The part that still stands: the skip does not drain, and `_buffer_max_segment_s` only resets on a drain, so `__run_language_switch` still reads `buffered_max_segment_s = pool.lid_buffer_max_segment_seconds()` for its substance gate. That is the same buffer-wide max you just replaced with `foreign_max_s` in `__detector_language_mismatch`, so a stale inflated max can still carry a short foreign fragment past the decide's own gate. Worth making the two consistent?

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py
```diff
@@ -5765,6 +5899,47 @@ def __handoff_clip_to_mulaw(self, synth, audio):
...
+            f"## Language note:\nThe user is now speaking {name} ('{label}'). From this point onward, "
+            f"respond only in {name}, regardless of the language used earlier in the conversation or "
+            f"elsewhere in this prompt. This instruction overrides all other language preferences, "
+            f"language-selection rules, and multilingual script variants. "
+            f"Immediately repeat the agent's last line before the language switch in {name} so the "
+            f"caller can follow, then continue the conversation from that point. "
+            f"For the remainder of the call, use only the {name} version of every question, FAQ, "
+            f"sample response, objection-handling response, and closing line. If a {name} version "
+            f"is not provided, translate the available version into clear, natural {name} while "
+            f"preserving its exact meaning."
```
san0808: The verbatim carve-out has gone missing from this directive. The reformat in `8646a2dc` dropped the closing sentence ("Never translate or alter proper nouns, brand names, alphanumeric identifiers, digits, codes, or lines these instructions mark as verbatim/legal"), while `06bff336` kept it in the graph-agent copy. So the two directives have diverged, and the one that loses it is the one pinned to every multilingual agent's system prompt, including legacy tool-driven ones, since `__apply_language_directive` is not gated on `llm_language_switch`.

"If a version is not provided, translate the available version" is now unqualified here, which is what the carve-out existed to bound: addresses read in English, brand names, order and reference codes. Was the removal deliberate, or lost in the reformat?

bolna/agent_manager/task_manager.py
```diff
@@ -5765,6 +5899,47 @@ def __handoff_clip_to_mulaw(self, synth, audio):
...
+        same "## Language note:" header as __switch_context_note so replacement logic
+        can find and strip whichever variant is currently installed.
+        """
+        name = LANGUAGE_NAMES.get(label, label)
+        return (
+            f"## Language note:\nThe user is now speaking {name} ('{label}'). From this point onward, "
+            f"respond only in {name}, regardless of the language used earlier in the conversation or "
+            f"elsewhere in this prompt. This instruction overrides all other language preferences, "
+            f"language-selection rules, and multilingual script variants. "
+            f"Immediately repeat the agent's last line before the language switch in {name} so the "
```
san0808: This sentence assumes a switch just happened, but `__language_directive` is also what `load_prompt` installs as the STARTING-language pin, where nothing has switched and the agent has no last line. The opener is then told to repeat a line that does not exist. Should the repeat instruction live in `__switch_context_note` (which is only built on an actual switch) rather than in the generic directive?

Separately, on a real switch this stacks: `__play_switch_handoff` already speaks a handoff clip in the new voice to cover the generation gap, then this asks the model to restate the previous line, then the follow-up answers the caller's turn. That is three utterances before the caller gets their answer. Was that sequence walked through on a live call?

bolna/helpers/language_switcher.py
```diff
@@ -55,19 +67,28 @@ def __init__(self, available_labels, run_id=None, model=None):
...
         self.latency_ms = None
+        self.hedge_won = False  # last decide was answered by the hedged request, not the first
         # Dedicated creds, NOT the agent's — an Azure/OpenAI agent would 404 the switch model.
         switch_llm_key, switch_llm_base, switch_llm_version = resolve_switch_llm_credentials(self.model)
         # A configured (e.g. azure) judge with no resolvable key would fail EVERY decide,
         # leaving switching inert for the flagged org. Fall back to the default judge, which
         # switching depends on, rather than shipping a dead judge.
         default_model = f"anthropic/{DEFAULT_LANGUAGE_SWITCH_LLM}"
-        if not switch_llm_key.strip() and self.model != default_model:
+        self._is_bedrock = self.model.startswith("bedrock/")
```
san0808, line 79: A Bedrock judge silently loses prompt caching. `_system_message` attaches `cache_control` only when the model starts with `anthropic/` or `claude`, and a litellm Bedrock id (`bedrock/anthropic.claude-...`) matches neither. So the whole rules block, which just grew with rules 4a and 8, is reprocessed on every decide and again on every hedged second request. `_hedged_generate`'s docstring says "both requests read the same cached prefix", which stops holding the moment the judge is Bedrock, and the hedge doubles the cost of the miss.

Anthropic on Bedrock does support caching, so is this just the prefix check needing to cover `bedrock/` too? Same question for `prewarm()`, which is described as seeding the prompt cache and would only be paying the TLS handshake there.

bolna/lid/sarvam.py
```diff
@@ -116,7 +116,12 @@ async def _receiver_loop(self):
             async for raw in self._ws:
                 try:
                     data = json.loads(raw) if isinstance(raw, str) else {}
+                    if data.get("type") != "data":
+                        # Error/quota/VAD frames were dropped silently, hiding a dead tap.
+                        self.unknown_frames += 1
+                        logger.warning(f"SarvamLID: non-data frame {str(raw)[:200]}")
```
san0808: This warns once per non-data frame with no bound, and VAD or event frames are per-utterance traffic rather than rare. The same commit solves exactly this shape one file over with `_lid_feed_error_logged` in `transcriber_pool` ("was debug: a feed raising every chunk silently killed switching"), where the fix was log-once rather than log-always. Worth the same treatment here, or counting into `unknown_frames` and emitting one summary at cleanup next to the detector-health record?

bolna/agent_manager/task_manager.py
```diff
@@ -5241,18 +5381,17 @@ async def __run_language_switch(
...
-        # decision and discard the stale buffered speech with it.
+        # Stale-decision guard: the language changed while this decision waited on the lock,
+        # so its LIVE transcript is mislabeled (it came from the PRE-switch recognizer) — that
+        # mislabeling caused the mr→hi→mr ping-pong in QA 1a16da82. Drop the DECISION only.
+        # The detector buffer is kept: the tap runs language-code=unknown, so its speech means
+        # the same before and after a switch, and draining it here deleted a caller's explicit
+        # "Can you speak in English?" mid-switch, costing a repeat + ~28s (QA 971254c0).
         if spawn_language is not None and spawn_language != self.language:
-            discarded, _ = pool.take_lid_transcript()
+            retained = pool.lid_buffer_age()
```
san0808, line 5473: Keeping the buffer makes sense for the explicit-request case, but what bounds the tail?

The segments retained here are the ones that landed while the winning decide was running, which is typically the tail of the caller's PRE-switch utterance in the old language. That tail then gets re-judged against the NEW active language on the next firing, which is a route back to the ping-pong this guard exists to prevent. The substance gate does not bound it either: a mid-sentence tail clears 0.7s easily.

Would it be worth keeping the buffer only when it carries something the switch could not have already consumed (an explicit request, or segments timestamped after the switch applied) rather than unconditionally?

bolna/helpers/language_switcher.py
```diff
@@ -108,45 +129,158 @@ async def _warm():
...
             },
         ]
+        start_time = time.time()
         try:
-            start_time = time.time()
-            # Needs a user message (system-only list breaks litellm); prompt mandates JSON.
-            response = await self._llm.generate(messages)
+            result = await self._hedged_generate(messages)
+            if result is None:
+                self._note_failure()
```
san0808: This counts a valid decision as a judge failure. `_parse_json` runs `json.loads` on a reply of literal `null` and returns None, which is the model saying "no target", not a broken judge. Two of those in a row hit `RUNTIME_FALLBACK_AFTER` and swap the judge for the rest of the call. Should `_note_failure` be driven off the exception path only, with a None-but-no-exception result treated as a successful decide that happened to return nothing?

puneet-bolna: resolved

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/lid/sarvam.py
puneet-bolna: resolved

bolna/helpers/language_switcher.py
puneet-bolna: resolvwed

bolna/agent_manager/task_manager.py
puneet-bolna: resolvewd

bolna/agent_manager/task_manager.py

san0808 (approved):
LGTM

bolna/agent_manager/task_manager.py
```diff
@@ -5010,6 +5006,183 @@ def __language_switch_enabled(self) -> bool:
...
+            # one-borrowed-word turn reads as ~2s of "real speech" under rule 8's duration guard.
+            detected_short = (r.get("detected_language") or "").split("-")[0].lower()
+            segment_durations = [
+                float(seg.get("audio_s") or 0.0)
+                for seg in r.get("detector_segments") or []
+                if (seg.get("lang") or "").split("-")[0].lower() == detected_short
+            ]
+            if segment_durations:
+                return max(segment_durations)
+            return r.get("buffered_max_segment_s") or 0.0
```
san0808: When no segment carries the judge's `detected_language`, this falls back to `buffered_max_segment_s`, which is the max over all segments including the active-language ones. That is exactly the case where the detector never heard the detected language, so a Hindi turn the detector tagged `hi` throughout, with the judge reporting `en`, reaches rule 8 as `en(2.5)` and clears its "entries around 1s or more are real speech" guard on duration no `en` segment ever contributed. Should it return 0.0 when there is no matching segment?

bolna/prompts.py
```diff
@@ -105,43 +105,46 @@
...
 7. If the dominant spoken language is NOT in the supported list, or you are unsure, stay (target_language = null).
+8. SUSTAINED DRIFT ACROSS TURNS IS A SWITCH — rule 6 judges THIS turn in isolation, which is right for one borrowed word but wrong when the caller has already moved. RECENT TURNS lists earlier turns as `lang(longest-segment-seconds)`. If the caller has produced language B on the recent turns AND this turn is also B, that repetition IS the substantive matrix rule 6 asks for, even when each turn alone looks short — switch to B (if supported). Weigh it by duration: entries around 1s or more are real speech; strings of sub-second entries are the acknowledgment mis-tags rule 6 describes and are NOT evidence, however many there are. Two or more substantive B turns in a row while the agent stayed in A means the agent got it wrong earlier — prefer switching over staying. This is also how a caller corrects a bad switch without naming a language: they simply keep answering in B.
```
san0808: `__recent_detected_turns` keeps only `(detected_language, duration)` per firing, so a turn the judge refused under rule 3a (digits) or rule 6 (one borrowed word) is indistinguishable in RECENT TURNS from an ordinary stay. Three order-number readouts in a row on a Hindi call then hand this rule `en(2.0), en(2.0)` with a current turn that also reads en, and rule 3a and rule 8 say opposite things about the same input with no stated precedence. Should the line carry the outcome so the judge can see those were refusals, or should rule-3a/rule-6 stays be filtered out of the feed?

bolna/agent_manager/task_manager.py
```diff
@@ -5473,6 +5682,11 @@ def as_float(value):
 
         # Truncate the in-flight old-language reply (barge-in cleanup) before switching.
         activity = self._inflight_response_activity()  # captured pre-truncation for telemetry
+        # Release the gate before cleanup: cleanup invalidates this sequence, so the output
+        # loop never re-polls it — the switched case wrote no playback_gate record at all.
+        held_gate = self.lid_playback_gate
+        if held_gate is not None:
+            self.__release_lid_playback_gate(held_gate, "decided")
```
san0808: Releasing here leaves the gate None across `output.handle_interruption()`, `synthesizer.handle_interruption()` and `sync_history`, all of which run before `invalidate_pending_responses()`. If the output loop's 50ms WAIT sleep expires in that window it sees a still-valid sequence and ships the old-language audio the gate was holding, and since `activity` was captured before the release `audio_playing` is False, so the 200ms gap is skipped and those chunks run straight into the new-language handoff. Could the telemetry record stay here and `self.lid_playback_gate = None` move to after `__cleanup_downstream_tasks()`?

bolna/agent_manager/task_manager.py
```diff
@@ -5201,25 +5334,32 @@ async def __lid_idle_watcher(self):
...
-                # cost of one extra LLM call, not a wrong switch.
-                streak_fired = mismatch and pool.lid_buffer_language_streak() >= mismatch_streak_fire
-                if streak_fired:
-                    threshold = 0.0
-                if age >= threshold:
+                # An all-active-language buffer can only produce "stay", so firing would spend
+                # ~1.5-2s of decide (and the lock it holds) on a foregone conclusion. Skipping
+                # deliberately does NOT drain: if the main ASR later delivers this turn, the
+                # turn-boundary decide still sees the full transcript.
+                nothing_to_decide = saw_tags and not mismatch
```
san0808, line 5391: This skip deliberately does not drain and the stale-decision guard above now retains as well, but `_buffer_max_segment_s` and `_buffer_segments` only reset in `take_turn_transcript`. A stretch with no draining decide therefore leaves a buffer spanning several utterances, and `buffered_max_segment_s` becomes a multi-utterance max that feeds both the telemetry record and the `__recent_detected_turns` fallback. Should the buffer get an age or size bound so a skipped generation cannot lend its duration to a later turn?

bolna/transcriber/transcriber_pool.py
```diff
@@ -606,6 +639,7 @@ async def cleanup(self):
 
         # Stop LID tap
         if self._lid is not None:
+            self._record_detector_health()
```
san0808, line 646: `cleanup()` is appended to `tasks_to_cancel` in task_manager and only awaited at the `asyncio.gather` at the end, well after `output["lid_detection_events"] = list(...)` is snapshotted. The code there already notes the invariant ("tasks_to_cancel is only awaited after the snapshot"), so this record lands on the pool list after the copy and never reaches `lid_shadow_events`, leaving the health signal log-only. Should it be recorded before the snapshot, or the transcriber cleanup awaited ahead of building the output?

puneet-bolna: resolved

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

bolna/prompts.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

PR #920, opened 2026-08-03
fix(gemini): defer tool dispatch post-stream to fix silent turns on p…

web1havv:
Problem
Gemini agents were going completely silent mid-call whenever a tool was triggered. The user would ask something, Gemini would decide to call a tool, and the agent would say absolutely nothing back — no tool result, no fallback text, just dead air.

Root cause in gemini_llm.py:

if not all(k in fn_args for k in required_keys):
    logger.warning("missing required params")
    continue  # ← dropped the entire turn silently
When Gemini returned a tool call without all declared required params (e.g. custom_task_get_product_price had 7 required params including system-injected ones like execution_id, function_name that Gemini can't infer from conversation), this continue skipped tool dispatch AND yielded no text. With synthesize=True in conversation mode, the empty buffer → no end-of-stream yield → pipeline got null → complete silence.

Confirmed on production: exec a2d9367f, agent 74466f27, gemini-2.5-flash, Plivo outbound, Hindi — 5 consecutive silent turns, caller heard nothing and hung up.

This bug was Gemini-specific — the OpenAI path has zero required-param validation and always dispatches.

What was changed
1. Removed the silent continue Required-params check is now warning-only. Tool dispatches regardless — matching OpenAI path behaviour. Downstream API handles validation if args are truly missing.

2. Args accumulation per call_id across stream chunks Added _pending_fn_args: dict[str, dict] to merge function call args from every chunk for the same call_id. Defensive against Gemini streaming partial args across multiple chunks (supported on Vertex AI with streamFunctionCallArguments).

3. Deferred dispatch — tools fire post-stream, exactly once Added _pending_dispatch: dict[str, dict] to collect context during the stream loop. All FunctionCallPayload yields now happen after the stream completes. This guarantees:

Each tool fires exactly once per call (no duplicate dispatches if same call_id appears in multiple chunks)
Args are always the fully merged complete set at dispatch time
4. End-of-stream safety net If synthesize=True, buffer is empty, and no tool was dispatched → logs a loud ERROR so dead turns are visible in Loki instead of vanishing silently.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.168"
+__version__ = "0.10.169"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/llms/gemini_llm.py
+++ b/bolna/llms/gemini_llm.py
@@ -273,6 +273,14 @@ class GeminiLLM(BaseLLM):
         # usage_metadata repeats on every chunk with cumulative counts, and some models omit
         # it on the final one, so keep the last non-empty value rather than the final chunk's.
         stream_usage = None
+        # Gemini streams function-call args incrementally across chunks; accumulate them
+        # per call_id so we dispatch with the full merged arg set, not a partial snapshot.
+        _pending_fn_args: dict[str, dict] = {}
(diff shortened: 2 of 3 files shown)


san0808 (approved):
LGTM

PR #921, opened 2026-08-03
fix(gemini): flush buffered text before dispatching a tool call

san0808:
Gemini never flushed buffered text before dispatching a tool call, so when the model emitted a goodbye and called `end_call` in the same turn, the caller heard nothing. `task_manager` dispatches on the function-call chunk and abandons the generator, so the post-loop flush never runs. Any closing line shorter than `buffer_size` (inherited from the synthesizer config, commonly 200) was lost.

It also skipped the recovery path: because `textual_response` was non-empty, `__execute_function_call` treated the goodbye as already spoken and bypassed the follow-up LLM that would otherwise have generated one.

`openai_base.py` already flushes on `OUTPUT_ITEM_ADDED` for a function call. This applies the same flush to Gemini.

Verified on staging against `gemini-2.5-flash`, simulating `task_manager`'s early return on the function-call chunk. Before: `FC -> TEXT(33)`, caller hears nothing. After: `TEXT(33) -> FC`, caller hears the goodbye. 6/6 both ways.

BOLNA-2115

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.169"
+__version__ = "0.10.170"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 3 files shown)


web1havv (approved):
LGTM

PR #922, opened 2026-08-03
fix: wait for stream_sid before building the welcome message packet

SubhrajitPrusty:
Commit authored by @san0808; opening the PR so it doesn't sit unraised.

## Problem

`__forced_first_message` polls for `stream_sid` every 10ms, but it rebuilds the entire welcome-message packet on **every iteration** before checking whether the stream is ready — a fresh `uuid4()`, the full `meta_info` dict, the audio chunk, and for sip-trunk a `pcm_to_ulaw()` conversion of the cached welcome audio. All of it is discarded when the check fails. It also logs one line per iteration.

A connection that never carries a `start` frame keeps `stream_sid` at `None` for the whole `timeout=10.0` window, so each one costs roughly **1,000 wasted packet builds and 1,000 log lines** before giving up.

That is normally invisible. It stopped being invisible during the connection storm on 3 Aug (BOLNA-2088), when 1,800–4,100 agent setups/min were reaching this loop while only 3–24/min ever received a `start` event. Fleet-wide log volume hit 853K–1.53M lines/min, and the packet rebuilds ran on the same event loops that were already starved — outbound LLM/transcriber/synthesizer handshakes were queueing at ~3,000ms against a 28ms baseline.

To be precise about scope, since the incident write-up overstates this: the loop **is** bounded, at 10s. It does not hang. The cost is the per-iteration work and log line multiplied across thousands of concurrent spurious connections, not an unbounded wait.

## Change

Split the wait from the work. The loop now does nothing but poll `stream_sid` / `output_handler_set` and check the deadline; everything else moved below it and runs once, after the stream is confirmed usable.

- The per-iteration `"Stream id is still None ..."` line becomes a single `"Waiting for stream_sid ..."` before the loop.
- The timeout path logs the observed `stream_sid` and `output_handler_set` values instead of a bare message, then ends the conversation and returns.
- No behaviour change on the success path. `synthesizer_start_time` is now stamped when the stream is ready rather than on the final poll iteration — within one 10ms tick of before.

Diff is 88/88 because most of it is de-indenting the body by one level.

## Not covered

There is a second instance of the same shape at `task_manager.py:6721` — `"Stream id is still None, so not passing it"` with a matching `await asyncio.sleep(0.01)`. Left alone here to keep this reviewable; it belongs with the broader input-handler audit (BOLNA-2096).

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.168"
+__version__ = "0.10.169"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -1309,103 +1309,103 @@ class TaskManager(BaseManager):
             if delay_ms > 0:
                 logger.info(f"Welcome message delay set to {delay_ms} ms")
                 await asyncio.sleep(delay_ms / 1000)
+            # Wait for the stream to be usable before building anything. A connection that never
+            # carries a start frame keeps stream_sid None for the whole timeout, so any work inside
+            # this loop is paid ~100 times a second per such connection and thrown away.
             start_time = asyncio.get_running_loop().time()
+            logger.info("Waiting for stream_sid before sending the welcome message")
             while True:
-                elapsed_time = asyncio.get_running_loop().time() - start_time
-                if elapsed_time > timeout:
-                    await self.__process_end_of_conversation()
-                    logger.warning("Timeout reached while waiting for stream_sid")
-                    break
-
-                text = self.kwargs.get("agent_welcome_message", None)
-                meta_info = {
-                    "io": self.tools["output"].get_provider(),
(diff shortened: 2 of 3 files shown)


san0808:
Rebased onto latest master (b12e9a60, version 0.10.170) and force-pushed, so the branch is now 1 commit ahead with no conflicts. The rebase dropped the bot's 0.10.169 bump commit, which was stale against master anyway; leaving the bump to the auto bot on this PR rather than carrying it by hand.

Verified on apollo (single-uvicorn ws box) with this build installed:

1. Real inbound PSTN call, 47s, 44 transcriber messages. Welcome message sent 10ms after the wait line.
2. Real outbound PSTN call. Same, with the welcome mark acked by the carrier so it played out.
3. Timeout path on a real agent build: fired at 10.006s, ended the conversation, 0 per-iteration log lines against roughly 1,000 before.
4. Success path on both inbound and outbound routes.

The 10s bound is unchanged; only the per-iteration work and log line are gone.

PR #930, opened 2026-08-05
fix: restore OpenAI transcriber utterance-timeout safety net to 3s

### LINKED ISSUE #711: OpenAI realtime whisper hears speech but final transcript comes too late, so agent never replies

greeshmab21:
## Summary

`OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S` was silently regressed from `3.0` to `0.5` in
#721 ("fix whisper issue") with no stated rationale. This constant governs
`OpenAITranscriber.monitor_utterance_timeout()`, a safety net that force-finalizes a
committed turn if OpenAI's realtime `transcription.completed` event never arrives.

At 0.5s, this fires on ordinary turns — OpenAI's realtime completion routinely takes
more than 500ms — and force-finalizing used to silently discard the turn with zero
downstream signal. The agent never even learns the caller said anything. From the
caller's side that looks exactly like #711 describes: interim words appear, then
nothing, then "are you still there?" or a hangup.

Closes #711.

## Root cause

Confirmed via git history: the constant was intentionally `3.0` when the OpenAI
transcriber was first added, then dropped to `0.5` in #721 as an incidental part of a
larger, squashed rewrite of the transcriber's turn-detection logic. Nothing in that PR
explains or justifies the drop.

## Changes

- Restore `OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S` to `3.0` in `bolna/constants.py`,
  with a comment explaining why this must stay well above typical completion latency.
- Harden `monitor_utterance_timeout()` so that even a genuine timeout doesn't throw
  the caller's speech away: it now falls back to whatever interim transcript deltas
  already arrived and pushes those downstream, instead of silently dropping the turn.

## Testing

Added `tests/test_openai_utterance_timeout.py`:
- confirms the constant is restored to `3.0`
- confirms the monitor does NOT fire within normal (1.5s) completion latency
- confirms it correctly falls back to the assembled interim transcript when genuinely
  stalled past 3s
- confirms it stays silent (no noise packet) when there's no interim text at all

All 4 tests pass locally. This has not yet been verified against a live OpenAI
realtime connection or a full end-to-end call — happy to follow up with that if
useful.

### DIFF
--- a/bolna/constants.py
+++ b/bolna/constants.py
@@ -11,7 +11,14 @@ WEB_BASED_CALL_PROVIDER = "web_based_call"
 WEBCALL_TTS_SAMPLE_RATE = 24000
 
 OPENAI_TRANSCRIBER_HEARTBEAT_INTERVAL_S = 5
-OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S = 0.5
+# Last-resort safety net for a missing/dropped transcription.completed event, NOT a
+# normal-path timeout. OpenAI's realtime transcription completion can legitimately take a
+# couple of seconds after commit, so this must stay well above typical completion latency.
+# Regressed from 3.0 to 0.5 in #721 with no stated rationale; at 0.5s the monitor fires on
+# ordinary turns and silently discards the user's speech before OpenAI's real transcript
+# ever arrives, which is what issue #711 ("final transcript comes too late, so agent never
+# replies") actually observed. Restored to the pre-regression value.
+OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S = 3.0
 
 # ElevenLabs realtime (scribe_v2_realtime) accepts up to 50 keyterms for biasing.
 ELEVENLABS_REALTIME_MAX_KEYTERMS = 50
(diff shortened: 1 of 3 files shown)


greeshmab21:
Found it in the git history — OPENAI_TRANSCRIBER_UTTERANCE_TIMEOUT_S was 3.0 originally and got quietly dropped to 0.5 in #721 with no explanation, so I just restored it.

PR #932, opened 2026-08-06
fix: recover from an empty Responses API turn instead of stalling the call

san0808:
When the Responses API ends a stream with `response.incomplete` and no output text, the agent speaks nothing and the call sits in dead air until the stall backstop force-hangs up, which then gets recorded as an inactivity timeout. A reasoning model can hit this whenever reasoning consumes the whole `max_output_tokens` budget before any visible token, since that budget covers reasoning as well as spoken text.

Three changes.

1. Log `incomplete_details.reason` on the incomplete event. It was previously dropped, so max_output_tokens could not be told apart from content_filter.

2. Retry once when an incomplete response emitted nothing at all. The guard requires no accumulated text and no function call, so a response that already streamed a partial answer keeps its current behaviour and nothing can be spoken twice. `invalidate_response_chain` has already run at that point, so the retry replays full history. The retry pins the flag off, bounding this at one extra attempt.

3. Clear `response_in_pipeline` when a turn produces no speakable text. It was only cleared when synthesizer audio actually got sent or when an exception propagated, so a silent turn left it set, which gates every recovery branch in `__check_for_completion`: the user-online check, repeat-after-silence, and the normal inactivity hangup. The graph agent's silent router hop hit the same stuck flag.

Both the HTTP SSE and the persistent WebSocket Responses paths are covered. Function calls return before the new check and are unaffected, since their follow-up generation still owes audio. An interrupted turn raises `CancelledError`, which is not caught there, so a barge-in still leaves the flag to the new turn.

### DIFF
--- a/bolna/llms/openai_base.py
+++ b/bolna/llms/openai_base.py
@@ -643,9 +643,7 @@ class OpenAICompatibleLLM(BaseLLM):
                     response_usage = event.response.usage
                 break
 
-        # An incomplete response that emitted nothing leaves the turn silent. Nothing was
-        # yielded yet, so one clean retry on full history is safe; a reasoning model that
-        # spent its whole max_output_tokens budget on reasoning usually stays under it here.
+        # Nothing was yielded yet, so a retry cannot duplicate speech.
         if incomplete and not answer and not func_call_args and retry_on_empty:
             logger.warning(f"Responses API returned no output (reason={incomplete_reason}), retrying once")
             if isinstance(meta_info, dict):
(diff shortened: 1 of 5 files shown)


bolna/llms/openai_base.py
```diff
@@ -493,7 +493,15 @@ def _build_responses_create_kwargs(
...
     async def _generate_stream_responses(
-        self, messages, synthesize=True, request_json=False, meta_info=None, tool_choice=None, tools=None
+        self,
+        messages,
+        synthesize=True,
+        request_json=False,
+        meta_info=None,
+        tool_choice=None,
+        tools=None,
+        *,
```
prateeksachan: why this `*` here?

san0808: acha was testing and removed now

prateeksachan (approved):

PR #934, opened 2026-08-07
Bolna 2179

puneet-bolna:


bolna/transcriber/sarvam_transcriber.py
puneet-bolna, line 137: @san0808  here i didnt change because i dont think it is required, just have a look here once

san0808: so saaras v4 works without the param ?

san0808 (approved):

bolna/transcriber/sarvam_transcriber.py
puneet-bolna: yeah

PR #940, opened 2026-08-10
added the logs for speculative llm decide

puneet-bolna:
PR description
Title: fix(lid): log committed speculative follow-up like a normal LLM turn

Problem
When a language switch commits its speculative follow-up (the reply generated during the judge's decide and spoken directly), that main-LLM call is invisible everywhere: no LLM request/response rows in the execution logs (admin and non-admin dashboards both read the S3 CSV), no turn latency, no token usage, and LLM cost/price shows 0. On a call where the speculation hits and no other turns follow, the execution looks like it never touched the conversation LLM at all.

The fallback path (__generate_switch_followup → __do_llm_generation) logs normally — which is why some switch calls show their post-switch response and others show nothing.

Root cause
__speculative_followup_text() consumes llm_agent.generate() privately: it never calls convert_to_request_log, skips the dict messages carrying request payloads, and drops the token fields from the stream.

Change
Capture-at-generate, emit-at-commit:

__speculative_followup_text() now snapshots the formatted request (format_messages), api-reported token counts (input/output/reasoning/cached), and the latency object into self.lid_spec_capture when generation completes with text. llm_start_time rides the speculation's meta_info so latency stamping works natively.
New __log_committed_speculation(), called only on the commit branch ("speaking speculative follow-up"): emits the LLM REQUEST and RESPONSE rows exactly like a normal turn — the response row carries the tokens, so usage breakdown and pricing follow through the existing channel — and appends a stamped turn-latency entry.
Discarded speculations are intentionally not logged; the capture is cleared in the switch handler's finally and consumed on commit, so it can never leak into a later switch.
The emission is wrapped in try/except: a telemetry failure can never break the audible follow-up.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -573,9 +573,6 @@ class TaskManager(BaseManager):
         # Records every language switch — manual tool call (legacy) or LLM-driven
         # (triggered_by="lid_llm") — used post-call for precision / latency analysis.
         self.language_switch_events: list[dict] = []
-        # Request/usage snapshot of the in-flight speculative follow-up; emitted as
-        # normal LLM logs only when the speculation is committed (spoken).
-        self.lid_spec_capture = None
         # Legacy-flow handoff state (populated by __inject_switch_language_tool
         # when the LLM-driven switch flow is NOT enabled for this call).
         self.switch_handoff_messages = {}
@@ -5326,10 +5323,13 @@ class TaskManager(BaseManager):
             # Discard any unconsumed speculative generation — covers every exit path
             # (stay decisions, gate rejections, timeouts, exceptions) without littering
             # the run with per-return cancels.
-            if spec is not None and not spec.done():
-                spec.cancel()
-                logger.info("LanguageSwitcher: speculative follow-up discarded")
-            self.lid_spec_capture = None
+            if spec is not None:
+                if not spec.done():
+                    spec.cancel()
+                    logger.info("LanguageSwitcher: speculative follow-up discarded")
+                elif not spec.cancelled() and spec.exception() is None:
+                    discarded_text, discarded_capture = spec.result()
+                    self.__log_discarded_speculation(discarded_text, discarded_capture)
 
     async def __lid_idle_watcher(self):
         """Recover the stuck-language deadlock.
@@ -5824,9 +5824,9 @@ class TaskManager(BaseManager):
         # the speculation answers stale content, so fall through to fresh generation
         # (the caller's finally discards the unconsumed task).
         if spec_task is not None and spec_target == target and transcript_corrected:
-            spec_text = ""
+            spec_text, spec_capture = "", None
             try:
-                spec_text = await asyncio.wait_for(spec_task, timeout=6.0)
+                spec_text, spec_capture = await asyncio.wait_for(spec_task, timeout=6.0)
             except asyncio.CancelledError:
                 # Always re-raise: wait_for cancels spec_task BEFORE raising, so a
                 # cancelled spec_task is the signature of THIS handler being cancelled
@@ -5844,7 +5844,7 @@ class TaskManager(BaseManager):
                 return None
             if spec_text:
                 self.conversation_history.append_assistant(spec_text)
-                self.__log_committed_speculation(spec_text)
+                self.__log_committed_speculation(spec_text, spec_capture)
                 synth_meta = {
                     "io": self.tools["output"].get_provider(),
                     "request_id": str(uuid.uuid4()),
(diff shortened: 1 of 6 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -5326,6 +5329,7 @@ async def handle_language_switch(
             if spec is not None and not spec.done():
                 spec.cancel()
                 logger.info("LanguageSwitcher: speculative follow-up discarded")
+            self.lid_spec_capture = None
```
san0808: This finally runs after the lock is released and after `__generate_switch_followup`, so if an earlier handler is still generating a fallback follow-up while a later one commits its speculation, this line clears the slot between the later handler filling its capture and its commit branch reading it, and `__log_committed_speculation` then returns at `if not capture` with no error. Prod had 332 calls with 2+ switches in the last 7 days, 243 of them with a speculation on a switch. Should the capture ride the spec task instead (returned from `__speculative_followup_text`, consumed as a local), the way `_spec_followup_task` was moved into a local for the same reason?

bolna/agent_manager/task_manager.py
```diff
@@ -6073,6 +6078,53 @@ def __switch_context_note(self, target: str, detector_transcript: str, reasoning
...
+                message=spec_text,
+                meta_info=spec_meta,
+                model=model,
+                component=LogComponent.LLM,
+                direction=LogDirection.RESPONSE,
+                run_id=self.run_id,
+                input_tokens=capture["input_tokens"],
+                output_tokens=capture["output_tokens"],
+                reasoning_tokens=capture["reasoning_tokens"],
+                cached_tokens=capture["cached_tokens"],
```
san0808, line 6112: A normal turn also reports these tokens to `on_turn_usage` from `__store_into_history` (task 0, gated on `input_tokens`), which is the PTU capacity tally, and those are its only two call sites. A committed speculation runs on the same deployment, so should this path report too?

bolna/agent_manager/task_manager.py
```diff
@@ -6108,20 +6160,46 @@ async def __speculative_followup_text(
...
             if llm_message.data:
                 text += " " + llm_message.data
             if llm_message.end_of_stream:
                 break
-        return text.strip()
+        text = text.strip()
+        if text:
+            # Request/usage snapshot for the commit path — logged like a normal turn
+            # ONLY if this speculation is actually spoken; discards stay unlogged.
+            self.lid_spec_capture = {
```
san0808: Over the last 7 days prod started 7,376 speculations and at most 1,452 of them were committable, so roughly 80% are discarded after spending a full-history input plus output on the conversation model, and no row is written for them. Should discards be logged under `LLM_LANGUAGE_SWITCH`? `calculate_cost_from_request_logs` has branches only for `llm`, `llm_hangup`, `llm_voicemail`, `graph_routing` and `synthesizer`, so that would make the spend visible without billing the customer for a reply nobody heard.

bolna/agent_manager/task_manager.py
```diff
@@ -6073,6 +6078,53 @@ def __switch_context_note(self, target: str, detector_transcript: str, reasoning
...
+                message=capture["request_message"],
+                meta_info=spec_meta,
+                model=model,
+                component=LogComponent.LLM,
+                direction=LogDirection.REQUEST,
+                run_id=self.run_id,
+            )
+            convert_to_request_log(
+                message=spec_text,
+                meta_info=spec_meta,
```
san0808: `spec_meta` carries `sequence_id: -1`, and `llm_turn_usage` in `pricing_helpers` falls back to a positional index when that column is not `isdigit()`, so this row lands in `llm_usage_breakdown.conversation.turns` under the same `seq` as the next real turn (running a CSV through bolna's writer and the backend parser gives two entries labelled `seq: 3`). Should the log rows carry a real sequence id, with -1 kept only on the synth meta where the gate bypass needs it?

bolna/agent_manager/task_manager.py
```diff
@@ -6073,6 +6078,53 @@ def __switch_context_note(self, target: str, detector_transcript: str, reasoning
...
+                self._stamp_llm_latency_dict(
+                    latency_dict,
+                    spec_meta,
+                    capture["input_tokens"],
+                    capture["output_tokens"],
+                    capture["reasoning_tokens"],
+                    capture["cached_tokens"],
+                    response_text=spec_text,
+                )
+                self.llm_latencies.turn_latencies.append(latency_dict)
```
san0808, line 6133: `datadog_helpers` emits `voiceai.llm.first_token_latency_ms` for every `turn_latencies` entry and counts them into `voiceai.llm.turn_count`, and this entry has no field marking it as speculation. Worth stamping `origin` from the spec meta?

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

san0808 (approved):

PR #944, opened 2026-08-11
feat: support eleven v3 conversational over the text-to-dialogue websocket

san0808:
Eleven v3 is rejected with a 403 on the `multi-stream-input` endpoint the existing synthesizer uses. It is served only from `/v1/text-to-dialogue/stream-input`, which is a different protocol: voices register in a first message rather than the URL path, text goes as `inputs` arrays, and turns end on `is_final_audio_for_turn` rather than `isFinal`. So this cannot be a config-only change.

`ElevenlabsBase` now holds what the two protocols genuinely share (credentials, host, wire-format selection, audio post-processing and the one-shot HTTP path), with `ElevenlabsSynthesizer` and `ElevenlabsV3Synthesizer` as siblings under it. They are not a specialisation of one another: v3 previously extended v1 and had to null out inherited context state it never uses. `providers.py` dispatches on the model prefix, so every non-v3 model resolves exactly as before.

Three properties of the endpoint shape the implementation. There are no contexts and no control message cancels a turn, so an interruption closes the socket and dials a replacement on a background task; a close we caused is told apart from one the provider caused, because swallowing the latter would leave the turn with no end-of-stream and playback stuck marked as in progress. The 20s idle cap is not adjustable here (`inactivity_timeout` is ignored), so a keep-alive heartbeat is required. And `optimize_streaming_latency` is rejected by v3, so the HTTP path used for welcome clips omits it.

Only `eleven_v3_conversational` is wired up. v3 stability is a three-way preset, so it is snapped to the nearest of 0.0/0.5/1.0, and `similarity_boost`, `speed` and `style` are measurably inert on v3 so none are sent.

Companion catalog and pricing changes are in bolna-ai/dashboard-backend#2807.

### DIFF
--- a/bolna/synthesizer/elevenlabs_synthesizer.py
+++ b/bolna/synthesizer/elevenlabs_synthesizer.py
@@ -442,6 +442,7 @@ class ElevenlabsV3Synthesizer(ElevenlabsSynthesizer):
         self._new_turn_pending = True
         self._connect_lock = asyncio.Lock()
         self._keep_alive_task = None
+        self._reconnect_task = None
         self._last_send_time = time.perf_counter()
 
     def get_sleep_time(self):
@@ -562,7 +563,9 @@ class ElevenlabsV3Synthesizer(ElevenlabsSynthesizer):
             self._new_turn_pending = True
             if self._is_ws_connected():
                 await self.websocket.close()
-            asyncio.create_task(self._ensure_connection())
+            # Held on the instance: a bare create_task can be garbage collected mid-flight,
+            # which would drop the redial and leave the next turn waiting on monitor_connection.
+            self._reconnect_task = asyncio.create_task(self._ensure_connection())
         except Exception as e:
             logger.info(f"Error handling ElevenLabs v3 interruption: {e}")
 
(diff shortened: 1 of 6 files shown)


bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -409,3 +409,359 @@ async def _generate_http(self, text, format=None):
...
+                # marker; is_final arrives only when the session closes.
+                if data.get("is_final_audio_for_turn"):
+                    logger.info(f"WS recv is_final_audio_for_turn trace_id={self.ws_trace_id}")
+                    audio_chunk_count = 0
+                    yield b"\x00", ""
+
+            except websockets.exceptions.ConnectionClosed:
+                # Keep looping rather than breaking as v1 does: barge-in closes the socket
+                # routinely, and returning here would end generate() for the rest of the call.
+                if self.conversation_ended or self.connection_error:
```
puneet-bolna, line 744: In the v3 receiver's ConnectionClosed handler, a provider drop yields b"\x00" when self.last_text_sent is true. But last_text_sent stays true from the previous completed turn until the next push resets it. Sequence: turn completes normally (is_final_audio_for_turn → sentinel yielded) → socket idles → keep-alive send fails once (its exception is swallowed) → provider drops at 20s → receiver yields a second sentinel for a turn that already ended. _generate_ws_loop then emits another end_of_synthesizer_stream + end_of_llm_stream packet on stale meta, which downstream can read as the agent finishing a turn again. v1 dedupes this with _eos_context_id; v3 has no per-turn dedup. Gate the yield on a turn actually being in flight (e.g. current_turn_start_time is not None or an "eos already emitted" flag).

bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -409,3 +409,359 @@ async def _generate_http(self, text, format=None):
...
+        async with self._connect_lock:
+            if self._is_ws_connected():
+                return True
+            websocket = await self.establish_connection()
+            if websocket is None:
+                return False
+            self.websocket = websocket
+            return True
+
+    async def monitor_connection(self):
```
puneet-bolna, line 528: monitor_connection/_ensure_connection never check conversation_ended. cleanup() (called in __listen_synthesizer's finally, task_manager.py:6456) closes the socket, but the monitor task is cancelled later (task_manager.py:7357). In between, the monitor's 1s poll sees a dead socket and redials a fresh authenticated connection that nobody closes. The base class has the same flaw,

bolna/synthesizer/elevenlabs_synthesizer.py
san0808: Fixed with a _turn_eos_emitted flag, set false when a turn's first fragment goes out and true whenever the sentinel is yielded

prateeksachan (approved):

PR #945, opened 2026-08-11
Added the turn id for pre synthesize message and dropped asr

puneet-bolna:
fix: give transcript rows the ids and timing the dashboard needs to place them

  ## Problem

  The dashboard can't reliably say *when* a transcript line was spoken, so clicking a line
  seeks to the wrong point in the recording (BLT: "tool call takes to 45s, actually spoken
  at 1:30"). Root cause is on this side: rows leave bolna without the identifiers and
  timing needed to place them, forcing the consumer to correlate by text.

  Measured on 100 long prod calls: **75% of transcript rows carried a fabricated timestamp.**

  ## Changes

  **1. User messages carry `turn_id`** (`conversation_history.py`, `task_manager.py:4397`)
  `append_user` took only content while `append_assistant` had `**kwargs` — an accident of
  `2ca5213`, which added kwargs for an unrelated reason and left the user path behind. The
  value was already in hand and logged on the very next line. Prod: assistant 8/9 messages
  carry `turn_id`, user **0/10**.

  Not changed: the two synthetic call sites (`:5812` idle-flush, `:6832` `_inject_and_run_llm`).
  Neither maps to a real ASR turn, so they stay id-less by design.

  **2. Whisper turn_ids are ints, and unique** (`openai_transcriber.py`)
  - Three sites emitted `f"turn_{n}"` though the counter was already monotonic. The string
    form tripped `task_manager.py:7551`'s `isinstance(_tid, int)` guard, which was silently
    dropping **every** Whisper transcriber turn from the `user_bot_latencies` enrichment.
  - OpenAI emits several transcription items per committed turn, all reusing
    `_last_committed_turn_id` (prod `2e032eca`: `turn_3`×2, `turn_13`×2, `turn_34`×2), so any
    consumer keyed on turn_id lost all but the last. Extras now get a fresh id via
    `_last_latency_turn_id`. Deliberately **not** `_upsert_turn_latency` — that collapses them
    and discards a real transcript.

  **3. Canned agent messages are tagged** (`task_manager.py`)
  `is_user_online_message` and `handoff` now carry `sequence_id=-1` + `message_category`.
  These are template strings with no LLM turn; previously nothing distinguished them from a
  model reply.

  **4. `synthesizer_chunk_marks` added to `progression_data`** (`task_manager.py:7493`)
  It was in `latency_dict` but absent from **930,797 / 930,797** observability blobs. It is the
  only record carrying both the *text* and the *send time* of template speech, so without it
  the ~130 canned agent rows per 100 calls cannot be timed at all — a consumer-side anchor
  written against it measured exactly 0 hits.

  ## LLM payload safety — verified, not assumed

  `self.history` is passed unfiltered (`task_manager.py:2966` → `contextual_conversational_agent.py:73`
  → `azure_llm.py:132` / `litellm.py:73`), so extra message keys already reach providers today:
  **643 messages carrying `turn_id`/`response_uid` across 145/150 prod calls, 0 failures.**

  | adapter | providers | behaviour |
  |---|---|---|
  | Azure/OpenAI chat | azure, openai, custom, ola | ignores unknown keys (proven in prod) |
  | Responses API | `use_responses_api` | `ChatMessage(**msg)` strips them (executed, pydantic 2.13.4) |
  | Gemini | google | `_prepare_history` reads only role/content/tool_calls |
  | LiteLLM | anthropic, groq, deepseek, bedrock, cohere, openrouter, together, perplexity, … | **8/8 models accepted**; `anthropic_messages_pt` rebuilds the payload — `turn_id` does not leak |

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -214,6 +214,14 @@ _NON_NODE_RESPONSE_CATEGORIES = frozenset(
 )
 
 
+def asr_id_to_int(value):
+    """Coerce OpenAI's "turn_3" ASR ids to int (Deepgram's are already ints); unparseable -> None."""
+    if isinstance(value, str):
+        digits = "".join(filter(str.isdigit, value))
+        return int(digits) if digits else None
+    return value
+
+
 def trailing_utterance_text(segments, gap_seconds=4.0):
     """Text of the caller's LAST utterance: trailing detector segments in the same
     language, back to a real silence boundary. Breaks on either a gap > gap_seconds
@@ -4405,8 +4413,8 @@ class TaskManager(BaseManager):
                 meta_info.get("response_uid"),
             )
 
-        # asr_turn_id, not meta_info["turn_id"] — that one counts responses, not ASR turns.
-        self.conversation_history.append_user(transcriber_message, asr_turn_id=meta_info.get("asr_turn_id"))
+        # asr_turn_id (int-coerced), not meta_info["turn_id"] — that one counts responses, not ASR turns.
+        self.conversation_history.append_user(transcriber_message, asr_turn_id=asr_id_to_int(meta_info.get("asr_turn_id")))
         logger.info(
             "BOLNA_TRACE_TM append_user seq=%s turn=%s response_uid=%s history_len=%s text=%r",
             meta_info.get("sequence_id"),
@@ -4786,7 +4794,13 @@ class TaskManager(BaseManager):
                             self.eager_llm_task = asyncio.create_task(
                                 self._run_llm_task(create_ws_data_packet(eager_transcript, meta_info))
                             )
-                            self.history.append({"role": "user", "content": eager_transcript})
+                            # Committed user row when EndOfTurn(was_eager) skips _handle_transcriber_output.
+                            # meta_info["asr_turn_id"] is stale until EndOfTurn, so read the live id.
+                            eager_user_row = {"role": "user", "content": eager_transcript}
+                            eager_asr_turn_id = asr_id_to_int(getattr(active_transcriber, "current_turn_id", None))
+                            if eager_asr_turn_id is not None:
+                                eager_user_row["asr_turn_id"] = eager_asr_turn_id
+                            self.history.append(eager_user_row)
                         else:
                             logger.info(f"Skipping speculative LLM (audio playing or welcome not done)")
 
@@ -7521,7 +7535,8 @@ class TaskManager(BaseManager):
                     "mark_tracking": output["latency_dict"]["mark_tracking"],
                     # Only record of when template speech (are-you-still-there, tool fillers,
                     # handoffs, goodbyes) was actually spoken — it has no LLM turn to anchor to.
-                    "synthesizer_chunk_marks": copy.deepcopy(output["latency_dict"]["synthesizer_chunk_marks"]),
+                    # Shared ref like mark_tracking — get_chunk_marks builds fresh dicts, nothing mutates them.
+                    "synthesizer_chunk_marks": output["latency_dict"]["synthesizer_chunk_marks"],
                     "hangup_triggered_ms": round(self.hangup_triggered_at * 1000 - self.conversation_start_init_ts, 2)
                     if self.hangup_triggered_at
                     else None,
@@ -7573,9 +7588,10 @@ class TaskManager(BaseManager):
                     if _ub.get("turn_id") is not None
                 }
                 for _tt in output["progression_data"]["transcriber_latencies"].get("turn_latencies", []):
-                    _tid = _tt.get("turn_id")
-                    # Any id will do — OpenAI's are strings ("turn_3"), Deepgram's are ints.
-                    # Requiring int here silently dropped every Whisper turn from this stamp.
+                    # _ub_turns holds ints; "turn_3" in {3} is always False, which duplicated every
+                    # covered Whisper turn. Compare/store as int (progression copy only).
+                    _tid = asr_id_to_int(_tt.get("turn_id"))
+                    _tt["turn_id"] = _tid
                     if _tid is None or _tid in _ub_turns or not _tt.get("final_transcript"):
                         continue
                     _u_start = _tt.get("asr_turn_start_ms")
@@ -7633,7 +7649,8 @@ class TaskManager(BaseManager):
 
                 _tts_turns = (output["latency_dict"]["synthesizer_latencies"] or {}).get("turn_latencies", [])
                 for _t in _tts_turns:
-                    _t.pop("tts_start_ms", None)
+                    for _f in ("tts_start_ms", "message_category"):
+                        _t.pop(_f, None)
 
                 for _e in output["latency_dict"]["user_bot_latencies"]:
                     for _f in ("user_first_start_ms", "agent_end_ms"):
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -4414,7 +4414,9 @@ class TaskManager(BaseManager):
             )
 
         # asr_turn_id (int-coerced), not meta_info["turn_id"] — that one counts responses, not ASR turns.
-        self.conversation_history.append_user(transcriber_message, asr_turn_id=asr_id_to_int(meta_info.get("asr_turn_id")))
+        self.conversation_history.append_user(
+            transcriber_message, asr_turn_id=asr_id_to_int(meta_info.get("asr_turn_id"))
+        )
         logger.info(
             "BOLNA_TRACE_TM append_user seq=%s turn=%s response_uid=%s history_len=%s text=%r",
             meta_info.get("sequence_id"),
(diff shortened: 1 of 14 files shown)


bolna/helpers/conversation_history.py
```diff
@@ -27,8 +27,10 @@ def append_welcome_message(self, content: str):
 
         self._interim = copy.deepcopy(self._messages)
 
-    def append_user(self, content: str):
-        self._messages.append({"role": ChatRole.USER, "content": content})
+    def append_user(self, content: str, **kwargs):
+        # kwargs (turn_id) let consumers correlate a user turn to its ASR turn by id
+        # instead of by text. LLM adapters ignore/strip unknown message keys.
```
SubhrajitPrusty: any evidence to back the claim that LLM adapters will strip the unknown message keys?

bolna/transcriber/openai_transcriber.py
```diff
@@ -88,7 +88,9 @@ def __init__(
         self._turn_start_epoch_ms = None
         # Saved at commit time so the receiver can still identify the turn after
         # _reset_turn_state() clears current_turn_id (e.g. utterance timeout race).
-        self._last_committed_turn_id: Optional[str] = None
+        self._last_committed_turn_id: Optional[int] = None
```
SubhrajitPrusty: turn_id became str to int without a mention?

bolna/helpers/conversation_history.py
puneet-bolna: added a strip before llm call so it drops before

bolna/transcriber/openai_transcriber.py
puneet-bolna: added back to string handling at progression now

bolna/helpers/conversation_history.py
SubhrajitPrusty: it was a question 😅

SubhrajitPrusty (approved):

bolna/agent_manager/task_manager.py
```diff
@@ -7538,14 +7560,17 @@ async def run(self):
                 }
                 for _tt in output["progression_data"]["transcriber_latencies"].get("turn_latencies", []):
                     _tid = _tt.get("turn_id")
-                    if not isinstance(_tid, int) or _tid in _ub_turns or not _tt.get("final_transcript"):
+                    # Any id will do — OpenAI's are strings ("turn_3"), Deepgram's are ints.
+                    # Requiring int here silently dropped every Whisper turn from this stamp.
+                    if _tid is None or _tid in _ub_turns or not _tt.get("final_transcript"):
```
san0808, line 7597: Should the `f"turn_{n}"` sites in `openai_transcriber` become plain ints rather than relaxing this guard? `_ub_turns` is built from `asr_turn_id`, which both `_stamp_llm_turn` and `_append_eager_llm_stub` read off the transcriber's `turn_counter` as an int, while Whisper entries carry `"turn_5"`. `"turn_5" in {5}` is False, so every Whisper turn that already has a real row gets a second all-None stub appended. Running this block with OpenAI-shaped data, two turns produce four rows with matching `user_start_ms`. The same type split makes the `_lt["turn_id"] = _asr_tid` promotion unjoinable as well, int on the LLM entry against `"turn_5"` on the ASR entry.

bolna/agent_manager/task_manager.py
```diff
@@ -4394,7 +4405,8 @@ async def _handle_transcriber_output(self, next_task, transcriber_message, meta_
                 meta_info.get("response_uid"),
             )
 
-        self.conversation_history.append_user(transcriber_message)
+        # asr_turn_id, not meta_info["turn_id"] — that one counts responses, not ASR turns.
+        self.conversation_history.append_user(transcriber_message, asr_turn_id=meta_info.get("asr_turn_id"))
```
san0808: When EndOfTurn arrives with `was_eager` set and the speculative task is reused, this method is skipped entirely, so the committed user row stays the bare `self.history.append` from the EagerEndOfTurn branch and never gets `asr_turn_id`. Is that intentional? It reads as a third real ASR turn alongside the two synthetic sites the description calls out, and on Flux with `eager_eot_threshold` set it is the normal path rather than an edge case.

bolna/synthesizer/stream_synthesizer.py
```diff
@@ -297,11 +302,13 @@ def _record_turn_latency(self):
                         "first_result_latency_ms": round((self.current_turn_ttfb or 0) * 1000),
                         "total_stream_duration_ms": round(total_stream_duration * 1000),
                         "characters": self.current_sequence_chars,
+                        "message_category": self.current_message_category,
```
san0808, line 305: Should `message_category` join the pop list for the synthesizer turns in the block that keeps `latency_dict` at master state? `turn_latencies` is an untyped `list`, so `model_dump()` passes this key straight through, and that block currently pops only `tts_start_ms`.

bolna/agent_manager/task_manager.py
```diff
@@ -7486,6 +7505,9 @@ async def run(self):
                     "interruption_stats": output["latency_dict"]["interruption_stats"],
                     "user_bot_latencies": copy.deepcopy(output["latency_dict"]["user_bot_latencies"]),
                     "mark_tracking": output["latency_dict"]["mark_tracking"],
+                    # Only record of when template speech (are-you-still-there, tool fillers,
+                    # handoffs, goodbyes) was actually spoken — it has no LLM turn to anchor to.
+                    "synthesizer_chunk_marks": copy.deepcopy(output["latency_dict"]["synthesizer_chunk_marks"]),
```
san0808: Could this share the reference the way `mark_tracking` does just above? `get_chunk_marks` already builds a fresh list of fresh dicts on every call and nothing mutates it after the snapshot, unlike `llm_latencies` and `user_bot_latencies` which are edited in place further down. It is one entry per audio chunk carrying `text_synthesized`, so it is the largest thing in the blob and this runs on the end-of-call path.

bolna/llms/message_models.py
```diff
@@ -22,6 +22,23 @@ class ChatMessage(BaseModel):
...
+# before the request is built: OpenAI-compatible providers forward unknown message keys
+# verbatim, and nothing should depend on the server ignoring them. Denylist rather than an
+# allowlist so a legitimate provider key (name, cache_control, multimodal parts) is never
+# silently dropped.
+INTERNAL_MESSAGE_KEYS = frozenset(
+    {"turn_id", "response_uid", "asr_turn_id", "sequence_id", "message_category", "exclude_from_llm"}
+)
+
+
+def strip_internal_keys(messages: list[dict]) -> list[dict]:
```
san0808, line 35: Should the non-stream `generate` paths in `openai_llm` and `azure_llm` strip too? They still pass `messages` through raw. Nothing reaches them with bookkeeping keys today since every caller builds a fresh prompt via `format_messages`, but the new tests assert the guarantee holds for every provider, and right now it only holds for the streaming path.

san0808 (commented):
Two things to fix before this lands:

1. Make the `openai_transcriber` turn ids plain ints. Relaxing the int guard as written doubles every Whisper turn in `user_bot_latencies` rather than recovering it, and leaves the LLM to ASR join broken. The description says the ids became ints, but all four `f"turn_{n}"` sites are still strings.
2. Stamp `asr_turn_id` on the eager-accepted user row, or say why it is excluded.

Minor: `message_category` now leaks into `latency_dict`, the `synthesizer_chunk_marks` deepcopy is redundant, and the non-stream `generate` paths are unstripped.

puneet-bolna:
> Two things to fix before this lands:
> 
> 1. Make the `openai_transcriber` turn ids plain ints. Relaxing the int guard as written doubles every Whisper turn in `user_bot_latencies` rather than recovering it, and leaves the LLM to ASR join broken. The description says the ids became ints, but all four `f"turn_{n}"` sites are still strings.
> 2. Stamp `asr_turn_id` on the eager-accepted user row, or say why it is excluded.
> 
> Minor: `message_category` now leaks into `latency_dict`, the `synthesizer_chunk_marks` deepcopy is redundant, and the non-stream `generate` paths are unstripped.



Kept the transcriber ids as strings per the earlier review thread, and fixed the join at the consumer instead: a single asr_id_to_int boundary coercion in task_manager (guard, stub, progression copy, history rows all int). Eager row now stamps the live current_turn_id — meta_info's copy is stale at that point since the latency upsert only lands at EndOfTurn. All three minors fixed."

bolna/llms/message_models.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

bolna/synthesizer/stream_synthesizer.py
puneet-bolna: resolved

bolna/agent_manager/task_manager.py

bolna/agent_manager/task_manager.py
puneet-bolna: resolved

san0808 (approved):

PR #951, opened 2026-08-12
fix: retry once on transient OpenAI server_error before terminating call

web1havv:
A single transient 500 server_error from OpenAI was immediately raising APIError → LLMError → call terminated with no retry. 
Add a one-shot retry with 500ms backoff specifically for server_error on the WS Responses API path, matching the existing retry pattern used for previous_response_not_found. The _server_error_attempt guard ensures we only retry once and don't loop on persistent failures.

### DIFF
--- a/bolna/llms/openai_llm.py
+++ b/bolna/llms/openai_llm.py
@@ -498,7 +498,14 @@ class OpenAiLLM(OpenAICompatibleLLM):
             return {"type": "text"}
 
     async def _generate_stream_ws_responses(
-        self, messages, synthesize=True, request_json=False, meta_info=None, tool_choice=None, tools=None, _server_error_attempt=0
+        self,
+        messages,
+        synthesize=True,
+        request_json=False,
+        meta_info=None,
+        tool_choice=None,
+        tools=None,
+        _server_error_attempt=0,
     ):
         """Stream via persistent WebSocket — same interface as _generate_stream_responses."""
         if not messages:
--- a/bolna/llms/openai_llm.py
+++ b/bolna/llms/openai_llm.py
@@ -498,7 +498,7 @@ class OpenAiLLM(OpenAICompatibleLLM):
             return {"type": "text"}
 
     async def _generate_stream_ws_responses(
-        self, messages, synthesize=True, request_json=False, meta_info=None, tool_choice=None, tools=None
+        self, messages, synthesize=True, request_json=False, meta_info=None, tool_choice=None, tools=None, _server_error_attempt=0
     ):
         """Stream via persistent WebSocket — same interface as _generate_stream_responses."""
         if not messages:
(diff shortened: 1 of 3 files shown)


san0808 (commented):
Main thing to sort out is that the retry deadlocks on the transport lock, so today it never actually retries. In priority order:

1. Move the resend into `stream_response` so it doesn't re-enter `_ws_transport._lock`
2. Forward `tools` on the retry call
3. Only retry when nothing has been emitted yet
4. Reset the response chain when the error follows `response.created`

bolna/llms/openai_llm.py
```diff
@@ -546,6 +553,14 @@ async def _generate_stream_ws_responses(
                         ):
                             yield chunk
                         return
+                    if error_code == "server_error" and _server_error_attempt == 0:
+                        logger.warning(f"WS transient server_error from OpenAI, retrying once after 500ms")
+                        await asyncio.sleep(0.5)
+                        async for chunk in self._generate_stream_ws_responses(
```
san0808, line 559: `stream_response` holds `_ws_transport._lock` across its entire body and yields each event before it checks `TERMINAL_EVENTS`, so at this point the transport generator is suspended at `yield evt` with the lock still held. Recursing here re-enters `async with self._lock` from the same task, which can never be granted, so the turn hangs instead of retrying (I ran the structure standalone and it never completes). Should the resend instead live inside `stream_response` as a `while` loop that re-sends `response.create`, where the lock is already ours? That also keeps `create_params` intact, which sidesteps the two issues below.

bolna/llms/openai_llm.py
```diff
@@ -546,6 +553,14 @@ async def _generate_stream_ws_responses(
...
+                            messages, synthesize, request_json, meta_info, tool_choice, _server_error_attempt=1
```
san0808, line 560: `tools` isn't forwarded here, so the retry falls through to `_parse_tools(None)` and picks up `self.tools`. Graph agents pass per-node tools via `_tools_for_node`, so wouldn't the retried turn get the agent-wide tool set instead of the current node's? The `previous_response_not_found` call just above drops it the same way.

bolna/llms/openai_llm.py
```diff
@@ -546,6 +553,14 @@ async def _generate_stream_ws_responses(
                         ):
                             yield chunk
                         return
+                    if error_code == "server_error" and _server_error_attempt == 0:
```
san0808, line 556: If the 500 lands mid-generation rather than at admission, some `OUTPUT_TEXT_DELTA` chunks have already been yielded to the synthesizer and the retry replays the turn from the top, so the caller hears the opening twice. Should this also require that nothing has been emitted yet?

san0808, line 556: Separately on this branch: when the error arrives after `response.created`, the CREATED handler has already pointed `self.previous_response_id` at the response that just failed, and the retry rebuilds `create_params` off it, so it chains onto a dead response. `FAILED` and `INCOMPLETE` both call `invalidate_response_chain()` first, should this path do the same? Reusing the original `create_params` would avoid the question entirely.

bolna/llms/openai_llm.py
```diff
@@ -546,6 +553,14 @@ async def _generate_stream_ws_responses(
...
+                        logger.warning(f"WS transient server_error from OpenAI, retrying once after 500ms")
```
san0808, line 557: Small one: f-string with no placeholders, and neither `error_info` nor `run_id` makes it into the log. Worth adding both, since this is the line you'd be grepping for when it fires in prod?

PR #954, opened 2026-08-13
{{}} variable support

puneet-bolna:
### Problem
Prompt substitution used `str.format_map`, which parses **every** brace in the prompt. A JSON
literal raised `ValueError`, and `update_prompt_with_context` swallowed it with
`except Exception: return prompt` — so a single JSON snippet silently left **every** variable in
that prompt unsubstituted. `{a.b}` dot notation failed identically via `AttributeError`.

Verified in prod (ClickHouse, 76,747 live agents): 3,733 agents contain a `{"` JSON literal,
861 contain `{}`, 104 contain `{ident:`, 93 contain `{a.b}`. Of those, ~995 also reference a
variable and were therefore losing substitution on every call. 1,795 transcripts in the last
30 days (8.2M calls) contain a literal `{placeholder}` — a missing variable renders *empty*
under `DictWithMissing`, so those can only come from a template returned unsubstituted.

### Change
`render_prompt(template, data, missing="")` in `helpers/utils.py` — a regex renderer that only
substitutes recognised variable patterns. JSON literals, format specs and `{}` match nothing and
pass through untouched, so JSON in a prompt structurally cannot break substitution.

- Supports `{{path}}` (new, matches Vapi/Retell convention) and `{path}` (legacy) in one pass.
- Nested access: `{{a.b.c}}`, `{{a.b.0.c}}` walk dicts/lists; legacy `{a[b][c]}` still resolves.
- `missing=None` partial-fill mode for template seeding — unresolved tokens stay byte-identical.
- Rewired `update_prompt_with_context` (~22 call sites: system prompts, welcome/hangup messages,
  graph nodes, static text) and `graph_agent.py` routing instructions (its `missing="NULL"` kept).
- `_PROMPT_VAR_PATTERN` now captures the ROOT identifier of either syntax, so `{{prior.loans}}`
  validates against `recipient_data["prior"]` — previously nested refs skipped validation entirely.

### Backward compatibility
Old behaviour is preserved exactly, verified by diffing old vs new over the top 120 real prod
token shapes weighted by frequency: **116/120 identical (96.2% of 210,930 occurrences)**, 1 fixed
(`{}`), 3 changed — all `{{identifier}}`, i.e. the new syntax.

- `{name}`, unknown vars → `""`, `{prior[loans][0][amount]}`, and legacy format specs
  (`{price:.2f}` → `12.50`) all render identically.
- **Objects keep Python `str()` under `{path}`**; JSON is used only for `{{path}}`. Prod
  `recipient_data` already carries object-valued variables (`product_details`, `items`,
  `cart_data_json`, `nearest_store` — 8.7% of sampled calls), so JSON-ifying `{path}` would have
  changed live output.
- Unresolvable spec tokens stay literal, so pseudo-JSON a prompt is describing
  (`{Gender: Female}`, `{HH:MM AM/PM}`, `{kb:company_name}` — all real prod shapes) is never
  deleted. Tested worst-case with every colon-prefix identifier also a passed variable: 0/24 mangled.

### Accepted trade-off
No blanket `{{` → `{` unescaping. Nested JSON ends in `}}`, so unescaping ate a closing brace off
every nested payload. A prompt that used `{{...}}` purely to escape a literal brace around
non-identifier content now keeps both braces (visible, not data loss); a double-braced *identifier*
still unescapes (`{{Unknown}}` → `{Unknown}`).

### Tests
49 tests in `tests/tests/test_render_prompt.py` covering the legacy table, JSON-in-prompt, both
syntaxes, nested paths, format specs, partial fill, and `call_sid`/`stream_sid` non-leakage through
both syntaxes. Suite: 863 passed; the 10 failures are pre-existing (verified by stashing this diff
and re-running — async-mock issues in `_log_transcriber_connection_error`).


bolna/helpers/utils.py
```diff
@@ -47,6 +47,188 @@ def __missing__(self, key):
...
+        double = match.group("double")
+        path = double if double is not None else match.group("single")
+        spec = match.group("spec")
+        found, value = resolve_variable_path(path, data)
+        if not found:
+            # Nested paths are the ones worth reporting: they fail silently and leave a literal
+            # token that only surfaces when the agent reads it aloud. Guarded to nested-only so
+            # ordinary {name} misses stay quiet.
+            if ("." in path or "[" in path) and missing is not None:
+                log_unresolved_nested_path(path, data)
```
SubhrajitPrusty: wont this be a very noisy log?

SubhrajitPrusty (commented):
LGTM, except the potential noisy log.

san0808 (approved):

PR #961, opened 2026-08-17
lid hand in message fix for web call

puneet-bolna:
### Why the gate isn't removed outright
  
  On `default` **off** the web path there is no single wire format to render against.
  `__setup_synthesizer` only forces `use_mulaw` in the telephony and web/freeswitch branches;
  that config falls through, so each synth keeps its own default — elevenlabs/cartesia `True`,
  maya/deepgram `False`, pixa/rime hardcoded `True`. Forcing it uniform would change behaviour
  for maya and deepgram. The wire-agreement guard above makes the mismatch unreachable regardless.
  
  ## Testing
  
  `tests/tests/test_handoff_prewarm.py` — 21 tests, 6 new in this round:
  
  - one-shot sentinel falls back to `synthesize()` instead of abandoning the label
  - a genuinely short fallback clip is still discarded
  - a synth whose `use_mulaw` disagrees with the call wire is skipped, not mislabeled
  - the wire decision tracks `SUPPORTED_OUTPUT_TELEPHONY_HANDLERS`
  - Cartesia one-shot sends the raw `pcm_s16le` output_format
  - `audio_to_pcm` rejects a positional target rate
  
  `ruff check .` and `ruff format --check .` clean. Suite: **1002 passed, 15 failed** — all 15
  pre-existing on master, verified against a `git archive master` export producing the identical
  15 (`test_event_injection_e2e.py` ×7, `test_expression_routing.py` ×3,
                                                                                                           copied 3443 chars to clipboard
────────────────────────────────────────────────────

## Testing

`tests/tests/test_handoff_prewarm.py` — 21 tests, 6 new in this round:

- one-shot sentinel falls back to `synthesize()` instead of abandoning the label
- a genuinely short fallback clip is still discarded
- a synth whose `use_mulaw` disagrees with the call wire is skipped, not mislabeled
- the wire decision tracks `SUPPORTED_OUTPUT_TELEPHONY_HANDLERS`
- Cartesia one-shot sends the raw `pcm_s16le` output_format
- `audio_to_pcm` rejects a positional target rate

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -6011,8 +6011,18 @@ class TaskManager(BaseManager):
             "sequence_id": -1,
             "message_category": "handoff",
             "text": handoff_text,
+            "type": "audio",
         }
-        clip = self.handoff_audio_cache.get(target)
+        # The cache holds mu-law@8k clips — telephony wire format. Web/freeswitch play raw
+        # PCM@24k, so the clip is unusable there (42b5f89b: silent handoff); live-synthesize.
+        mulaw_transport = self.tools["output"].get_provider() in (
+            TelephonyProvider.PLIVO.value,
+            TelephonyProvider.TWILIO.value,
+            TelephonyProvider.EXOTEL.value,
+            TelephonyProvider.VOBIZ.value,
+            TelephonyProvider.SIP_TRUNK.value,
+        )
+        clip = self.handoff_audio_cache.get(target) if mulaw_transport else None
         if clip:
             # Pre-warmed mu-law clip pushed straight to telephony. Both end-flags required
             handoff_meta.update(
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -96,6 +96,7 @@ from bolna.helpers.utils import (
     yield_chunks_from_memory,
     process_task_cancellation,
     audio_to_mulaw8k,
+    audio_to_pcm,
     pcm_to_ulaw,
     format_error_message,
     enrich_context_with_time_variables,
@@ -6013,21 +6014,13 @@ class TaskManager(BaseManager):
             "text": handoff_text,
             "type": "audio",
         }
-        # The cache holds mu-law@8k clips — telephony wire format. Web/freeswitch play raw
-        # PCM@24k, so the clip is unusable there (42b5f89b: silent handoff); live-synthesize.
-        mulaw_transport = self.tools["output"].get_provider() in (
-            TelephonyProvider.PLIVO.value,
-            TelephonyProvider.TWILIO.value,
-            TelephonyProvider.EXOTEL.value,
-            TelephonyProvider.VOBIZ.value,
-            TelephonyProvider.SIP_TRUNK.value,
-        )
-        clip = self.handoff_audio_cache.get(target) if mulaw_transport else None
+        # The prewarm rendered the clip in this call's wire format, so it is always usable.
+        clip = self.handoff_audio_cache.get(target)
         if clip:
-            # Pre-warmed mu-law clip pushed straight to telephony. Both end-flags required
+            # Pre-warmed wire-format clip pushed straight to the transport. Both end-flags required
             handoff_meta.update(
                 {
-                    "format": "mulaw",
+                    "format": "mulaw" if self.__handoff_mulaw_wire() else "pcm",
                     "end_of_llm_stream": True,
                     "end_of_synthesizer_stream": True,
                     "is_first_chunk": True,
@@ -6066,12 +6059,25 @@ class TaskManager(BaseManager):
             "{language}", LANGUAGE_NAMES.get(label, label)
         )
 
+    def __handoff_mulaw_wire(self) -> bool:
+        """Telephony transports take the mu-law@8k clip; web/freeswitch play raw PCM@24k."""
+        return self.tools["output"].get_provider() in (
+            TelephonyProvider.PLIVO.value,
+            TelephonyProvider.TWILIO.value,
+            TelephonyProvider.EXOTEL.value,
+            TelephonyProvider.VOBIZ.value,
+            TelephonyProvider.SIP_TRUNK.value,
+        )
+
     async def __prewarm_handoff_clips(self):
         """Pre-render each language's handoff on its own voice via one-shot synthesize().
         Clips are static for the call; non-active labels first (likely first targets)."""
         pool = self.tools.get("synthesizer")
         if not isinstance(pool, SynthesizerPool):
             return
+
+        # A call has exactly one transport, so render in its wire format only.
+        mulaw_wire = self.__handoff_mulaw_wire()
 
         async def render(label, synth):
             text = self.__handoff_text_for(label)
@@ -6081,6 +6087,7 @@ class TaskManager(BaseManager):
                 synth.__class__.__name__,
                 getattr(synth, "voice_id", None) or getattr(synth, "voice", None),
                 text,
+                "mulaw" if mulaw_wire else f"pcm{WEBCALL_TTS_SAMPLE_RATE}",
             )
             cached = HANDOFF_CLIP_CACHE.get(cache_key)
             if cached:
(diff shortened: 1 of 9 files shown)


SubhrajitPrusty (approved):

bolna/agent_manager/task_manager.py
```diff
@@ -6148,13 +6151,26 @@ def __handoff_text_for(self, label):
...
+
     async def __prewarm_handoff_clips(self):
         """Pre-render each language's handoff on its own voice via one-shot synthesize().
         Clips are static for the call; non-active labels first (likely first targets)."""
         pool = self.tools.get("synthesizer")
         if not isinstance(pool, SynthesizerPool):
             return
 
+        # A call has exactly one transport, so render in its wire format only.
+        mulaw_wire = self.__handoff_mulaw_wire()
```
san0808, line 6154: `__setup_synthesizer` only creates the prewarm task when the output provider is not `default`, so this PCM branch is reachable only from freeswitch. Was that gate the workaround for the mu-law mislabeling this PR fixes? If so can it come out here, or is there a reason browser calls on the default handler should keep paying a live TTS round trip on every switch?

bolna/agent_manager/task_manager.py
```diff
@@ -6148,13 +6151,26 @@ def __handoff_text_for(self, label):
             "{language}", LANGUAGE_NAMES.get(label, label)
         )
 
+    def __handoff_mulaw_wire(self) -> bool:
```
san0808, line 6142: This is the same five-provider set as `SUPPORTED_OUTPUT_TELEPHONY_HANDLERS` and `TelephonyProvider.telephony_values()`, and the third spelling of the wire decision in this file (`__setup_synthesizer` has its own `is_telephony` tuple, `__check_for_backchanneling` uses `is_web_based_call or provider == FREESWITCH`). Could it key off `SUPPORTED_OUTPUT_TELEPHONY_HANDLERS` so a sixth carrier cannot silently land in the PCM branch?

Related: for provider `default` with `is_web_based_call` False, `__setup_synthesizer` leaves ElevenLabs `use_mulaw` at its `True` default (streaming `ulaw_8000`) while this returns False, so the clip and the rest of the turn would be on different wire formats if the prewarm ever ran there.

bolna/agent_manager/task_manager.py
```diff
@@ -6163,23 +6179,35 @@ async def render(label, synth):
...
                     audio = await synth.synthesize(text)
                     if not audio:
                         return
                     # pydub decode shells out to ffprobe/ffmpeg — keep it off the event loop.
-                    clip = await asyncio.to_thread(self.__handoff_clip_to_mulaw, synth, audio)
+                    convert = self.__handoff_clip_to_mulaw if mulaw_wire else self.__handoff_clip_to_pcm
+                    clip = await asyncio.to_thread(convert, synth, audio)
+                # Some one-shots return a truthy error sentinel (e.g. deepgram's b"\x00" on
+                # non-200); anything under ~50ms can't be a handoff line — live-synth instead.
+                if clip and len(clip) < (400 if mulaw_wire else 2400):
```
san0808: This runs after the `if not clip:` decision, so a one-shot that returns a truthy sentinel skips the `synthesize()` plus converter fallback and then gets discarded here, leaving the label with no clip even though `synthesize()` would likely have produced a usable one. Would setting `clip = None` and falling through, or moving this check above the fallback, recover the prewarm instead of abandoning it?

bolna/agent_manager/task_manager.py
```diff
@@ -6192,6 +6220,17 @@ async def render(label, synth):
 
         await asyncio.gather(*(render(label, synth) for label, synth in pool.synthesizers.items()))
 
+    def __handoff_clip_to_pcm(self, synth, audio):
```
san0808: This is `__handoff_clip_to_mulaw` with one different call and one extra argument, down to the same rate-hint getattr chain and the same error string, so neither log says which wire failed. Worth folding into one `__handoff_clip_convert(synth, audio, mulaw_wire)`? The caller already selects between them with a ternary.

bolna/helpers/utils.py
```diff
@@ -973,11 +973,26 @@ def audio_to_mulaw8k(audio, rate_hint=8000, format_hint=""):
...
+def audio_to_mulaw8k(audio, rate_hint=8000, format_hint=""):
+    """One-shot synth output → mono 16-bit 8kHz mu-law, or None if undecodable."""
+    segment = decode_audio_segment(audio, rate_hint, format_hint)
+    if segment is None:
+        return None
     segment = segment.set_frame_rate(8000).set_channels(1).set_sample_width(2)
     return pcm_to_ulaw(segment.raw_data)
 
 
+def audio_to_pcm(audio, sample_rate, rate_hint=8000, format_hint=""):
```
san0808: The second positional parameter here is the target rate, but in `audio_to_mulaw8k` just above it is the source hint. `audio_to_pcm(audio, 8000)` written by analogy would resample to 8k and decode headerless input at the default 8000 hint, with nothing raising. Make `sample_rate` keyword-only, or name it `target_sample_rate`?

bolna/synthesizer/elevenlabs_synthesizer.py
```diff
@@ -96,6 +96,12 @@ def _get_output_format(self):
     async def synthesize(self, text):
         return await self._generate_http(text, format="mp3_44100_128")
 
+    async def synthesize_pcm_clip(self, text, sample_rate):
```
san0808, line 99: ElevenLabs is the only provider with a native PCM one-shot, so every other voice on the web path falls back to `synthesize()` plus a pydub decode. Cartesia returns mp3/44100 there, which means an ffmpeg shell out per language per call and no prewarm at all where ffmpeg is unavailable. Worth a matching `synthesize_pcm_clip` on Cartesia (`container: raw`, `encoding: pcm_s16le`) as a follow-up?

san0808 (commented):
Traced the freeswitch path and the fix looks right: `FreeSwitchOutputHandler.handle` gates on `meta_info.get("type") == "audio"`, so without it the clip was dropped while `is_final` still marked the turn complete. A few questions inline.

bolna/agent_manager/task_manager.py
puneet-bolna: added — != "default" gate

 On default off the web path, the call has no 
  single wire format to render against: __setup_synthesizer only forces use_mulaw in the telephony and web/freeswitch branches, so that    
  config falls through and every synth applies its own default — elevenlabs/cartesia True, maya/deepgram False, pixa/rime always True. A   
  per-call boolean can't describe a pool that disagrees with itself. Forcing use_mulaw=True there to make it uniform would flip maya and   
  deepgram, so I didn't.

bolna/agent_manager/task_manager.py
puneet-bolna: __handoff_mulaw_wire now keys off SUPPORTED_OUTPUT_TELEPHONY_HANDLERS, and I collapsed both is_telephony tuples in __setup_synthesizer   
  onto it too (lines 1691 and 1776 — identical sets). New test asserts the wire decision tracks the registry, so a sixth carrier can't     
  drift.

bolna/agent_manager/task_manager.py
puneet-bolna: Fixed by moving the check above the fallback.

bolna/agent_manager/task_manager.py
puneet-bolna: Folded into __handoff_clip_convert(synth, audio, mulaw_wire)

bolna/helpers/utils.py
puneet-bolna: Renamed and keyword-only, since the rename alone still leaves audio_to_pcm(audio, 8000) compiling. Docstring says why. Test asserts the positional form raises TypeError.

bolna/synthesizer/elevenlabs_synthesizer.py
puneet-bolna: synthesize_pcm_clip uses the same {"container": "raw", "encoding": "pcm_s16le", "sample_rate": …} its WS path already uses when use_mulaw is False; _generate_http gained an output_format override defaulting to the existing mp3/44100.

san0808 (commented):
All six earlier threads look resolved, thanks. Cartesia's one-shot matches `form_payload`'s non-mulaw branch exactly and `_generate_http` does no post-processing, so the raw bytes pass through clean. Two things left.

bolna/agent_manager/task_manager.py
```diff
@@ -6163,23 +6168,53 @@ async def render(label, synth):
...
+                "mulaw" if mulaw_wire else f"pcm{WEBCALL_TTS_SAMPLE_RATE}",
             )
             cached = HANDOFF_CLIP_CACHE.get(cache_key)
             if cached:
                 self.handoff_audio_cache[label] = cached
                 return
+            # A synth that disagrees with the call's wire format would be cached at one
+            # encoding and streamed at the other, so skip it rather than mislabel the clip.
+            synth_mulaw = getattr(synth, "use_mulaw", None)
+            if synth_mulaw is not None and bool(synth_mulaw) != mulaw_wire:
```
san0808: Is there a path this protects that I am missing? It looks unreachable on every config that gets here: telephony forces `use_mulaw=True` against `mulaw_wire` True, freeswitch and browser web force False against False, and the one unforced path (non-web `default`) is excluded by the gate in `__setup_synthesizer`.

The only synths that can trip it are `PixaSynthesizer` and `RimeSynthesizer`, which hardcode `self.use_mulaw = True` and ignore the kwarg, so on a web or freeswitch call this skips the prewarm and logs an `error` per language per call for a voice whose whole stream is already mu-law on a 24k PCM wire.

For the mislabeling case itself: `__handoff_clip_convert` normalizes anything decodable to the requested wire, `synthesize_telephony_clip` already returns None when `use_mulaw` is False, and `synthesize_pcm_clip` honors the requested rate regardless of it, so I could not find a combination that yields a mislabeled clip. My earlier note about the `default` divergence was an argument for the shared predicate you added rather than for a second check on top of it.

tests/tests/test_handoff_prewarm.py
```diff
@@ -25,7 +25,7 @@ def _wav_bytes(duration_ms=100, rate=16000):
 
 
 def _to_mulaw(synth, audio):
```
san0808, line 27: The module docstring above still says clips are pre-rendered as mu-law, which `test_prewarm_renders_pcm_for_non_mulaw_wire` and `test_clip_cache_keys_are_wire_specific` now contradict. Worth rewording it to say the render format follows the call's wire?

bolna/agent_manager/task_manager.py
puneet-bolna: No path it protects. Removed.

tests/tests/test_handoff_prewarm.py
puneet-bolna: done

san0808 (approved):
LGTM

PR #962, opened 2026-08-17
estimate for mark in last agent message for transcript

puneet-bolna:
Fix: last agent message truncated in transcript when playback marks are lost at call end

Problem

When a caller hangs up while (or just after) the agent's final message plays, the stored transcript can cut off mid-sentence even though the recording contains the full line. Prod example: last reply synthesized as 12 chunks — chunks 1–10 ACKed, chunk 11 (9.6s of audio) played fully but its mark never returned (marks lagged playback by 5.4s avg / 13.5s max across the call; 15 of 46 were never delivered). End-of-call sync_history trimmed the message to ACK-confirmed text, dropping the sentence the caller actually heard. This also degrades everything downstream of the transcript: conversation view, webhooks, and post-call extractions.

Root cause

The end-of-call trim treats "no ACK" as "not played". Telephony marks are delivery confirmations that arrive only after playout and are routinely late or lost when the call ends — so the trim systematically under-reports the final message on any call that ends before the tail marks arrive.

Fix

sync_history gains an opt-in extend_with_playback_estimate flag, passed only from the end-of-call sync. When ACKed text exists and unACKed chunks remain for the same turn, the tail is credited all-or-nothing: only if the wall-clock elapsed since the turn's last ACK covers the tail's entire audio duration — proof playback had time to finish. A mid-chunk hangup keeps today's strict ACK trim; we never stamp a partial guess of text that may not have been heard.

- MarkEventMetaData.get_last_ack_ts_for_turn(turn_id) — read-only helper returning the turn's last confirmed playback point.
- A chunk-boundary space is restored when joining (heard-text accumulators strip trailing spaces; without it the downstream exact-match trim silently drops the tail again).

Not changed

- All interruption/barge-in paths call sync_history with the flag default (False) — byte-identical behavior.
- Nothing during the live call changes; the extension runs only at teardown, after the conversation is over.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -2168,8 +2168,10 @@ class TaskManager(BaseManager):
             or (self.execute_function_call_task is not None and not self.execute_function_call_task.done()),
         }
 
-    async def sync_history(self, mark_events_data, interruption_processed_at):
-        """Sync history to reflect only what was actually spoken. Uses confirmed text or falls back to pending marks."""
+    async def sync_history(self, mark_events_data, interruption_processed_at, extend_with_playback_estimate=False):
+        """Sync history to reflect only what was actually spoken. Uses confirmed text or falls back to pending marks.
+        extend_with_playback_estimate: end-of-call only — credit the unACKed tail when the wall
+        clock since the last ACK covers its full duration (marks can lag playback and get lost)."""
         try:
             mark_events_data = list(mark_events_data)
             target_turn_id = self._get_latest_turn_id_from_marks(mark_events_data)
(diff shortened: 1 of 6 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -2211,6 +2213,31 @@ async def sync_history(self, mark_events_data, interruption_processed_at):
...
+                    if target_turn_id is not None and mark_data.get("turn_id") != target_turn_id:
+                        continue
+                    pending_tail.append({"text": text, "duration": mark_data.get("duration", 0)})
+                last_ack_ts = self.mark_event_meta_data.get_last_ack_ts_for_turn(target_turn_id)
+                if pending_tail and last_ack_ts:
+                    # All-or-nothing: credit only when playback provably finished; mid-chunk
+                    # hangup keeps the strict ACK trim (never stamp a partial guess).
+                    tail_play_time = max(0.0, interruption_processed_at - last_ack_ts)
+                    pending_duration = sum(chunk["duration"] or 0 for chunk in pending_tail)
+                    if tail_play_time >= pending_duration > 0:
```
san0808: I replayed the call from the linked issue against this branch. The turn's last ACK lands 6.97s before the terminal sync_history, and the unacked tail is a single chunk of 9.61s (plus a 0s final marker), so tail_play_time 6.97 < pending_duration 9.61 and this branch never runs. The stored transcript comes out identical to today.

The recipient hung up about 2.6s into that chunk, so it was partly heard, not fully. That is also the general shape of this bug: the caller drops during or just after the agent's closing line, which is exactly when elapsed time is shorter than the tail. Would proportional crediting fit better, reusing _trim_partial_to_complete_words the way the estimate branch below already does? 6.97/9.61 of the 133 chars would recover most of what was actually heard.

Can you re-check on that call before merging?

san0808 (commented):
Reviewed this against the call in the linked issue. The root cause reads right and the change itself is safe: opt-in, interruption paths untouched, and anchoring on the last ACK under-credits rather than over-credits, which is the correct direction.

Two things before merge:

1. Replaying that call, the gate does not fire, so the transcript there is still truncated at the same point. Numbers inline.
2. The description says the tail chunk "played fully but its mark never returned". The recipient hung up roughly 2.6s before it would have finished, so about 72% of it was heard, not all of it.

Separately, item 2 of the issue (the user message showing in the debug view but not the conversation view) is not covered here or on any other branch. Worth splitting into its own issue so it does not get closed with this one?

bolna/agent_manager/task_manager.py
```diff
@@ -2211,6 +2213,31 @@ async def sync_history(self, mark_events_data, interruption_processed_at):
             if response_heard:
                 logger.info(f"response_heard (last 10 chars): {response_heard[-10:]}")
 
+            if extend_with_playback_estimate and response_heard:
+                pending_tail = []
+                for mark_id, mark_data in mark_events_data:
+                    text = mark_data.get("text_synthesized", "")
+                    if mark_data.get("type") in ["pre_mark_message", "backchanneling"] or not text:
```
san0808: NON_EVIDENCE_MARK_TYPES is already imported and used in _get_latest_turn_id_from_marks just above. Use it here instead of repeating the tuple, so the two filters cannot drift?

bolna/helpers/mark_event_meta_data.py
```diff
@@ -249,6 +249,17 @@ def get_mark_tracking_summary(self) -> dict:
     def fetch_cleared_mark_event_data(self):
         return self.previous_mark_event_meta_data
 
+    def get_last_ack_ts_for_turn(self, turn_id) -> Optional[float]:
```
san0808, line 252: This class already keeps a playout clock in get_audio_playing_until(), added for the inactivity fix. Did that one not work for "has the queued audio finished", or is the per-turn anchor needed for a reason worth a comment here?

bolna/agent_manager/task_manager.py

bolna/helpers/mark_event_meta_data.py
puneet-bolna: resolved

san0808 (approved):

PR #967, opened 2026-08-18
fixed lid in false interruption

puneet-bolna:
Problem

Two production failures on the same class of calls (QA: b381ba0a, 75f79ad0, bcced1f3 — barcode/serial readouts and multi-part utterances):

The judge switches language on code readouts. On b381ba0a the caller was reading a cooler barcode; the judge switched hi→en at 0.92 confidence with reasoning "Caller speaking English ('This B1')". Prompt rule 3a explicitly forbids switching on a predominantly numeric/alphanumeric turn ("this precedence is absolute") — the model ignored its own rule because the fragment parses as English. A Hindi-speaking caller was answered in English mid-readout. VBL came off the LID exclusion list on 10 Aug and its core flow is asset-tag readouts, so this shape is now common.
One logical answer shatters across many synthesizer turns. A final transcript that lands while a response is in flight triggers cleanup_before_user_append: cancel LLM + synth, send telephony clear, regenerate — with no hysteresis. A caller reading a code produces a final every ~1–2s ("बारकोड, ठीक।" → "ये B1।" → "V1" → "21, 65, 11, 69"), so the pipeline loops cut→regen→cut: b381ba0a produced 27 LLM turns (5 cancelled) in 140s, and the caller heard mid-word fragments ("ठीक है, मैं आपका support ticket rais") with no interruption on their side. On 75f79ad0 the instant regeneration re-asked the identical question into the caller's still-ongoing speech.
Change

1. Rule-3a code enforcement (is_alphanumeric_readout() + veto in __run_language_switch)

A turn with ≥1 digit-bearing token and ≤2 other tokens is a data readout → the switch is vetoed after the substance gate, emitting a new gated:alphanumeric_readout telemetry outcome (countable in lid_shadow_events).
Reuses the substance gate's explicit_bypass, so a rule-1 explicit by-name request keeps precedence.
Bare letters deliberately do not count as code tokens — "I want English" must never trip the veto (regression-tested).
Veto direction is fail-safe: worst case a legitimate switch waits for the caller's next substantive turn.

2. Settle-window regeneration (LLM_REGEN_SETTLE_S = 0.7 + debounce in _handle_transcriber_output)

The llm-branch kickoff is extracted into kickoff_llm_generation() (verbatim old behavior). A final that overlaps an in-flight response now merges (existing pop_and_merge_user), cleans up as before, then arms a 0.7s settle timer instead of regenerating immediately; further finals inside the window re-merge and re-arm; quiet ⇒ one regeneration against the full merged utterance.
A live settle timer itself counts as overlap (a second final merges into the pending regen instead of racing it; when nothing is in flight, it merges without cleanup).
0.7s sits just under Deepgram's 1s utterance-end floor, so a continuing speaker always re-arms before the timer fires.
Supersede semantics: __cleanup_downstream_tasks (barge-in, LID-switch truncation) cancels a pending regen; the overlapped path re-arms after cleanup, so the newest turn always wins; teardown cancels the timer.
Non-overlapped finals keep the existing zero-latency path — the window only exists when the caller's speech provably overlapped a response. Finals during tool calls were already deferred upstream and are untouched.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -158,8 +158,7 @@ def _inject_end_call_tool(api_tools, *, scope, nodes, description=None):
 
 
 def is_alphanumeric_readout(text: str) -> bool:
-    """Code/number readout, not language evidence (judge rule 3a): ≥1 digit-bearing token and
-    ≤2 others. Bare letters never count, so "I want English" must not trip this."""
+    """Code readout, not language evidence (rule 3a): ≥1 digit-bearing token, ≤2 others."""
     tokens = re.findall(r"\w+", text or "", flags=re.UNICODE)
     if not tokens:
         return False
@@ -601,8 +600,7 @@ class TaskManager(BaseManager):
         # Records every language switch — manual tool call (legacy) or LLM-driven
         # (triggered_by="lid_llm") — used post-call for precision / latency analysis.
         self.language_switch_events: list[dict] = []
-        # Regeneration debounce for finals that overlap an in-flight response — one
-        # regen per merged utterance instead of a cut/regenerate cycle per fragment.
+        # Debounce for overlapped finals: one regen per merged utterance, not per fragment.
         self.regen_settle_task = None
         self.regen_settle_payload = None
         # Legacy-flow handoff state (populated by __inject_switch_language_tool
@@ -4521,8 +4519,7 @@ class TaskManager(BaseManager):
             # Re-register the seq_id allocated by __get_updated_meta_info, else the audio blocks.
             self.interruption_manager.revalidate_sequence_id(meta_info["sequence_id"])
 
-        # Unconditional: if the old task completed and the interruption path invalidated
-        # responses, the new seq_id would never be re-added and all audio would stay BLOCKed.
+        # Unconditional: an un-re-added seq_id leaves every chunk of this turn BLOCKed.
         self.interruption_manager.revalidate_sequence_id(meta_info["sequence_id"])
         self.response_in_pipeline = True
         # Background once-per-turn switch decision; gates this turn's AUDIO, not its generation.
@@ -4534,8 +4531,7 @@ class TaskManager(BaseManager):
         return self.regen_settle_task is not None and not self.regen_settle_task.done()
 
     def regen_settle_can_fire(self):
-        """False when the active transcriber is on the exclusion list — its endpointing means no
-        follow-up final can land inside the window, so waiting only adds latency."""
+        """False for excluded transcribers: no final can land inside the window, so waiting only costs."""
         transcriber = self.tools.get("transcriber")
         active = (
             transcriber.transcribers.get(transcriber.active_label, transcriber)
@@ -4637,8 +4633,7 @@ class TaskManager(BaseManager):
                 logger.info(f"Merged transcript with unheard response: {transcriber_message}")
             if any(activity.values()):
                 await self.__cleanup_downstream_tasks()
-            # Cleanup invalidates every pending seq id; re-register this turn's or _synthesize
-            # later drops the fresh response as an invalid sequence.
+            # Cleanup invalidated every pending seq id; without this _synthesize drops this turn.
             self.interruption_manager.revalidate_sequence_id(current_sequence_id)
             logger.info(
                 "BOLNA_TRACE_TM revalidated_current_seq_after_cleanup seq=%s turn=%s response_uid=%s",
@@ -4666,8 +4661,7 @@ class TaskManager(BaseManager):
         if next_task == "llm":
             meta_info["origin"] = "transcriber"
             if overlapped and (self.regen_settle_armed() or self.regen_settle_can_fire()):
-                # More finals are likely coming — regenerate once after they settle.
-                # An already-armed window always absorbs the final so no payload strands.
+                # More finals likely coming; an armed window always absorbs one, so none strands.
                 self.arm_regen_settle(transcriber_message, meta_info)
             else:
                 self.kickoff_llm_generation(transcriber_message, meta_info)
@@ -4876,8 +4870,7 @@ class TaskManager(BaseManager):
                         interim_transcript_len += len(message["data"].get("content").strip().split(" "))
                         transcript_content = message["data"].get("content", "")
 
-                        # Late re-delivery of a transcript already being processed (or owed a
-                        # regen by the armed settle window) is not new user speech.
+                        # Re-delivery of a transcript already processed or owed a regen isn't new speech.
                         if (
                             self.response_in_pipeline or self.regen_settle_armed()
                         ) and self.conversation_history.is_duplicate_user(transcript_content):
@@ -5097,16 +5090,14 @@ class TaskManager(BaseManager):
                         )
 
                         if was_eager and self.eager_llm_task is not None and self.regen_settle_armed():
-                            # A regen is owed for the merged turn — drop the eager reply (generated
-                            # without it) and let this final merge into the pending window instead.
+                            # A regen is owed the merged turn, so drop the eager reply built without it.
                             logger.info("EagerEOT reply dropped: settle window armed — merging final into regen")
                             self.eager_llm_task.cancel()
                             self.eager_llm_task = None
                             eager_stub_text = (self.eager_meta_info or {}).get("eager_transcript")
                             self.eager_meta_info = None
                             self.eager_history_snapshot = None
-                            # Pop only the eager stub; a merged turn (stub + later final) reads
-                            # differently and must survive — it's what the pending regen answers.
+                            # Pop only the stub: a merged turn reads differently and the regen answers it.
                             last_row = self.history[-1] if self.history else None
                             if (
                                 last_row
@@ -5771,8 +5762,7 @@ class TaskManager(BaseManager):
             return
         # One selection, used by BOTH the speculative copy and the real history append —
         idle_flush_user_text = trailing_utterance_text(detector_segments) or detector_transcript
-        # Snapshot before the multi-second decide: if a user turn lands meanwhile, the
-        # idle-flush append below would duplicate it (prod 43571cba / 8e98dbc2).
+        # Pre-decide snapshot: a turn landing meanwhile would be duplicated below.
         history_signature_at_decide = self.conversation_history.user_turn_signature()
         active = self.language
 
@@ -5996,8 +5986,7 @@ class TaskManager(BaseManager):
             emit_lid_decision("gated:short_audio")
             return
 
-        # Rule-3a backstop: the judge can still read "This B1" as English (prod b381ba0a);
-        # explicit by-name requests keep rule-1 precedence via the substance gate's bypass.
+        # Rule-3a backstop: the judge still reads "This B1" as English.
         if not explicit_bypass and is_alphanumeric_readout(detector_transcript):
             logger.info(
                 f"LanguageSwitcher: target '{target}' vetoed — rule-3a alphanumeric readout "
@@ -6086,8 +6075,7 @@ class TaskManager(BaseManager):
                     f"LanguageSwitcher: corrected user turn to detector transcript {detector_transcript[:80]!r}"
                 )
         elif self.conversation_history.user_turn_signature() != history_signature_at_decide:
-            # A main turn for this audio landed during the decide — appending would duplicate
-            # it and re-route the graph on phantom input; answer the real turn instead.
+            # A main turn landed during the decide; appending would re-route on phantom input.
             transcript_corrected = False
             logger.info(
                 "LanguageSwitcher: idle-flush skipped — user turn arrived during decide; "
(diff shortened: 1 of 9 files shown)


bolna/constants.py
```diff
@@ -48,6 +48,10 @@
 # treated as stale — the detector hears the same audio and has produced nothing that long.
 LANGUAGE_SWITCH_SPEAKING_STALE_CAP_S = 2.5
 
+# Overlapped finals (multi-part utterances) regenerate once after this quiet window instead of
+# per fragment; sized under Deepgram's 1s utterance_end floor so a continuing speaker re-arms it.
+LLM_REGEN_SETTLE_S = 0.7
```
san0808, line 52: On the call in the description the overlapped finals landed 1.19s, 1.80s, 3.02s, 4.15s and 9.64s after the preceding final, so a 0.7s timer fires before the next fragment in every one of those cases and each of those turns just pays the extra 700ms before the same cut and regenerate happens anyway. endpointing defaults to 400ms, so a final already needs ~400ms of trailing silence and the next one cannot physically land inside 700ms unless the next fragment is under ~300ms of speech. Should the window be sized off the measured gap distribution, and if that comes out at 1.5s or more, is a debounce still the mechanism we want?

bolna/agent_manager/task_manager.py
```diff
@@ -4403,6 +4424,58 @@ def _retire_dropped_response(self, meta_info, reason):
...
+        # Background once-per-turn switch decision; gates this turn's AUDIO, not its generation.
+        self._spawn_language_switch_decision(transcriber_message, meta_info)
+        self.llm_task = asyncio.create_task(self._run_llm_task(transcriber_package))
+
+    def arm_regen_settle(self, transcriber_message, meta_info):
+        """(Re)arm the regeneration debounce with the latest merged turn."""
+        if self.regen_settle_task is not None and not self.regen_settle_task.done():
+            self.regen_settle_task.cancel()
+        self.regen_settle_payload = (transcriber_message, meta_info)
+        self.regen_settle_task = asyncio.create_task(self.__regen_after_settle())
```
san0808, line 4548: Arming the timer leaves response_in_pipeline False (cleanup just cleared it) with no audio playing, so for the length of the window the pipeline looks idle to everything except the new check in _handle_transcriber_output. The speculative gate in _listen_transcriber is exactly `not is_audio_being_played_to_user() and not self.response_in_pipeline`, so EagerEOT can start a second generation inside the window, the was_eager branch adopts it as llm_task without cancelling regen_settle_task, and the timer then fires and regenerates over it; if that eager reply already started playing, the regen path never calls __cleanup_downstream_tasks so nothing clears the telephony buffer and the caller hears both. Should the armed timer be folded into _inflight_response_activity (or set response_in_pipeline) so the eager gate and the inactivity watchdogs see it too?

bolna/agent_manager/task_manager.py
```diff
@@ -2422,6 +2438,11 @@ async def __cleanup_downstream_tasks(self):
         logger.info(f"Cleaning up downstream task")
         start_time = time.time()
         self._cancel_in_flight_llm_response()
+        # A pending regen belongs to the turn being torn down; the overlapped final
+        # path re-arms after this cleanup, so the newest turn always wins.
+        if self.regen_settle_task is not None and not self.regen_settle_task.done():
+            self.regen_settle_task.cancel()
+        self.regen_settle_payload = None
```
san0808, line 2508: The re-arm after this cleanup only happens when a later final reaches the overlapped branch, and there is at least one path where it never does: during the window response_in_pipeline is False, so the interim guard for a Deepgram late re-delivery (which is gated on response_in_pipeline) misses, the duplicate interim falls into the continuation branch and calls this cleanup, and the duplicate final that follows is then dropped by is_duplicate_user. The merged user turn is left in history with nothing generating for it, so the agent goes silent until repeat_after_silence or the stall backstop. Should the payload survive a cleanup that has no guaranteed re-arm?

bolna/agent_manager/task_manager.py
```diff
@@ -4486,41 +4561,20 @@ async def _handle_transcriber_output(self, next_task, transcriber_message, meta_
...
-            # task so it never delays the main LLM. When the detector already tagged this
-            # Gates this turn's AUDIO (not its generation) when the detector disagrees with the
-            # active language — see _spawn_language_switch_decision, which arms it for both this
-            # path and the eager one.
-            self._spawn_language_switch_decision(transcriber_message, meta_info)
-            self.llm_task = asyncio.create_task(self._run_llm_task(transcriber_package))
+            if overlapped:
+                # Overlapped speech means more finals are likely coming — regenerate once
+                # after they settle instead of a cut/regenerate cycle per fragment.
+                self.arm_regen_settle(transcriber_message, meta_info)
```
san0808, line 4665: On the call in the description a large share of these overlapping finals are not caller fragments: after each language switch the standby transcriber in the pool keeps emitting into the shared transcriber_output_queue, so the same utterance arrives again seconds later under the non-active label (one Hindi sentence came through three times over 30s). pop_and_merge_user on a stale duplicate concatenates the same sentence twice instead of assembling a fragment, and no debounce window merges finals that are 8s apart. Should we filter pool output by active_label first (the closed #957) and then measure how much of this churn is actually left to debounce?

bolna/agent_manager/task_manager.py
```diff
@@ -148,6 +150,16 @@ def _inject_end_call_tool(api_tools, *, scope, nodes, description=None):
...
 
 
+def is_alphanumeric_readout(text: str) -> bool:
+    """Code/number readout, not language evidence (judge rule 3a): ≥1 digit-bearing token and
+    ≤2 others. Bare letters never count, so "I want English" must not trip this."""
+    tokens = re.findall(r"\w+", text or "", flags=re.UNICODE)
+    if not tokens:
+        return False
+    code_tokens = [t for t in tokens if any(ch.isdigit() for ch in t)]
+    return len(code_tokens) >= 1 and (len(tokens) - len(code_tokens)) <= 2
```
san0808, line 166: This needs a digit-bearing token, but the LID tap frequently renders spoken digits as words: on the same call the readout reached the judge as "Five." and "Double one.", which returns False here (the judge happened to apply rule 3a correctly on its own for that one). The other-token bar is also an absolute count rather than a ratio, so "ye B1" is vetoed while "barcode, theek. ye B1" is not, which makes the veto depend on how many segments happened to accumulate before the buffer was drained. Should this be a ratio over the foreign-language segments instead of a fixed count?

puneet-bolna: intentional — the veto is a cheap pre-gate that only fires when it's unambiguously a readout, everything murkier falls to the judge which has the same rule 3a (and got "Five." right on that call). a ratio widens what the veto eats and I'm more scared of false-vetoing a real switch request that happens to contain number words than of the judge having to do its job. if we start seeing the judge blow the fall-through cases I'll revisit.

bolna/agent_manager/task_manager.py
puneet-bolna: the late-delivery guard only looked at response_in_pipeline so a dup interim during the window walked right past it into the continuation cleanup. guard is now response_in_pipeline or regen_settle_armed(). checked the non-dup case too — if a genuinely continuing interim kills the window, the merged text is already in history so the new final regenerates over it, nothing lost. only the duplicate path could strand the turn and that's closed now.

bolna/agent_manager/task_manager.py
puneet-bolna: added a regen_settle_armed() check to the eager start gate so nothing spawns inside the window, and in the was_eager adoption branch if a window is armed I drop the eager reply instead of adopting it (it was generated without the merged turn anyway) and let the final fall through to the normal path so it merges into the pending regen. one response, full context.

bolna/constants.py
puneet-bolna: the deepgram math checks out, but the 0.7 wasn't sized for deepgram — soniox emits finals much faster. on 241491f8 two finals landed inside the window and merged into one generation, so it does earn its keep there. if I size it off the measured gaps (1.2-1.8s) then every overlapped turn eats that as latency, which is worse than the occasional extra cut. but I think your point 4 below is the real answer — most of this churn is stale pool duplicates that no debounce can merge anyway. want to filter those first and then relook at the distribution, if gaps are still >1s after that then a fixed timer is the wrong tool and it should be activity based (hold while interims are still coming).

we can re visit this more over a hack in these cases

bolna/agent_manager/task_manager.py
san0808: Verified on the new head: the start gate and the drop branch both close it, and a final arriving after the drop merges into the pending window as intended. Resolving this one. I raised one thing about the history revert inside the drop separately.

bolna/agent_manager/task_manager.py
san0808: Verified. A duplicate interim now short circuits before the continuation branch, and I agree the non-duplicate case is safe since the merged text is already in history. Resolving.

bolna/agent_manager/task_manager.py
san0808: Fair, the fail-safe direction is the right default here and the judge did apply rule 3a itself on the word-rendered readout. Happy to leave it as a cheap pre-gate.

bolna/agent_manager/task_manager.py
```diff
@@ -5038,6 +5092,19 @@ async def _listen_transcriber(self):
...
+                            # A regen is owed for the merged turn — drop the eager reply (generated
+                            # without it) and let this final merge into the pending window instead.
+                            logger.info("EagerEOT reply dropped: settle window armed — merging final into regen")
+                            self.eager_llm_task.cancel()
+                            self.eager_llm_task = None
+                            self.eager_meta_info = None
+                            snapshot = getattr(self, "eager_history_snapshot", None)
+                            self.eager_history_snapshot = None
+                            if snapshot is not None and len(self.history) > snapshot:
+                                self.history = self.history[:snapshot]
```
san0808: history is a property over conversation_history._messages, so this truncates the real history. The snapshot was taken when the eager task started, which is necessarily before this window was armed, and arming runs pop_and_merge_user plus append_user, so the row sitting above the snapshot is the merged user turn the pending regen exists to answer. Once it is gone, _handle_transcriber_output re-appends only the current final, so the earlier fragment is dropped from what the LLM sees. Should this pop just the eager row (matched by content or asr_turn_id) rather than truncating back to the snapshot?

bolna/constants.py
san0808: Agreed on the sequencing: filter the pool duplicates first, then re-measure the distribution. One ask meanwhile, should the fixed timer be scoped to the providers where it can actually fire (or made interim driven as you suggest) rather than every overlapped Deepgram turn paying 700ms for a merge that cannot happen at 400ms endpointing?

puneet-bolna: done, went with the dumb version on purpose — an exclusion list (deepgram for now) checked against the active pool member, so deepgram's overlapped finals kick off immediately and soniox keeps the window. didn't want endpointing math with per-provider unit quirks for something the interim-driven rewrite replaces anyway after we re-measure on clean data. one guard: an already-armed window still absorbs finals regardless of the list, so nothing strands if the active transcriber changes mid-call.

bolna/agent_manager/task_manager.py
puneet-bolna: right, the snapshot predates the arming so the truncation was eating the merged turn too. took your suggestion — pop just the eager row: the stub text rides on eager_meta_info and the drop branch removes the last row only when it matches exactly. a merged row is always stub + final so it never matches and survives. if the stub already got merged the pop is a no-op, which is the correct outcome there too.

bolna/agent_manager/task_manager.py
puneet-bolna: yes — checked the code, pool output carries no label at all today so the consumer can't even tell a stale standby final from a live one. plan: stamp the label at the pool boundary (queue proxy, one choke point, no per-provider changes), drop non-active transcript packets at consumption with a counter log so we get the churn number you're asking for. one subtlety: pre-switch speech only exists in the old transcriber's late finals, but the switch path already commits that utterance from the detector transcript, so the drop doesn't lose i

can take this as separate PR?

bolna/agent_manager/task_manager.py
san0808: Verified: the content match is exact against the row the eager path appends, a merged row reads as stub plus final so it never matches and survives, and an already-merged stub makes the pop a no-op. Resolving.

bolna/constants.py
san0808: Verified. The active-member read plus the already-armed absorb covers the mid-call pool switch, and DeepgramTranscriber matches the prefix so flux and nova both skip the window. Good enough for now given the interim-driven rewrite replaces it. Resolving.

bolna/agent_manager/task_manager.py
san0808: Separate PR is the right call, this one is already carrying two unrelated fixes. Stamping the label at the pool boundary and dropping non-active packets at consumption with a counter is exactly the shape I would want, and agreed on the pre-switch case since the switch path commits that utterance from the detector transcript. Resolving here and following it there.

san0808 (approved):

PR #968, opened 2026-08-20
feat: add Telnyx as a native telephony provider

### LINKED ISSUE #867: Native Telnyx Telephony Provider Support
Hi team,

Would you consider adding Telnyx as a first-class telephony provider alongside Twilio and Plivo?

Telnyx is widely used in the AI voice ecosystem and offers competitive pricing, SIP trunking, media streaming, and phone number availability in regions where Twilio coverage can be limited.

I noticed the open-source repository already mentions Telnyx as a potential telephony extension, and the platform currently supports providers such as Twilio and Plivo. Native Telnyx support would make adoption easier for teams already using Telnyx infrastructure and would eliminate the need for custom forks and integrations.

Thanks!

mahi0601:
## Summary

Bolna currently supports Twilio, Plivo, Exotel, Vobiz, and Asterisk SIP trunks for phone calls, but not Telnyx. This adds Telnyx as a first-class `TelephonyProvider`, following the exact same registration pattern the existing providers use (`SUPPORTED_INPUT_HANDLERS`/`SUPPORTED_OUTPUT_HANDLERS` in `providers.py`, `TelephonyProvider.telephony_providers()`/`mulaw_providers()` in `enums.py`), so no new branching logic is needed elsewhere - `task_manager.py`'s mulaw/audio-format selection and `models.py`'s provider validation pick it up automatically.

Closes #867

## What's here

- `bolna/input_handlers/telephony_providers/telnyx.py` - parses Telnyx Media Streaming's `start`/`media`/`mark`/`stop` events
- `bolna/output_handlers/telephony_providers/telnyx.py` - formats outgoing `media`/`mark`/`clear` events, converting PCM to mulaw
- Registration in `enums.py`, `providers.py`, `input_handlers/__init__.py`, `output_handlers/__init__.py`
- Two mulaw-provider lists in `task_manager.py` updated so TTS output is correctly forced to mulaw@8kHz for Telnyx calls
- `tests/tests/test_telnyx_provider.py` - 8 tests covering registration, call-start parsing, and outbound message formatting

## Two protocol details worth flagging for review

Telnyx's Media Streaming protocol looks like Twilio's on the surface (same event names), but differs in two ways. I verified both against [Telnyx's docs](https://developers.telnyx.com/api-reference/websockets/stream-call-media-over-websocket) and cross-checked against pipecat's production `TelnyxFrameSerializer` rather than assuming Twilio parity:

1. The call id lives at `start.call_control_id`, with the stream id at the **top level** of every frame as `stream_id` - not nested inside `start` like Twilio's `streamSid`.
2. Outbound `media`/`mark`/`clear` frames carry **no session id at all** - Twilio/Plivo/Vobiz all stamp their session id into every outgoing frame, but Telnyx's spec (and pipecat's implementation) send just the bare event + payload, since the websocket connection itself identifies the stream. Called this out in a docstring on the output handler so it doesn't read as an oversight.

I did **not** override `requires_custom_voicemail_detection()` to `False` (i.e. left it at the base class's `True` default, same as Plivo/Vobiz) since I haven't verified Telnyx's native AMD meets the contract that flag implies for Twilio/Exotel/sip-trunk - happy to wire that in as a follow-up if a maintainer confirms it's safe.

## Deliberately out of scope

A `local_setup/telephony_server/telnyx_api_server.py` dev script (mirroring `twilio_api_server.py`) for originating outbound calls. Telnyx originates calls via Call Control REST + webhook events rather than Twilio's synchronous TwiML response - different enough surface that it deserves its own PR and review rather than being bundled in here.

## Testing

- `pytest tests/` - 970 passed (962 pre-existing baseline + 8 new), same 38 pre-existing unrelated failures as unpatched master, no regressions
- `ruff check` / `ruff format --check` - clean on all changed files


bolna/output_handlers/telephony_providers/telnyx.py
```diff
@@ -0,0 +1,46 @@
...
+
+        super().__init__(io_provider, websocket, mark_event_meta_data, log_dir_name)
+        self.is_chunking_supported = True
+
+    async def handle_interruption(self):
+        if self._closed:
+            return
+        try:
+            logger.info("interrupting because user spoke in between")
+            await self.websocket.send_text(json.dumps({"event": "clear"}))
```
san0808: Should this go through `self._send_text()` rather than `self.websocket.send_text()`? #970 wrapped every telephony send in a 5s `asyncio.wait_for` because a carrier socket can go half-dead, at which point `send_text` never returns and never raises. `__cleanup_downstream_tasks` awaits `handle_interruption` first with nothing above it bounding the wait, so a Plivo call froze there for ~10.5 minutes. Adding `TelnyxOutputHandler` to the parametrize list in `tests/tests/test_telephony_output_send_timeout.py` fails on this line with a `TimeoutError` while Twilio passes.

bolna/output_handlers/telephony_providers/telnyx.py
```diff
@@ -0,0 +1,46 @@
...
+            self.mark_event_meta_data.clear_data()
+        except Exception as e:
+            logger.info(f"WebSocket closed during interruption: {e}")
+            self._closed = True
+
+    async def form_media_message(self, audio_data, audio_format="wav"):
+        if audio_format != "mulaw":
+            audio_data = audioop.lin2ulaw(audio_data, 2)
```
san0808, line 40: Telnyx documents a minimum outbound chunk that the other providers don't: "Provided chunks of audio can be in a size of 20 milliseconds to 30 seconds" (https://developers.telnyx.com/docs/voice/programmable-voice/media-streaming). `output_chunk_size` is 4096, so 512ms of mulaw@8k for the body, but the tail slice is `len % 4096` and the base handler also pads a 1-byte chunk to 2 bytes before sending. Do those undersized frames need a floor here, or does Telnyx tolerate them?

bolna/input_handlers/telephony_providers/telnyx.py
```diff
@@ -0,0 +1,42 @@
...
+from dotenv import load_dotenv
+from bolna.helpers.logger_config import configure_logger
+
+logger = configure_logger(__name__)
+load_dotenv()
+
+
+class TelnyxInputHandler(TelephonyInputHandler):
+    """Telnyx Media Streaming (https://developers.telnyx.com/docs/voice/programmable-voice/media-streaming)
+    uses the same start/media/mark/stop event shape as Twilio, but nests the call id
```
san0808, line 11: Telnyx adds a caveat Twilio doesn't: "The order of events is not guaranteed and the chunk number can be used to reorder the events" (https://developers.telnyx.com/docs/voice/programmable-voice/media-streaming). The base `_listen` appends media in arrival order and ships every 10 frames without reading `chunk`. Safe to rely on because it is a single TCP websocket, or worth reordering?

bolna/output_handlers/telephony_providers/telnyx.py
mahi0601: I didn't check the floor against the tail-chunk case. Let me check what pipecat's TelnyxFrameSerializer does with sub-20ms remainders and confirm with Telnyx's behavior before deciding between padding-to-floor vs. carrying the remainder into the next chunk. Will follow up with a fix or an explicit doc note.

bolna/input_handlers/telephony_providers/telnyx.py
mahi0601: Since it's one TCP connection per call, in-order delivery is guaranteed at the socket layer regardless of what Telnyx's backend does upstream — the chunk field's "reorder" caveat should only matter if there were multiple connections or app-level buffering in play, neither of which apply here. I didn't see pipecat's serializer use chunk for reordering either, only for gap detection potentially. Happy to add a gap-detection log if you think dropped-frame visibility is worth it, but I don't think reordering logic is needed.

bolna/output_handlers/telephony_providers/telnyx.py
mahi0601: Good catch, missed that this needs to go through _send_text() like the rest of the handler. Fixing — will match the pattern used for media/mark and make sure the timeout test passes.

PR #971, opened 2026-08-20
feat: track whether the caller ever spoke on a call

MujeerAhmed:
Linear: BOLNA-2464

## What

Adds `TaskManager.user_spoke` — `False` by default, flipped to `True` the moment the caller's own words land in conversation history as a user turn. It ships on the conversation task's output dict as `user_spoke`, next to `ended_by_assistant`.

Set at every path where a real caller turn is appended:
- the streaming transcriber turn in `_handle_transcriber_output`
- the LanguageSwitcher idle-flush turn (locked-pool ASR produced nothing, the detector heard the caller)
- the s2s realtime `InputTranscript` event
- the turn-based / text conversation path

Deliberately **not** set by synthetic user turns — the `[silence] User was silent for Ns` message the stall watchdog injects via `_inject_and_run_llm` is a nudge to the LLM, not the caller speaking. So a call where only the agent ever spoke stays `False`.

## Why

Post-call analysis that needs a two-sided transcript (dispositions, in the dashboard-backend) currently runs on calls where the callee never said a word, so the whole transcript is the agent's opening line. Any verdict from that is noise. Consumers need a reliable "was this one-sided?" signal, and inferring it from the transcript can't distinguish a real user turn from an injected one.

Paired with bolna-ai/dashboard-backend, which uses the flag to skip dispositions on one-sided calls for opted-in accounts.

## Notes

Read-only addition — nothing branches on the flag inside bolna.


bolna/agent_manager/task_manager.py
puneet-bolna, line 5122: This path never sets user_spoke, so a caller who did speak comes out as False.
 When EndOfTurn arrives after an eager EOT, we reuse the speculative LLM and skip _handle_transcriber_output entirely (your own comment on line 5145 says as much) — and that method is the only place the flag gets set. Meanwhile the caller's turn is already sitting in history from line 5022, since self.history is just conversation_history.messages (property on 1144). So the transcript has both sides but the flag says one-sided, and dashboard-backend ends up skipping dispositions on a real conversation.

bolna/agent_manager/task_manager.py
```diff
@@ -3985,6 +3989,7 @@ async def _process_conversation_task(self, message, sequence, meta_info):
         self._append_eager_llm_stub(meta_info)
 
         if self.turn_based_conversation:
+            self.user_spoke = True
```
puneet-bolna: Minor, but in turn-based mode this catches turns the caller didn't make.

It runs for every _process_conversation_task, and two of those aren't the user: the [silence] nudge from _inject_and_run_llm (7181), which the PR description says should stay excluded, and _generate_proactive (2939), which fires with empty content and whose comment literally says "without adding a user message". The nudge is reachable here because the watchdog gate on 8190 is not turn_based or enforce_streaming.

MujeerAhmed: moved it out of _process_conversation_task to _listen_llm_input_queue — that only ever carries real typed input, so the silence nudge and _generate_proactive both drop out.

bolna/agent_manager/task_manager.py
MujeerAhmed: moved it to where eager_user_row gets appended, so the flag lands whether or not EndOfTurn goes through _handle_transcriber_output.

puneet-bolna (approved):

PR #972, opened 2026-08-21
perf: trim the graph routing hop and keep its prompt prefix cacheable

san0808:
The routing hop runs serially before the answering LLM on every turn of a graph agent, and it is a non-streaming full completion, so it pays for every output token rather than just time-to-first-token. Measured across two days of production graph-agent traffic (26,880 routing calls), it is p50 716ms, p90 1002ms, against a p50 369ms time-to-first-token for the answering call it precedes.

Two things were making it worse than it needs to be.

**The rationale is generated after the decision it explains.** Routing is a tool call with `tool_choice: "required"`, so the decision is the function name. `reasoning` and `confidence` are arguments inside that function, emitted after the name is already committed. The rationale therefore cannot influence the choice, and nothing downstream reads it except the observability payload. It is the median 110 characters of a 43-token response, roughly 64% of the output tokens, at about 6 to 7ms per token. It is now opt-in via `GRAPH_ROUTING_REASONING=true` for debugging a misrouting agent; `confidence` stays, since it is a couple of tokens.

**The routing prompt busts its own prefix cache every turn.** `_enrich_routing_context` refreshes time variables live and writes `_node_turns`, `_total_turns` and `_silence_repeats` into `context_data` each turn, and `_decide_next_node_llm` serialised all of that into the middle of the system prompt while substituting routing instructions against live time. A prompt referencing `{current_time}` diverged 15% of the way in, on every turn. The answering prompt already solved this by rendering through `_prompt_context()`, which freezes time per call; routing now does the same, and the internal bookkeeping keys stay out of the prompt text. Production cache hit rate for routing is 25.2% against 99.3% for the answering call on the same model.

To be clear about what this second part buys: it is a cost and PTU-capacity fix, not a latency one. Cached and uncached routing calls measure 729ms and 712ms, and going from under 1k to 4-6k input tokens only costs 64ms, so prefill was never the bottleneck.

Behaviour worth flagging for review:

`routing_latencies[].reasoning` is null for LLM hops unless the flag is set. Deterministic and router hops still carry their prefixed reason strings, so the `routing_type` classification is unchanged.

Routing instructions now render time variables frozen at call start rather than live, matching the spoken prompt. Expression edges still evaluate against live values, since they read `context_data` directly.

`_total_turns` and friends are no longer dumped into the routing prompt for the model to read. Instructions can still name one explicitly, and expression edges are unaffected.

### DIFF
--- a/bolna/agent_types/graph_agent.py
+++ b/bolna/agent_types/graph_agent.py
@@ -47,7 +47,7 @@ logger = configure_logger(__name__)
 _DETERMINISTIC_REASONING_PREFIX = "deterministic:"
 _ROUTER_REASONING_PREFIX = f"{_DETERMINISTIC_REASONING_PREFIX}router:"
 # The router's rationale is explanatory only, and its tokens sit on a live turn's critical path.
-_ROUTING_REASONING_ENABLED = os.getenv("GRAPH_ROUTING_REASONING", "").strip().lower() in ("1", "true", "yes")
+_ROUTER_RATIONALE_ENABLED = os.getenv("GRAPH_ROUTER_RATIONALE", "").strip().lower() in ("1", "true", "yes")
 _PROMPT_VAR_PATTERN = re.compile(r"\{([a-zA-Z_][a-zA-Z0-9_]*)\}")
 
 # Time variables frozen per call for the conversation prompt; see _prompt_context.
@@ -405,7 +405,7 @@ class GraphAgent(BaseAgent):
                     }
                     parameters["required"].append(param_name)
 
-            if _ROUTING_REASONING_ENABLED:
+            if _ROUTER_RATIONALE_ENABLED:
                 parameters["properties"]["reasoning"] = {
                     "type": "string",
                     "description": "Brief explanation of why this routing decision was made",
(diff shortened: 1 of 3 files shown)


bolna/agent_types/graph_agent.py
```diff
@@ -428,6 +431,16 @@ def _build_transition_tools_for_edges(self, edges: list, allow_stay: bool = True
...
         if allow_stay:
+            stay_properties = {}
+            if _ROUTER_RATIONALE_ENABLED:
+                stay_properties["reasoning"] = {
+                    "type": "string",
+                    "description": "Brief explanation of why this routing decision was made",
+                }
+            stay_properties["confidence"] = {
+                "type": "number",
+                "description": "Confidence score from 0.0 to 1.0 for this routing decision",
```
SubhrajitPrusty: small nit: I see these descriptions being repeated, can we move them to a const? Ideally that will give us a library of the error strings / tooltips in one place and updates to them will be much easier.

SubhrajitPrusty (approved):
small nit, rest LGTM

PR #979, opened 2026-08-24
fix: barge-in leaves an orphaned function_call in the Responses-API chain — calls die mid-tool-call

puneet-bolna:
## What

A web call died right after its tool call SUCCEEDED: the agent spoke the filler, the
API returned the data, and instead of answering, the agent hung up (~25s,
`hangup_by=Agent`). Example prod run: `daceca18-6647-4065-b50f-a958c42c6b90`
(GoodScore, 21 Aug 18:11 IST).

## Root cause — reproduced live against wss://api.openai.com/v1/responses

The agent's prompt makes the LLM call its context tool on the caller's FIRST turn, so
the turn-1 generation was mid-function-call when the caller barged in. We cancelled the
generation, but `cancel_in_flight_response` deliberately kept `previous_response_id` —
and the cancel can lose the race with generation: the server commits a `function_call`
item that this client never consumed, so no tool ever runs for it and no output is ever
recorded.

Every later request chained onto a lineage carrying that orphaned, unanswerable
function call. When turn-2's real tool finished and we sent its output back, OpenAI's
validator counted one output against TWO unanswered calls and rejected the request in
21ms:

    invalid_request_error: "No tool output found for function call call_7aYe…" (param: input)

The `APIError` escaped `task_manager.run` → full pipeline teardown → fork closed → FS
hung up the call. A recoverable validation error killed a live call whose tool had
already succeeded.

Ruled out along the way: the follow-up request itself was well-formed (the execution's
request-log CSV shows the `function_call_output` present), and the deployed 0.10.200
wheel is byte-identical to master on this path (not a version drift). Live replays
reproduce the byte-identical error whenever an unanswered function_call sits anywhere
in the chained lineage, and confirm the canonical chain works when the lineage is clean.

## Fix

`bolna/llms/openai_llm.py` + `bolna/llms/openai_base.py` (+178 / −10 incl. tests):

1. **A cancel drops the response chain.** After `response.cancel`, the server-side
   state is unknowable — the fc may or may not be committed. `cancel_in_flight_response`
   now clears `previous_response_id` + `_pending_call_ids` after sending the cancel
   frame (which still targets the old id). Cost: one full-history request per barge-in;
   the chain re-establishes on that response. Deliberately NOT
   `invalidate_response_chain()` — that would also wipe the interruption hint, which
   `sync_history` sets immediately before cancelling.

2. **The interruption hint rides the chainless path.** `_build_responses_input` only
   prepended the "*the user only heard: …*" hint on the chain-alive path — but after
   (1), the chainless full-history request IS the normal post-barge-in request. The
   hint is now prepended there too, so what the LLM learns about the interruption is
   unchanged.

3. **Chained rejections retry with full history instead of killing the call.**
   `_generate_stream_ws_responses` already recovers from `previous_response_not_found`
   by dropping the chain and retrying; that same recovery now covers any
   `invalid_request_error` on a CHAINED request. Full history carries every
   `function_call`/`function_call_output` pair inline, so it is self-consistent by
   construction. Gated on nothing-yielded (no text, no fc args, no pre-call message),
   so a retry can never duplicate speech; an UNCHAINED invalid request still raises —
   no retry loop.

Also fixed in passing: the pre-existing `previous_response_not_found` retry dropped
`tools` on recursion, silently disabling function calling for the retried turn.

## Why this is safe

- Nothing the caller hears changes: the chain is a token-saving delta mechanism, and
  full-history requests already run in prod today (WS 60-min reconnects, `not_found`
  retries, HTTP SSE fallback, the pending-tool-output guard). `store=True` prompt
  caching absorbs most of the latency of the one extra full upload per barge-in.
- Calls without interruptions are completely unaffected.
- Two tests that pinned the old behavior are inverted deliberately, with comments
  explaining why (`cancel_in_flight_does_not_invalidate_chain` →
  `cancel_in_flight_drops_chain_but_keeps_hint`; `hint_not_injected_without_chain` →
  `hint_injected_without_chain`).

## Tests

- New `TestChainedRequestRejectionRetry` (3): a replay of this incident's event
  sequence asserting the retry goes out unchained with the fc/output pair inline, the
  answer streams, and a clean chain re-establishes; unchained errors still raise after
  exactly one attempt; the retry carries the turn's `tools`.
- Updated hint/cancel pins (2): cancel sends the frame for the old id, drops the chain,
  keeps the hint; the hint is injected on the chainless path.
- Full suite: **1150 passed, 1 skipped** — no regressions. `ruff check` +
  `ruff format --check` clean.

### DIFF
--- a/bolna/llms/openai_base.py
+++ b/bolna/llms/openai_base.py
@@ -289,7 +289,8 @@ class OpenAICompatibleLLM(BaseLLM):
         With previous_response_id set, only sends items after the last
         assistant. Falls back to full history when pending tool outputs are
         missing. Any pending interruption hint is consumed exactly once and
-        only prepended on the chain-alive happy path.
+        prepended on both the chained path and the chainless full-history path
+        (the latter is the normal post-barge-in path — cancel drops the chain).
         """
         hint = self._interruption_hint
         self._interruption_hint = None
@@ -305,7 +306,12 @@ class OpenAICompatibleLLM(BaseLLM):
             if hint is not None:
                 input_items = [self._build_interruption_hint_item(hint), *input_items]
             return instructions, input_items
-        return MessageFormatAdapter.chat_to_responses_input(messages)
+        # No chain — the post-barge-in path since cancel_in_flight_response drops it. The hint
+        # matters most exactly here, so it rides the full-history request too.
+        instructions, input_items = MessageFormatAdapter.chat_to_responses_input(messages)
+        if hint is not None:
+            input_items = [self._build_interruption_hint_item(hint), *input_items]
+        return instructions, input_items
 
     @staticmethod
     def _build_interruption_hint_item(heard_text: str) -> dict:
--- a/bolna/llms/openai_base.py
+++ b/bolna/llms/openai_base.py
@@ -289,24 +289,22 @@ class OpenAICompatibleLLM(BaseLLM):
         With previous_response_id set, only sends items after the last
         assistant. Falls back to full history when pending tool outputs are
         missing. Any pending interruption hint is consumed exactly once and
-        prepended on both the chained and chainless paths.
+        prepended on every path.
         """
         hint = self._interruption_hint
         self._interruption_hint = None
 
-        if self.previous_response_id:
-            if self._pending_call_ids:
-                completed = {m.get("tool_call_id") for m in messages if m.get("role") == ChatRole.TOOL}
-                if not self._pending_call_ids.issubset(completed):
-                    logger.info("Pending tool call outputs missing, sending full context")
-                    self.previous_response_id = None
-                    return MessageFormatAdapter.chat_to_responses_input(messages)
+        chained = bool(self.previous_response_id)
+        if chained and self._pending_call_ids:
+            completed = {m.get("tool_call_id") for m in messages if m.get("role") == ChatRole.TOOL}
+            if not self._pending_call_ids.issubset(completed):
+                logger.info("Pending tool call outputs missing, sending full context")
+                self.previous_response_id = None
+                chained = False
+        if chained:
             instructions, input_items = self._extract_new_input(messages)
-            if hint is not None:
-                input_items = [self._build_interruption_hint_item(hint), *input_items]
-            return instructions, input_items
-        # no chain = the post-barge-in path (cancel drops it) — the hint rides it too
-        instructions, input_items = MessageFormatAdapter.chat_to_responses_input(messages)
+        else:
+            instructions, input_items = MessageFormatAdapter.chat_to_responses_input(messages)
         if hint is not None:
             input_items = [self._build_interruption_hint_item(hint), *input_items]
         return instructions, input_items
--- a/bolna/llms/openai_base.py
+++ b/bolna/llms/openai_base.py
@@ -289,8 +289,7 @@ class OpenAICompatibleLLM(BaseLLM):
         With previous_response_id set, only sends items after the last
         assistant. Falls back to full history when pending tool outputs are
         missing. Any pending interruption hint is consumed exactly once and
-        prepended on both the chained path and the chainless full-history path
-        (the latter is the normal post-barge-in path — cancel drops the chain).
+        prepended on both the chained and chainless paths.
         """
         hint = self._interruption_hint
         self._interruption_hint = None
@@ -306,8 +305,7 @@ class OpenAICompatibleLLM(BaseLLM):
             if hint is not None:
                 input_items = [self._build_interruption_hint_item(hint), *input_items]
             return instructions, input_items
-        # No chain — the post-barge-in path since cancel_in_flight_response drops it. The hint
-        # matters most exactly here, so it rides the full-history request too.
+        # no chain = the post-barge-in path (cancel drops it) — the hint rides it too
         instructions, input_items = MessageFormatAdapter.chat_to_responses_input(messages)
         if hint is not None:
             input_items = [self._build_interruption_hint_item(hint), *input_items]
--- a/bolna/llms/openai_llm.py
+++ b/bolna/llms/openai_llm.py
@@ -554,12 +554,7 @@ class OpenAiLLM(OpenAICompatibleLLM):
                         ):
                             yield chunk
                         return
-                    # Any other rejection of a CHAINED request gets one full-history retry: the
-                    # server-side lineage can be unsatisfiable in ways the client can't see (e.g.
-                    # a cancelled response that committed a function_call nobody will ever answer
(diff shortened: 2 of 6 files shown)


bolna/llms/openai_base.py
```diff
@@ -289,7 +289,7 @@ def _build_responses_input(self, messages):
         With previous_response_id set, only sends items after the last
         assistant. Falls back to full history when pending tool outputs are
         missing. Any pending interruption hint is consumed exactly once and
-        only prepended on the chain-alive happy path.
+        prepended on both the chained and chainless paths.
```
san0808: This updated docstring says the hint is prepended on both the chained and chainless paths, but the pending-tool-output fallback a few lines below returns MessageFormatAdapter.chat_to_responses_input(messages) right after setting previous_response_id = None, without prepending the hint. That path is chainless yet drops a pending hint. It is not reachable in today's barge-in flow (cancel clears the chain before sync_history sets the hint, so previous_response_id is None on the next build and this branch is skipped), so this is a latent inconsistency rather than a live bug. Either prepend the hint there too, or scope the docstring to exclude that fallback.

bolna/llms/openai_llm.py
```diff
@@ -550,7 +550,24 @@ async def _generate_stream_ws_responses(
...
+                    # a rejected CHAINED request gets one full-history retry (nothing yielded yet)
+                    if (
+                        self.previous_response_id
+                        and error_info.get("type") == "invalid_request_error"
+                        and not answer
+                        and not func_call_args
+                        and not gave_pre_call_msg
+                    ):
+                        logger.warning(
+                            f"WS chained request rejected ({error_info.get('message')}), retrying with full history"
```
san0808, line 577: This recovery is not recorded anywhere, unlike the sibling incomplete-empty retry below it which appends to meta_info['_non_fatal_errors']. Since this is the exact poisoned-chain incident the PR fixes, adding a _non_fatal_errors entry here would let prod analytics and Datadog see how often the chained-rejection recovery actually fires.

bolna/llms/openai_llm.py
```diff
@@ -550,7 +550,24 @@ async def _generate_stream_ws_responses(
                         logger.warning(f"WS previous_response_id not found, retrying with full history")
                         self.previous_response_id = None
                         async for chunk in self._generate_stream_ws_responses(
-                            messages, synthesize, request_json, meta_info, tool_choice
+                            messages, synthesize, request_json, meta_info, tool_choice, tools
+                        ):
+                            yield chunk
+                        return
+                    # a rejected CHAINED request gets one full-history retry (nothing yielded yet)
```
san0808: This block is a near-verbatim copy of the previous_response_not_found retry directly above it (drop chain, recurse with the same args including tools, yield, return). A single _retry_full_history helper would keep the recursion in one place, so future changes such as the telemetry entry noted above or a retry-depth guard do not have to be made twice.

bolna/llms/openai_llm.py
```diff
@@ -550,7 +550,24 @@ async def _generate_stream_ws_responses(
...
+                    if (
+                        self.previous_response_id
+                        and error_info.get("type") == "invalid_request_error"
```
san0808, line 569: This retry catches any invalid_request_error on a chained request, including non-recoverable ones like context_length_exceeded (OpenAI reports it as invalid_request_error) or content-policy rejections. For those, the retry re-sends the full history (a strictly larger payload in the context-length case), fails the same way, and only then raises on the unchained attempt. The net cost on a live call is one wasted round-trip of dead air before the turn errors out. Consider narrowing to the tool-output/lineage errors this fix targets, or noting that the extra round-trip is an accepted cost.

bolna/llms/openai_llm.py
```diff
@@ -732,9 +749,13 @@ async def _generate_stream_ws_responses(
         self.started_streaming = False
 
     def cancel_in_flight_response(self):
-        """Cancel the in-flight WS response; keeps previous_response_id alive."""
+        """Cancel the in-flight WS response and drop the chain — a lost cancel race can leave a
+        server-committed function_call this client never saw, poisoning every later chained
+        request. Chain state only: the interruption hint was just set by sync_history."""
```
san0808: This rationale is inverted relative to the actual call order. In task_manager.__cleanup_downstream_tasks, _cancel_in_flight_llm_response() runs first and sync_history() (which sets the hint) runs afterward, and that is the only cancel call site. So the hint is not yet set when this runs. The behavior is still correct (cancel does not touch the hint, and it is set right after and consumed on the next generate), but the comment describes an ordering that does not hold and will mislead a future maintainer.

bolna/llms/openai_base.py
puneet-bolna: I made the behavior uniform: _build_responses_input now resolves which item set to send first and prepends the hint at a single exit, so every path — chained, chainless, and the pending-tool fallback — carries it. You're also right that the fallback is unreachable with a hint in today's flow (cancel drops the chain before sync_history sets it), so this is future-proofing, not a behavior change. Test renamed to test_hint_injected_on_pending_tool_fallback to pin it.

bolna/llms/openai_llm.py
puneet-bolna: Added — the retry now appends {"error_type": "chained_request_rejected", "error": <message>, "model": ...} to meta_info["_non_fatal_errors"], same shape as the incomplete-empty retry below it. The incident replay test asserts it. This lands in the usage pipeline, so we can count real-world fires.

bolna/llms/openai_llm.py
puneet-bolna: Done — both retries now delegate to a _retry_full_history() async-gen helper (drop chain → re-run on full history). The telemetry from #2 and any future retry-depth guard have one home. As a side effect this is also where the tools-dropped-on-recursion bug from the earlier round stays fixed for both branches.

bolna/llms/openai_llm.py
puneet-bolna: context_length_exceeded retried on a strictly larger payload is the worst case. Narrowed to code falsy and param == "input", which is exactly the lineage/pairing rejection shape from the incident (code None, param 'input'); anything with a code — context length, content policy — raises immediately as before. Added test_coded_invalid_request_does_not_retry pinning that a chained context_length_exceeded gets exactly one attempt. The residual accepted cost: an uncoded param-input rejection that a retry can't fix costs one extra round-trip before raising — that's the price of catching the poisoned-lineage class, which is also uncoded.

bolna/llms/openai_llm.py
puneet-bolna: fixed

san0808 (approved):
lgtm

PR #980, opened 2026-08-24
Lid explicit requests

puneet-bolna:
One language note instead of two
Problem

Multilingual agents were running two different language instructions at different points in the same call. At setup we install __language_directive (the long standing order). On the first LID switch, that gets replaced by __switch_context_note — a shorter, switch-specific note. Two prod calls show what that costs.

772a996b — agent switched te→en, and the entire reply was:

You said, "Yeah, tell me."

Then it switched back en→te and said "You said you want to speak in Telugu, OK I'll continue in Telugu" — and repeated that same meta line on the next turn. Two turns burned, no content.

bb8e8fb3 — same echo, and it persisted for the rest of the call. Pulling the actual request payload for turn 4 shows why. Forty seconds and three turns after the switch, the system prompt still read:

"…In your NEXT reply only, first briefly restate your previous line in English… Their latest message: "Tell me". Reason: Caller speaking English substantively. LIVE confirms. Switch from Telugu."

Root cause

The switch note is one-shot text living in a permanent slot:

"In your NEXT reply only" — nothing ever removes the note, so every turn is "the next reply". The echo is a standing instruction, not the model drifting.
Their latest message: "Tell me" — turn 1's text, still presented as the caller's latest on turn 4, when they had just described a driver accident in Hindi.
Reason: … — the judge's internal reasoning, left in the prompt, which is what gets the agent talking about languages instead of just switching.

And because it replaces the standing directive, a switch also silently drops that directive's rules: use the target-language version of every question/FAQ/objection/closing line, translate when a variant is missing, and never translate proper nouns, brand names, alphanumeric identifiers, digits or codes — lost at exactly the moment language handling matters most.

Change

Delete __switch_context_note. Every path now installs __language_directive:

both LID switch sites (including the in-flight-function-call path)
the legacy switch_language tool path (already fell through to it)
the speculative follow-up's prompt copy, so a committed speculation and a real switch are identical

Nothing is lost by dropping the caller's message from the note — it was a duplicate of what the switch already writes into conversation history as the user turn, which the follow-up generation reads. The agent still answers what the caller said; it just stops narrating it.

Tests

test_simple_agent_language_directive.py rewritten: the third test now asserts the installed text carries no switch-moment wording (NEXT reply, latest message, Reason:), so re-introducing this fails CI. Four other test doubles repointed at the surviving builder.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.207"
+__version__ = "0.10.208"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.204"
+__version__ = "0.10.205"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -6009,7 +6009,7 @@ class TaskManager(BaseManager):
             self.lid_playback_gate = None
             logger.info("LanguageSwitcher: in-flight function call — switching in parallel, not truncating the action")
             if target != self.language:
-                context_note = self.__switch_context_note(target, detector_transcript, reasoning)
+                context_note = self.__language_directive(target)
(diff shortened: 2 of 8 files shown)


prabalrastogi (approved):
LGTM

PR #983, opened 2026-08-25
feat: explicit-only language-switch mode — per-agent toggle where the judge switches only on an explicit caller request

puneet-bolna:
Description:

What

A new opt-in mode for the LLM language switch. When an agent has tools_config["language_switch_explicit_only"] set (per-agent toggle from the dashboard's multilingual config), the switch judge runs a different prompt whose contract is: switch only when the caller explicitly requests, selects, or confirms a language — the language they happen to be speaking never triggers a switch.

Default is off. With the toggle absent, nothing changes — the ambient judge, its prompt, and every gate run exactly as today.

Why

Some customers ask the caller for a language preference in their own flow (or operate in high code-mixing environments) and want deterministic switching: only on "Hindi please" / "can you speak English?" / a "yes" to the agent's own language question — never because the caller uttered a sentence in another language.

How

New prompt pair (bolna/prompts.py): EXPLICIT_LANGUAGE_SWITCH_SYSTEM_PROMPT — an 18-rule controller prompt covering direct requests, capability questions, agent-prompted selection (one-word answers, ordinals, yes/no confirmations), corrections, rejections, mentions/quotes/hypotheticals, unsupported and already-active requests — plus EXPLICIT_LANGUAGE_SWITCH_TURN_PROMPT. The system block keeps the same cache_control treatment as the ambient prompt.

One new input: last_agent_turn. The explicit prompt interprets the caller's reply in the context of the agent's immediately-preceding line ("Would you prefer Hindi or Telugu?" → "Telugu." switches). Sourced from a new ConversationHistory.last_assistant_content() (skips tool rows). The explicit mode reads this instead of the ambient prompt's RECENT TURNS drift history.

Gate semantics in explicit mode: the judge's authorization is the gate. The detection-confidence gates (confidence/corroboration, substance/short-audio, rule-3a alphanumeric veto) are bypassed — they exist to filter noisy ambient detections and would veto legitimate one-word answers like "Telugu." The structural gates are kept: hangup/teardown re-check, target-in-labels, target-has-a-synth-voice, concurrent-switch — those are physical-capability checks, not detection checks.

Telemetry: lid_detection_events records now carry the explicit judge's request_status (switch|already_active|unsupported|ambiguous|no_request) and request_source (direct_request|specific_capability_question|agent_prompted_selection|none); None on ambient decisions.

Not changed

The ambient path is untouched code, not a re-tested equivalent: all 94 pre-existing LID tests pass without a single assertion edited. Speculation, playback gate, idle-flush watcher, switch mechanics, and the language directive are shared unchanged by both modes.

Tests

tests/test_language_switch_explicit.py (14):

prompt selection per mode; last_agent_turn present in explicit / absent in ambient; (none) marker; empty-LIVE stays "" in explicit (ambient keeps its idle-flush marker)
the three canonical examples end-to-end through decide(): speaks-Hindi-no-request → stay; one-word "Telugu." after the agent's question → switch; "yes" to two offered languages → ambiguous/stay
task_manager: explicit switch proceeds past evidence that gates ambient mode (0.3s low-prob segment, self-confidence 0.2); same evidence in ambient mode → gated:low_confidence; unsupported target and missing synth voice still block in explicit mode; telemetry fields recorded

Test-harness hardening: five existing suites build TaskManager stand-ins from MagicMock, whose auto-created lid_explicit_only is truthy — they'd have silently taken the new branch. Set False in the shared conftest factory + 3 files.

Full suite: 1186 passed, 1 skipped. ruff check + format clean.

Config contract

tools_config["language_switch_explicit_only"]: bool — set per call by dashboard-backend from multilingual_config.explicit_switch_only (companion PR). Only read when llm_language_switch is enabled.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -618,7 +618,6 @@ class TaskManager(BaseManager):
         # Dedicated LLM that decides language switches from the unbiased detector
         # transcript. Set up in __setup_transcriber only for gated multilingual agents.
         self.language_switcher = None
-        self.lid_explicit_only = False
         # Serializes switch decisions so overlapping turns can't interleave two
         # switch+follow-up sequences. Background-only: the caller-facing pipeline
         # (ASR -> main LLM -> TTS) never waits on this lock.
@@ -1622,16 +1621,15 @@ class TaskManager(BaseManager):
                         lid_config["sarvam_model"] = lid_model
                     switch_enabled = self.__language_switch_enabled()
                     if switch_enabled:
-                        # Per-agent toggle: judge switches only on an explicit request/selection.
-                        self.lid_explicit_only = bool(
-                            self.task_config.get("tools_config", {}).get("language_switch_explicit_only")
-                        )
                         # language_switch_llm (from the azure flag) overrides the model; absent → default.
                         self.language_switcher = LanguageSwitcher(
                             available_labels=list(transcribers.keys()),
                             run_id=self.run_id,
                             model=self.task_config.get("tools_config", {}).get("language_switch_llm"),
-                            explicit_only=self.lid_explicit_only,
+                            # Per-agent toggle: judge switches only on an explicit request/selection.
+                            explicit_only=bool(
+                                self.task_config.get("tools_config", {}).get("language_switch_explicit_only")
+                            ),
                         )
                         self.language_switcher.prewarm()  # pay the TLS handshake now
 
@@ -5568,13 +5566,15 @@ class TaskManager(BaseManager):
         if events is None or any(e.get("type") == "lid_usage" for e in events):
             return
         switcher = self.language_switcher
-        detector_seconds = pool.lid_audio_seconds() if hasattr(pool, "lid_audio_seconds") else None
+        detector_seconds = pool.lid_audio_seconds()
         if switcher is None and not detector_seconds:
             return
         record = {"type": "lid_usage", "ts": time.time(), "detector_audio_seconds": detector_seconds}
         if switcher is not None:
             record.update(
                 {
+                    # Every model that answered this call — the runtime fallback can swap mid-call.
+                    "judge_models": switcher.models_used or [switcher.model],
                     "judge_model": switcher.model,
                     "judge_requests": switcher.usage_totals.get("requests", 0),
                     "judge_input_tokens": switcher.usage_totals.get("input_tokens", 0),
@@ -5867,6 +5867,7 @@ class TaskManager(BaseManager):
         spec_agent_type = self.task_config["tools_config"]["llm_agent"].get("agent_type", "simple_llm_agent")
         if (
             spec_agent_type == "simple_llm_agent"
+            and not self.language_switcher.explicit_only
             and detected_lang
             and detected_lang != active
             and detected_lang in labels
@@ -5930,7 +5931,7 @@ class TaskManager(BaseManager):
                     detector_transcript,
                     active_transcript,
                     active,
-                    recent_turns=self.__recent_detected_turns(pool),
+                    recent_turns=None if self.language_switcher.explicit_only else self.__recent_detected_turns(pool),
                     last_agent_turn=self.conversation_history.last_assistant_content(),
                 ),
                 timeout=decide_timeout_s,
@@ -6008,14 +6009,20 @@ class TaskManager(BaseManager):
                     None,
                 )
             )
-        # Explicit-only mode: the judge's own authorization IS the gate — its prompt only
-        # ever targets an explicit request, so the ambient detection gates below would veto
-        # legitimate requests (a one-word "Telugu." answer fails every one of them).
-        if self.lid_explicit_only:
+        # Explicit-only mode: a switch is authorized iff the judge returned a consistent
+        # explicit verdict (request_status="switch" AND explicit_request) — the ambient
+        # detection gates below grade evidence the explicit contract does not produce.
+        if self.language_switcher.explicit_only:
+            if decision.get("request_status") != "switch" or not decision.get("explicit_request"):
+                logger.info(
+                    f"LanguageSwitcher: explicit-only mode — target '{target}' without an explicit verdict "
+                    f"(status={decision.get('request_status')}, explicit={decision.get('explicit_request')}) — no switch"
+                )
+                emit_lid_decision("gated:not_explicit")
+                return
             logger.info(
                 f"LanguageSwitcher: explicit-only mode — switch to '{target}' authorized "
-                f"(status={decision.get('request_status')}, source={decision.get('request_source')}, "
-                f"conf={target_conf}); detection gates bypassed"
+                f"(status=switch, source={decision.get('request_source')}, conf={target_conf})"
             )
         else:
             # Corroboration: when the detector independently agrees on the target, accept a lower LLM
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -210,6 +210,9 @@ def build_lid_decision_record(
         "target_language": dec.get("target_language"),
         "target_confidence": dec.get("target_confidence"),
         "explicit_request": dec.get("explicit_request"),
+        # Explicit-only judge fields; None on the ambient prompt.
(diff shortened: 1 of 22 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -5959,53 +5995,63 @@ def as_float(value):
...
-        # them), so a one-token "okay" could lend its 1.0 token-share purity to a whole Hindi turn.
-        corroborated = self.__detector_corroborates(detector_segments, target)
-        effective_min_conf = min_conf
-        if corroborated:
-            effective_min_conf = float(os.getenv("LANGUAGE_SWITCH_CORROBORATED_MIN_CONFIDENCE", "0.55"))
-        if target_conf is None or target_conf < effective_min_conf:
+        # Explicit-only mode: the judge's own authorization IS the gate — its prompt only
+        # ever targets an explicit request, so the ambient detection gates below would veto
+        # legitimate requests (a one-word "Telugu." answer fails every one of them).
+        if self.lid_explicit_only:
```
san0808: Blocking: explicit mode needs to check the judge's verdict, not just that a target came back.

Verified on this branch: a decision of `{"target_language": "hi", "request_status": "ambiguous", "explicit_request": false, "target_confidence": 0.0}` switches, and logs itself as `authorized (status=ambiguous, source=none, conf=0.0)`. Rule 18 of the prompt is the only thing that prevents that shape, and this is the one mode where every drift backstop is off.

```python
if self.lid_explicit_only:
    if decision.get("request_status") != "switch" or not decision.get("explicit_request"):
        emit_lid_decision("gated:not_explicit")
        return
```

It also gives contract violations their own outcome instead of recording them as clean switches.

bolna/agent_manager/task_manager.py
```diff
@@ -1618,11 +1622,16 @@ def __setup_transcriber(self):
                         lid_config["sarvam_model"] = lid_model
                     switch_enabled = self.__language_switch_enabled()
                     if switch_enabled:
+                        # Per-agent toggle: judge switches only on an explicit request/selection.
+                        self.lid_explicit_only = bool(
```
san0808: Speculation and the playback gate fire on the wrong turns in this mode.

`__detector_language_mismatch` and the `spec_task` block in `__run_language_switch` both key off `detected_lang != active`. On the flows this feature exists for the caller answers in the active language: the prompt's own canonical examples give `detected_language: "en"` for `"Telugu."` answering an English question, and for a `"Yes."` confirmation. So on the turns that actually switch neither the gate nor the speculation fires, and the English reply starts and then gets truncated mid sentence. On foreign turns with no request both fire and the result is always discarded.

`and not self.lid_explicit_only` on the spec condition is the cheap half. The truncation on selection turns is not something the gate can fix, so it is worth hearing on a test call before this reaches a customer.

bolna/agent_manager/task_manager.py
```diff
@@ -5537,8 +5546,31 @@ def __snapshot_lid_events(self) -> list:
                 record()
             except Exception as e:
                 logger.warning(f"detector_health record failed: {e}")
+        self.__record_lid_usage(pool)
         return list(getattr(pool, "lid_detection_events", []))
 
+    def __record_lid_usage(self, pool) -> None:
```
san0808, line 5563: This, plus `bytes_fed` / `audio_seconds_fed`, `usage_totals`, the `ret_metadata` change and `test_lid_usage_tracking.py`, is a separate feature from explicit-only switching and is not in the PR description.

Blast radius is contained (`LanguageSwitcher` is the only thing that constructs a `LiteLLM`, so the aux hangup and voicemail paths on `OpenAiLLM` are untouched), so this is a scope call rather than a risk one. It still belongs in its own PR.

bolna/helpers/language_switcher.py
```diff
@@ -112,25 +122,38 @@ def _system_message(self):
...
+        system_text = EXPLICIT_LANGUAGE_SWITCH_SYSTEM_PROMPT if self.explicit_only else LANGUAGE_SWITCH_SYSTEM_PROMPT
+        block = {"type": "text", "text": system_text}
         cacheable = self.model.startswith(("anthropic/", "claude")) or (
             self.model.startswith("bedrock/") and "claude" in self.model
         )
         if cacheable:
             block["cache_control"] = {"type": "ephemeral"}
         return {"role": "system", "content": [block]}
 
+    def _tally_usage(self, usage):
```
san0808, line 135: Three gaps in the spend numbers this feeds:

- `_note_failure` can swap `self.model` mid call, so `judge_model` names the fallback while the tokens are summed across both models.
- Under hedging the losing attempt is cancelled in the `finally`. If it dies before reaching here, tokens the provider already billed go uncounted.
- The early return on empty usage means `judge_requests` counts usage-bearing responses, not requests. `test_missing_usage_logs_none_not_crash` asserts `requests == 0` for a decide that did happen. Either count in `attempt()` or rename the field.

bolna/agent_manager/task_manager.py
```diff
@@ -615,6 +618,7 @@ def __init__(
         # Dedicated LLM that decides language switches from the unbiased detector
         # transcript. Set up in __setup_transcriber only for gated multilingual agents.
         self.language_switcher = None
+        self.lid_explicit_only = False
```
san0808: `self.lid_explicit_only` and `self.language_switcher.explicit_only` are the same bit stored twice. The gate in `__run_language_switch` only runs when a switcher exists, so reading `self.language_switcher.explicit_only` there drops this attribute.

bolna/lid/sarvam.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, on_language, config):
             self._encoding = "linear16"
             self._input_sr = self._sr
         self._resample_state = None
+        self.input_bytes_per_second = 8000 if self._encoding == "mulaw" else 2 * self._input_sr
```
san0808: The same expression is copy-pasted into `soniox.py`, and both hardcode 8000 where `_input_sr` is already 8000 on the line above (mulaw is one byte per sample). A backend that forgets this line reports 0.0 seconds silently rather than failing. Worth computing in `LIDBackend` from the encoding and rate the subclass already sets.

bolna/agent_manager/task_manager.py
```diff
@@ -5537,8 +5546,31 @@ def __snapshot_lid_events(self) -> list:
...
+        self.__record_lid_usage(pool)
         return list(getattr(pool, "lid_detection_events", []))
 
+    def __record_lid_usage(self, pool) -> None:
+        """One per-call spend record in lid_detection_events: judge tokens + detector audio seconds."""
+        events = getattr(pool, "lid_detection_events", None)
+        if events is None or any(e.get("type") == "lid_usage" for e in events):
+            return
+        switcher = self.language_switcher
+        detector_seconds = pool.lid_audio_seconds() if hasattr(pool, "lid_audio_seconds") else None
```
san0808: Minor, collected in one place:

- The `hasattr` guard is dead here. The function has already returned unless `pool` is a real `TranscriberPool`.
- `recent_turns=self.__recent_detected_turns(pool)` is still computed and passed to `decide()` in explicit mode, where it is ignored.
- The explicit-only comment in `__run_language_switch` argues from a specific case ("a one-word 'Telugu.' answer fails every one of them") where stating the rule would do.

bolna/lid/sarvam.py
puneet-bolna: done

bolna/agent_manager/task_manager.py

bolna/helpers/language_switcher.py
puneet-bolna: done

bolna/agent_manager/task_manager.py

san0808: Speculation half confirmed. The other half is still open: `__detector_language_mismatch` and `__arm_lid_playback_gate` are unchanged, so a foreign turn with no request still holds the reply audio for a decide that in this mode always stays. The gate self-releases once generation lands, so the cost is small, and the truncation on a selection turn is not something the gate can fix either way. Leaving this open until it has been heard on a test call.

san0808 (approved):

PR #984, opened 2026-08-26
feat: real playback marks for FreeSWITCH webcalls — in-band echoes replace the duration estimator

puneet-bolna:
## What

Webcall (FreeSWITCH) playback tracking is simulated today: every TTS chunk's mark is
acked in ONE batch when a duration estimate (`first_send + audio_bytes/rate`) elapses.
Telephony gets real per-chunk acks (Twilio/Plivo echo marks at the playhead; sip-trunk
mirrors it via Asterisk's MEDIA_MARK_PROCESSED). Prod evidence of the simulation
(run `daceca18…`): 19 marks sent over ~130ms, all 19 "acked" in the same 10ms window
2.9s later — `avg_delay 2.921s` is an artifact, not latency.

What the simulation costs:
- **Barge-in truncation is blind mid-response** — `sync_history` reconstructs "what the
  caller heard" from per-mark ack timestamps; a mid-response barge-in has zero acked
  marks, so the truncated transcript (and the interruption hint fed to the LLM) is wrong.
- **Turn-end drifts** — the estimator assumes playout starts at first send; FS injection
  timing + the browser jitter buffer + slow TTS all push real playback later, so the
  "agent is speaking" window closes early and real barge-ins aren't classified.
- **Tool/hangup gating rides the same clock** (`wait_for_current_message`).
- mark_tracking stats are noise on webcalls — nothing alertable.

## How

Twilio's documented mark semantics, adapted to the fork protocol (paired
fs-media-server PR adds the module side):

- **Output handler** sends `{"type":"mark","name":id}` in-band after each chunk's
  streamAudio frames (mark registered before the send, so an echo can never race it).
  The patched module echoes `{"type":"markPlayed"}` when its playhead crosses that
  offset — sample-accurate, per chunk.
- **Input handler** routes echoes to the existing generic `process_mark_message`
  (real per-chunk ack timestamps for sync_history/latency) and notifies the output
  handler. `"cleared":true` echoes (killAudio flush / ring overflow = dropped unplayed)
  ack the registry but NEVER drive turn-end.
- **Turn-end** = the final mark's echo + `WEBCALL_PLAYBACK_SETTLE_S` (env, default
  0.15s) covering the browser's jitter buffer — the exact reason the module's
  `playoutDone` was left unwired. Completion reuses `_complete_after_playout`, so it
  stays cancellable and idempotent.
- **Estimator demoted to fallback, not removed**: with an unpatched module no echoes
  arrive and behavior is byte-identical to today (verified: the old module's
  processMessage ignores unknown types, and registry acks pop-on-fetch so the second
  ack no-ops). No version negotiation needed; either side deploys first.

Stale-echo hardening (both found in self-review, regression-tested): `_final_mark_id`
is cleared in EVERY completion path, so a late final echo (estimator ran early, module
still draining) can't match into the next response and clear `is_audio_being_played`
mid-speech.

## Not touched

Telephony handlers, sip_trunk, `mark_event_meta_data` API, task_manager, `playoutDone`
handling, dashboard-backend (handlers are constructed via bolna's provider registry;
deploying needs only the usual version bump there).

## Tests

New `tests/test_freeswitch_marks.py` (9): in-band send ordering; final-echo completes
the turn and cancels the estimator; no-echo falls back to the estimator; mid-response
echo does not complete; interruption resets; stale-final-echo after estimator
completion is ignored; cleared echo acks but never ends the turn; input routing;
callback wiring. Full suite: **1174 passed**, ruff check + format clean.

### DIFF
--- a/bolna/input_handlers/telephony_providers/freeswitch.py
+++ b/bolna/input_handlers/telephony_providers/freeswitch.py
@@ -33,11 +33,16 @@ class FreeSwitchInputHandler(DefaultInputHandler):
 
     async def process_message(self, message):
         if message.get("type") == "markPlayed":
-            # real per-chunk playback ack: registry ack here, turn-end in the output handler
             name = message.get("name")
+            # cleared = dropped unplayed (killAudio/ring overflow): pop the registry WITHOUT
+            # recording heard text or driving turn-end — the caller never heard this audio
+            if message.get("cleared"):
+                if self.mark_event_meta_data:
+                    self.mark_event_meta_data.fetch_data(name)
+                return
+            # real per-chunk playback ack: registry ack here, turn-end in the output handler
             self.process_mark_message({"type": "mark", "name": name})
-            # cleared = dropped unplayed (killAudio/ring overflow): ack registry, never drive turn-end
-            if not message.get("cleared") and self.on_mark_played:
+            if self.on_mark_played:
                 self.on_mark_played(name)
             return
         if message.get("type") == "playoutDone":
--- a/bolna/input_handlers/telephony_providers/freeswitch.py
+++ b/bolna/input_handlers/telephony_providers/freeswitch.py
@@ -34,11 +34,11 @@ class FreeSwitchInputHandler(DefaultInputHandler):
     async def process_message(self, message):
         if message.get("type") == "markPlayed":
             name = message.get("name")
-            # cleared = dropped unplayed (killAudio/ring overflow): pop the registry WITHOUT
-            # recording heard text or driving turn-end — the caller never heard this audio
+            # cleared = dropped unplayed (killAudio/ring overflow): drop from the registry with
+            # no ack stamp — heard text, tail-credit and turn-end must never count it
             if message.get("cleared"):
                 if self.mark_event_meta_data:
-                    self.mark_event_meta_data.fetch_data(name)
+                    self.mark_event_meta_data.drop_data(name)
                 return
             # real per-chunk playback ack: registry ack here, turn-end in the output handler
             self.process_mark_message({"type": "mark", "name": name})
--- a/bolna/input_handlers/telephony_providers/freeswitch.py
+++ b/bolna/input_handlers/telephony_providers/freeswitch.py
@@ -28,8 +28,22 @@ class FreeSwitchInputHandler(DefaultInputHandler):
         # set by FreeSwitchOutputHandler: called when mod_audio_stream reports its playout
         # buffer ACTUALLY drained (real playback-complete, vs the byte-count estimate)
         self.on_playout_done = None
+        # set by FreeSwitchOutputHandler: per-mark playback echoes from the patched module
+        self.on_mark_played = None
 
     async def process_message(self, message):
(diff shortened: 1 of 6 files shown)


bolna/output_handlers/telephony_providers/freeswitch.py
```diff
@@ -186,11 +214,12 @@ async def handle(self, packet):
...
             self._pending_marks.append(mark_id)
+            # in-band mark after the chunk's frames (ordered WS); an unpatched module ignores the type
+            await self.websocket.send_text(json.dumps({"type": "mark", "name": mark_id}))
 
-            # on the final chunk, arm completion: the module's playoutDone event finishes the
-            # response at the REAL drain moment; the estimated timer is the fallback if the
-            # event is lost (WS blip) — whichever fires first wins, both are idempotent.
+            # on the final chunk, arm completion: final-mark echo wins, estimated timer is the fallback
             if is_final:
+                self._final_mark_id = mark_id
```
san0808, line 230: The estimator preempts this echo path. `remaining = (first_send + total_dur) - now` has no margin, but the real echo can't arrive before `first_send + startup + total_dur`, so the estimator always fires first, clears `_final_mark_id`, and `+playback_settle_s` never applies. Turn-end stays at the bare estimate (same as master), so the tail-clip isn't fixed. Per-chunk acks for barge-in truncation do land. `test_final_mark_echo_completes_turn_and_supersedes_estimator` only passes because it fires the echo before the estimator elapses; prod inverts that.

Fix: when `marks_echoed`, give the estimator a margin >= settle or make it a long watchdog so the echo governs turn-end. Not a blocker, but the settle is the point.

puneet-bolna: done

bolna/input_handlers/telephony_providers/freeswitch.py
```diff
@@ -28,8 +28,23 @@ def __init__(self, *args, **kwargs):
...
+        if message.get("type") == "markPlayed":
+            name = message.get("name")
+            # cleared = dropped unplayed (killAudio/ring overflow): pop the registry WITHOUT
+            # recording heard text or driving turn-end — the caller never heard this audio
+            if message.get("cleared"):
+                if self.mark_event_meta_data:
+                    self.mark_event_meta_data.fetch_data(name)
+                return
+            # real per-chunk playback ack: registry ack here, turn-end in the output handler
+            self.process_mark_message({"type": "mark", "name": name})
```
san0808, line 44: `cleared:true` echoes still reach `process_mark_message`, which appends `text_synthesized` to `response_heard_by_user`. On killAudio this is harmless (`clear_data()` already emptied the dict), but on mark-ring overflow (>64 in-flight marks, no interruption) the dropped chunk's text gets recorded as heard, corrupting the sync_history transcript. Skip this call when `cleared` is set, or record it as dropped. Minor, not blocking.

puneet-bolna: done

san0808 (approved):

PR #986, opened 2026-08-26
Add Kalpa Labs TTS synthesizer

pshishodiaa:
Kalpa Labs TTS: streaming WSS /v1/tts/{voice_id}/stream, one-shot POST /v1/tts/{voice_id}.

Kalpa's conversational speech models render named voices from GET /v1/voices

Models:
1. kalpa-tts-multilingual-beta-v0.1 speaks English and Hindi, including code-switched Hinglish with no language parameter
2. kalpa-tts-beta-v0.1 is English-only.

Flow
1. The client authenticates with an initializeConnection frame (the handshake itself is unauthenticated),
2. The server confirms with sessionCreated
3. Each sendText frame appends text and flush=true renders the buffered utterance.
4. Audio arrives as responseAudio frames carrying base64 raw 24 kHz mono s16le PCM.
5. Every flushed utterance terminates with exactly one responseDone (status "completed", or "cancelled" after a cancelResponse).
6. For telephony, mu-law encodes audio from 24KHz to 8KHz

## Config

```json
{
  "provider": "kalpa",
  "provider_config": {
    "voice": "Kiara",
    "model": "kalpa-tts-multilingual-beta-v0.1"
  },
  "stream": true
}
```

Voices resolve by display name against GET /v1/voices ("Kiara" → catalog UUID, cached), or pass `voice_id` directly. Optional sampling controls: `temperature`, `top_k`, `acoustic_temperature`, `max_new_tokens`, `audio_quality` — omitted by default so Kalpa's tuned production sampling applies. Env: `KALPA_API_KEY` (or `synthesizer_key`), `KALPA_API_HOST` to override the host.

### DIFF
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -126,7 +126,6 @@ class KalpaConfig(BaseModel):
     voice_id: Optional[str] = None
     model: str = "kalpa-tts-multilingual-beta-v0.1"
     temperature: Optional[float] = None
-    top_k: Optional[int] = None
     acoustic_temperature: Optional[float] = None
     max_new_tokens: Optional[int] = None
     audio_quality: Optional[str] = None
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -129,6 +129,7 @@ class KalpaConfig(BaseModel):
     acoustic_temperature: Optional[float] = None
     max_new_tokens: Optional[int] = None
     audio_quality: Optional[str] = None
+    chunk_length_schedule: Optional[List[int]] = None
 
 
 class AzureConfig(BaseModel):
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -122,15 +122,9 @@ class MayaConfig(BaseModel):
 
 
 class KalpaConfig(BaseModel):
-    # Display name from GET /v1/voices (e.g. "Kiara (hindi)"; the base name "Kiara" also
-    # resolves). voice_id is the catalog's opaque id and wins when both are set.
     voice: Optional[str] = None
     voice_id: Optional[str] = None
-    # "kalpa-tts-multilingual-beta-v0.1" (English + Hindi, the default here) or
-    # "kalpa-tts-beta-v0.1" (English). No language parameter — the model detects it,
-    # code-switched Hinglish included.
     model: str = "kalpa-tts-multilingual-beta-v0.1"
-    # Optional sampling controls (Kalpa applies its tuned production defaults when omitted)
     temperature: Optional[float] = None
     top_k: Optional[int] = None
     acoustic_temperature: Optional[float] = None
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -122,7 +122,7 @@ class MayaConfig(BaseModel):
 
 
 class KalpaConfig(BaseModel):
-    voice: Optional[str] = None
+    voice: str = "Kiara"
     voice_id: Optional[str] = None
     model: str = "kalpa-tts-multilingual-beta-v0.1"
     temperature: Optional[float] = None
--- a/bolna/models.py
+++ b/bolna/models.py
@@ -121,6 +121,23 @@ class MayaConfig(BaseModel):
     language: Optional[str] = "en"
 
 
+class KalpaConfig(BaseModel):
+    # Display name from GET /v1/voices (e.g. "Kiara (hindi)"; the base name "Kiara" also
+    # resolves). voice_id is the catalog's opaque id and wins when both are set.
+    voice: Optional[str] = None
+    voice_id: Optional[str] = None
+    # "kalpa-tts-multilingual-beta-v0.1" (English + Hindi, the default here) or
+    # "kalpa-tts-beta-v0.1" (English). No language parameter — the model detects it,
+    # code-switched Hinglish included.
+    model: str = "kalpa-tts-multilingual-beta-v0.1"
+    # Optional sampling controls (Kalpa applies its tuned production defaults when omitted)
+    temperature: Optional[float] = None
+    top_k: Optional[int] = None
+    acoustic_temperature: Optional[float] = None
+    max_new_tokens: Optional[int] = None
+    audio_quality: Optional[str] = None
+
+
 class AzureConfig(BaseModel):
     voice: str
     model: str
@@ -169,6 +186,7 @@ class Synthesizer(BaseModel):
         DeepgramConfig,
         OpenAIConfig,
         MayaConfig,
+        KalpaConfig,
     ] = Field(union_mode="smart")
     stream: bool = False
     buffer_size: Optional[int] = 40  # 40 characters in a buffer
@@ -215,6 +233,9 @@ class Synthesizer(BaseModel):
         elif provider == "maya":
             if isinstance(config, dict):
                 values["provider_config"] = MayaConfig(**config)
+        elif provider == "kalpa":
+            if isinstance(config, dict):
+                values["provider_config"] = KalpaConfig(**config)
 
         return values
 
--- a/bolna/synthesizer/kalpa_synthesizer.py
+++ b/bolna/synthesizer/kalpa_synthesizer.py
@@ -64,8 +64,7 @@ class KalpaSynthesizer(StreamSynthesizer):
         acoustic_temperature=None,
         max_new_tokens=None,
         audio_quality=None,
-        audio_format="pcm",
-        sampling_rate="8000",
+        sampling_rate="24000",
         stream=False,
         buffer_size=400,
         caching=True,
@@ -125,8 +124,19 @@ class KalpaSynthesizer(StreamSynthesizer):
         self._text_buffer = []
         self._buffer_seq = None
         # Set while no response is generating on the socket (one in flight per connection).
+        # The counter tracks outstanding flushes: normally 0/1, briefly 2 when the idle
+        # timeout lets a newer turn flush past a wedged response — responses settle in
+        # flush order, so only the done that drains it back to 0 owns the slot.
         self._response_idle = asyncio.Event()
         self._response_idle.set()
+        self._inflight_flushes = 0
+        # After a cancel, audio frames already on the wire keep arriving until the
+        # response's own responseDone; ids in this set are dropped instead of played.
+        self._ignored_response_ids = set()
+        self._current_response_id = None
+        # A barge-in can land before responseCreated delivers the in-flight response's id;
+        # this flag marks the abandonment id-independently until that response settles.
+        self._abandoned_in_flight = False
 
     def get_sleep_time(self):
         return 0.01
@@ -239,14 +249,19 @@ class KalpaSynthesizer(StreamSynthesizer):
         text) — moot here, since this integration never leaves either behind."""
         self._text_buffer = []
         self._buffer_seq = None
+        if not self._response_idle.is_set():
+            self._abandoned_in_flight = True
+            if self._current_response_id is not None:
+                self._ignored_response_ids.add(self._current_response_id)
+        # The cancelled turn's end-of-stream is never forwarded, so the next turn must be
+        # re-detected as new for stale text_queue entries to be pruned — even when the
+        # cancel send below fails with the socket.
+        self.current_turn_start_time = None
         try:
             ws = self.websocket
             if ws is not None and ws.state is websockets.protocol.State.OPEN:
                 await ws.send(json.dumps({"type": "cancelResponse"}))
                 logger.info("Sent cancelResponse to Kalpa TTS WebSocket")
-            # The cancelled turn's end-of-stream is never forwarded, so the next turn must
-            # be re-detected as new for stale text_queue entries to be pruned.
-            self.current_turn_start_time = None
         except Exception as e:
             logger.error(f"Error handling Kalpa interruption: {e}")
 
@@ -262,16 +277,15 @@ class KalpaSynthesizer(StreamSynthesizer):
                 logger.info(f"Not synthesizing: sequence_id {sequence_id} not current")
                 return
 
-            await self._wait_for_ws()
-
-            # The wait suspends for real while the socket is reconnecting, and a barge-in can
-            # retire this sequence (and clear the shared buffer) in that gap — appending after
-            # it would leak this turn's fragment into the next turn's utterance.
-            if not self.should_synthesize_response(sequence_id):
-                logger.info(f"Not synthesizing (post-wait): sequence_id {sequence_id} not current")
-                return
-
+            # Appends happen before any await: sender tasks start in push order and a
+            # non-EOS sender never suspends, so chunks cannot aggregate out of order even
+            # while the socket is reconnecting. Only the EOS task below ever waits.
             if text:
+                # Guard the ownership invariant: a chunk whose turn's buffer was already
+                # dropped (barge-in between task creation and start) must not append.
+                if self._buffer_seq != sequence_id:
+                    logger.info(f"Dropping Kalpa chunk for superseded seq={sequence_id} (buffer seq={self._buffer_seq})")
+                    return
                 self._text_buffer.append(text)
 
             # Keep buffering until the LLM turn ends: one atomic text+flush frame per turn, so
@@ -280,8 +294,9 @@ class KalpaSynthesizer(StreamSynthesizer):
             if not end_of_llm_stream:
                 return
 
-            # LLM chunks carry their own spacing, so they concatenate verbatim.
-            full_text = "".join(self._text_buffer).strip()
+            # The streaming LLM wrappers split chunks with rsplit(" ", 1) and drop the
+            # boundary space, so it has to be restored here.
+            full_text = " ".join(self._text_buffer).strip()
             self._text_buffer = []
             self._buffer_seq = None
             self.last_text_sent = True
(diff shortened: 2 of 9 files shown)


bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+        terminates with its own responseDone (status "cancelled"), which frees the
+        in-flight slot without emitting an end-of-stream sentinel. On current gateways a
+        bare cancel also wipes undelivered server-side state (queued flush, buffered
+        text) — moot here, since this integration never leaves either behind."""
+        self._text_buffer = []
+        self._buffer_seq = None
+        try:
+            ws = self.websocket
+            if ws is not None and ws.state is websockets.protocol.State.OPEN:
+                await ws.send(json.dumps({"type": "cancelResponse"}))
```
san0808, line 337: After `cancelResponse`, `responseAudio` frames already on the wire keep arriving until `responseDone(cancelled)`. `_generate_ws_loop` pops one `text_queue` entry per yielded frame, so if the next turn's first push lands in that window the stale frames get stamped with the new `sequence_id` and `is_first_chunk=True`, pass the sequence gate in `__listen_synthesizer`, and play as the start of the new reply. Small window, but nonzero under load.

Cartesia handles the same thing with `context_ids_to_ignore`; the Kalpa frames all carry `response_id`, so the same shape fits: track the current id from `responseCreated`, add it to an ignore set here, drop matching `responseAudio` in `receiver`, clear it on its `responseDone`.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+                logger.info(f"Not synthesizing (post-wait): sequence_id {sequence_id} not current")
+                return
+
+            if text:
+                self._text_buffer.append(text)
+
+            # Keep buffering until the LLM turn ends: one atomic text+flush frame per turn, so
+            # an abandoned turn's fragments never sit in a server-side buffer and every turn
+            # settles with exactly one responseDone.
+            if not end_of_llm_stream:
```
san0808: This buffers the whole LLM turn and flushes once at EOS, so first audio waits for full LLM generation plus Kalpa TTFB. Every other streaming synth here (Deepgram, Rime, Sarvam, Smallest) sends each `buffer_size` chunk as it streams and only flushes at EOS. For a 3-4 sentence reply that is roughly 1-2s of extra silence per turn, which is the single biggest thing callers will notice about this provider.

Since Kalpa renders only on flush, the fix is to flush per pushed chunk and yield `b"\x00"` only on the `responseDone` that follows the EOS flush (a `_final_flush_in_flight` flag is enough given single-flight). If whole-turn is deliberate because of the older gateways that rejected mid-generation flushes, please add measured TTFB vs Deepgram or Cartesia to the description so the trade-off is on record.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+                        # is what ends this loop, and auth failures won't fix themselves.
+                        logger.error(f"Kalpa fatal error: {err}")
+                        if err.get("type") == "authentication_error":
+                            self.connection_error = err.get("message") or "authentication error"
+                    else:
+                        # A non-fatal error is a rejected client frame. This integration only
+                        # sends whole-turn flushes, so the rejected turn will never produce
+                        # audio or a responseDone — terminate it and free the in-flight slot.
+                        logger.error(f"Kalpa TTS error: {err}")
+                        self._response_idle.set()
```
san0808: This yields a sentinel unconditionally; the `ConnectionClosed` branch below guards with `if not self._response_idle.is_set()` so an idle-time sentinel cannot pop the next turn's `meta_info`. Per the docs the `error` frame carries `response_id` "when the error belongs to a specific generation", so the precise gate is: terminate the turn only when `response_id` matches the in-flight response (or, simpler, when `_response_idle` is not set); otherwise just log.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+                logger.info(f"Not synthesizing (inner): sequence_id {sequence_id} not current")
+                return
+
+            try:
+                self._response_idle.clear()
+                if self.ws_send_time is None:
+                    self.ws_send_time = time.perf_counter()
+                await self._send_json({"type": "sendText", "text": full_text, "flush": True})
+            except Exception as e:
+                logger.error(f"Error sending text to Kalpa: {e}")
```
san0808: `_send_json` already logs and sets `connection_error` before re-raising, so this branch logs the same failure twice and sets the same field twice. Catch and `return` is enough.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+        self,
+        voice=None,
+        voice_id=None,
+        model=KALPA_DEFAULT_MODEL,
+        temperature=None,
+        top_k=None,
+        acoustic_temperature=None,
+        max_new_tokens=None,
+        audio_quality=None,
+        audio_format="pcm",
```
san0808: Accepted and never read. Drop it or wire it into `_get_audio_format`.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+            full_text = "".join(self._text_buffer).strip()
+            self._text_buffer = []
+            self._buffer_seq = None
+            self.last_text_sent = True
+
+            if not full_text:
+                return
+            if len(full_text) > MAX_TEXT_CHARS:
+                logger.warning(f"Kalpa utterance exceeds {MAX_TEXT_CHARS} chars; truncating")
+                full_text = full_text[:MAX_TEXT_CHARS]
```
san0808: A hard slice cuts mid-word and the caller hears it. If truncation stays, cut at the last sentence or whitespace boundary before the cap.

tests/test_kalpa_synthesizer.py
```diff
@@ -0,0 +1,746 @@
...
+    asyncio.run(s.sender("stale fragment", sequence_id=1, end_of_llm_stream=False))
+    assert s._text_buffer == []
+    assert s.sent == []
+
+
+def test_a_dead_socket_settles_the_in_flight_turn():
+    """A response that dies with the socket never gets its responseDone. The receiver must
+    free the single-flight slot and emit the end-of-stream sentinel so playback terminates
+    instead of waiting on a frame that will never arrive."""
+    import websockets as _ws
```
san0808: Three tests import `websockets` locally (here, `test_interruption_cancels_and_clears_the_buffered_turn`, `test_a_dead_socket_with_nothing_in_flight_stays_silent`); hoist it to the module imports.

san0808 (commented):
Thanks for the contribution, this is a clean integration and follows the StreamSynthesizer pattern well. Two call-behaviour points (interruption audio leak, whole-turn buffering latency) and a few smaller quality items inline.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+                self._text_buffer.append(text)
+
+            # Keep buffering until the LLM turn ends: one atomic text+flush frame per turn, so
+            # an abandoned turn's fragments never sit in a server-side buffer and every turn
+            # settles with exactly one responseDone.
+            if not end_of_llm_stream:
+                return
+
+            # LLM chunks carry their own spacing, so they concatenate verbatim.
+            full_text = "".join(self._text_buffer).strip()
```
san0808: Every streaming LLM wrapper (`openai_llm`, `openai_base`, `azure_llm`, `gemini_llm`, `litellm`) splits at `buffer_size` with `rsplit(" ", 1)` and drops the boundary space; `_handle_llm_output` re-adds it only to the transcript (`llm_response += " " + text_chunk`) and pushes the chunk unchanged. So `"".join` glues a word at every chunk boundary of every turn: `'...let me' + 'check your account'` renders as "let mecheck your account". `" ".join(...)` fixes it.

The aggregation test passes only because it hand-places a leading space the real LLM path never produces; worth fixing that fixture too.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+            return None
+        audio = await self._generate_http(text)
+        if not audio:
+            return None
+        return self._process_http_audio(audio)
+
+    def _process_http_audio(self, audio):
+        """Convert the one-shot WAV to the format the non-streaming loop should emit."""
+        if not audio:
+            return audio
```
san0808: Returns `None` on a failed one-shot render. `_generate_http_loop` has no guard, so the packet reaches `DefaultOutputHandler.handle`, `base64.b64encode(None)` raises, the handler sets `_closed`, and the playground session is mute for the rest of the call. Maya and Smallest return `b"\x00"` here; same fix. `stream=False` is the `models.py` default and is forced for turn-based conversations, so this path is live.

bolna/models.py
```diff
@@ -121,6 +121,17 @@ class MayaConfig(BaseModel):
     language: Optional[str] = "en"
 
 
+class KalpaConfig(BaseModel):
+    voice: Optional[str] = None
```
san0808: This is the only synthesizer config where `voice` can be `None`, and task_manager assumes it is not: `__setup_synthesizer` reads `provider_config["voice"]`, and with backchanneling on it calls `self.synthesizer_voice.lower()` at call setup. A `voice_id`-only agent, which the README and PR body advertise, crashes before any audio. `Synthesizer(provider="kalpa", provider_config={})` also validates today and only fails on the first call.

Simplest fix that matches the siblings: give `voice` a default name (Maya defaults to `"Ananya"`) and keep `voice_id` as the optional override.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,548 @@
...
+
+            # The wait suspends for real while the socket is reconnecting, and a barge-in can
+            # retire this sequence (and clear the shared buffer) in that gap — appending after
+            # it would leak this turn's fragment into the next turn's utterance.
+            if not self.should_synthesize_response(sequence_id):
+                logger.info(f"Not synthesizing (post-wait): sequence_id {sequence_id} not current")
+                return
+
+            if text:
+                self._text_buffer.append(text)
```
san0808: Two consequences of buffering across independent sender tasks, both of which per-chunk flushing (see the thread below on the `end_of_llm_stream` early return) would remove.

1. `_handle_llm_output` drops whitespace-only chunks before `push()`, so when the LLM's final chunk is empty the `end_of_llm_stream` flag never reaches this synth and the entire buffered turn is discarded silently. Siblings only lose the trailing flush.
2. During a reconnect each parked sender sleeps its own 1s in `_wait_for_ws` and wakes out of push order, so chunk 1 can append after chunks 2+3 already flushed and ride into the next turn's utterance; `_on_push` does not catch it because `_buffer_seq` is `None` by then.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Done — the receiver keeps an ignore set keyed on response_id (same shape as Cartesia): a barge-in adds the in-flight id, matching responseAudio is dropped, and the id clears on its responseDone. The window where the barge-in lands before responseCreated delivers the id is covered too. Fixed in 1c361c81.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Reworked — the client now streams each chunk as it arrives. The gateway added server-side segmentation (generation_config.chunk_length_schedule, [50, 80, 120, 150]), so audio starts once buffered text crosses a threshold and ends at a sentence. The turn stays one utterance with one responseDone and continuous prosody, so first audio no longer waits for the full LLM turn. Fixed in f44f1665.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Fixed — the error branch only terminates a turn when one is actually on the connection, and the completion is scoped to the sequence that owns the slot; idle-time errors just log. 1c361c81, tightened in f44f1665.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: That branch is gone — frames now go out through a local _send_frame that logs once and doesn't set connection_error (send failures are transient so the turn can replay after a reconnect). 5d1b97ae.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Dropped in 1c361c81.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Fixed — truncation cuts at the last word boundary before the cap, on both the streaming and one-shot HTTP paths. 1c361c81, shared helper in 5d1b97ae.

tests/test_kalpa_synthesizer.py
gautijha37: Hoisted — the test module has a single top-level import now. 1c361c81.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Fixed — chunks rejoin with the boundary space the wrappers' rsplit consumed, and the fixture was corrected. One refinement since: a chunk with no space in it was an unbreakable token (long URL, number) cut at buffer size, so the next chunk glues on directly instead of getting a phantom space. 1c361c81, refined in bcb02401.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Fixed — a failed one-shot render returns b"\x00" like Maya and Smallest, so the non-streaming loop ends the turn instead of muting the session. 1c361c81.

bolna/models.py
gautijha37: Fixed — KalpaConfig.voice defaults to "Kiara" with voice_id as the optional override, matching Maya's shape. 1c361c81.

bolna/synthesizer/kalpa_synthesizer.py
gautijha37: Both fixed. (1) _handle_llm_output now forwards an empty final buffer so end_of_llm_stream always reaches the synthesizer (4f498658). (2) The whole-turn buffering is gone — senders take a lock in push order and stream chunks immediately, so a reconnect can't reorder fragments, and stale chunks are ownership-guarded (f44f1665).

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,904 @@
...
+        self._dirty_conn_gen = None
+        # The cancelled turn's end-of-stream is never forwarded, so the next turn must be
+        # re-detected as new for stale text_queue entries to be pruned — even when the
+        # cancel send below fails with the socket.
+        self.current_turn_start_time = None
+        try:
+            if not self._response_idle.is_set():
+                self._slot_abandoned = True
+                if self._wire_rid is None and turn_open:
+                    await self._close_ws("barge-in on an un-flushed utterance with no response yet")
```
san0808: Non-blocking, measured against the live API. A barge-in that lands after text was streamed but before a response exists closes the socket, and the next turn paid 3416ms from interrupt to first audio versus 904-1477ms for normal turns in the same session. The redial itself is 905ms; the rest is `monitor_connection`'s 1s cadence plus `_wait_for_ws`'s default 1s poll.

Since the docstring's second case says a flush commits the utterance server-side so a response and its done are guaranteed, would `sendText flush:true` followed by `cancelResponse` settle this state through the existing `responseDone(cancelled)` path in milliseconds, with the created-while-abandoned response dropped by the correlation you already have? If the server cannot guarantee a done for that sequence, the cheaper half is to shorten the redial: poll `_wait_for_ws` at 50ms here and have `_close_ws` trigger the redial instead of waiting out the monitor's 1s tick.

bolna/synthesizer/kalpa_synthesizer.py
```diff
@@ -0,0 +1,904 @@
...
+    # Voice resolution
+    # ------------------------------------------------------------------
+
+    async def _resolve_voice_id(self):
+        """Return the opaque voice id, resolving a display name via GET /v1/voices once.
+
+        Names match case-insensitively, on the full name ("Kiara (hindi)") or its base
+        before the qualifier ("Kiara") — so agent configs don't have to carry UUIDs.
+        """
+        if self.voice_id:
```
san0808, line 261: Non-blocking. The name-to-id lookup is cached on the instance, and there is one instance per call, so every call pays a `GET /v1/voices` round trip (measured ~670ms) before the socket opens. A module-level dict keyed on `(host, name)` makes it once per process.

hvg639: Name→id resolutions now land in a module-level _VOICE_IDS map keyed on (host, lowercased name), so one GET /v1/voices fetch per process serves every later call instead of ~670ms on every call's connect path.

bolna/synthesizer/kalpa_synthesizer.py
hvg639: `handle_interruption` no longer closes the socket when a barge-in lands on an un-flushed utterance with no response yet — it sends `sendText{flush:true}` then `cancelResponse`, so the server-committed utterance's guaranteed `responseDone(cancelled)` frees the slot in milliseconds on the same connection. The identical triage in _open_utterance (superseded turn, no response) got the same treatment, with the socket reset kept only as the path for sends that die with the socket.

san0808 (approved):
lgtm

PR #988, opened 2026-08-26
feat: SSRF-guard custom LLM base_url before outbound requests

san0808:
Makes the `custom` (self-hosted, OpenAI-compatible) LLM provider safe to expose to customers.

**SSRF-guards the customer-supplied `base_url`** before any outbound request, reusing the existing `validate_outbound_url` check. It covers all three places a custom endpoint is dialled: the conversation LLM, the graph-agent routing hop, and the knowledge-base agent. The guard runs once per instance and is a no-op when no `base_url` is set, so default OpenAI traffic is unaffected. A blocked endpoint raises with the resolved address scrubbed, so an internal IP is never spoken to the caller or echoed into the LLM context, and the call ends instead of looping.

**Forces chat completions when `provider == "custom"`.** The Responses API and its `previous_response_id` chaining are OpenAI-specific: a self-hosted endpoint serves the first turn, then 404s on the chained follow-up and the call drops. Real OpenAI and Azure agents keep the Responses API.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.211"
+__version__ = "0.10.212"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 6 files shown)


prateeksachan (approved):

SubhrajitPrusty (approved):
LGTM

PR #989, opened 2026-08-27
fix(trace): stamp graph routing rows at hop start and surface LLM/routing latency

puneet-bolna:
## Problem

Reported by Spinny on exec `677cfe93-8db4-40c9-af09-ab0b53303f88` (graph agent, 4 turns):
"the LLM response was logged long after the LLM request", and "graph-agent trace data is
often not sequential".

Confirmed against the actual trace CSV. Two distinct defects:

**1. `graph_routing` request rows are stamped at hop *completion*, not hop start.**
Both the REQUEST and RESPONSE rows are emitted from the same `yield {"routing_info": …}`
handler in `__do_llm_generation`, which the graph agent only reaches after
`decide_next_node_with_functions()` has already returned. So the request row is displaced
forward by the full routing latency, every hop reads as **0 ms**, and the real cost shows
up as an unexplained gap after the transcriber row:

| seq | ASR final → routing request | `routing_latency_ms` in that row's metadata |
|---|---|---|
| 2 | 696 ms | 691.8 ms |
| 3 | 897 ms | 895.8 ms |
| 4 | 785 ms | 783.7 ms |

**2. The `Latency` column is empty on every `llm` and `graph_routing` row.**
`convert_to_request_log` reads `meta_info["llm_latency"]` — nothing in bolna ever sets that
key (only `_stamp_llm_latency_dict`, which feeds the separate `llm_latencies` progression
struct). `routing_latency_ms` existed only buried in the Metadata JSON. With no latency
field, the only readable number is the request→response timestamp gap — which for a
streamed turn spans the **whole stream**, not the latency. On seq 3 that gap is 1834 ms
while time-to-first-token was 833 ms, and two complete synthesizer request→response pairs
sit inside the window. Correct streaming behaviour, unreadable presentation.

Note: the `llm` response row itself was *not* late. `_handle_llm_output` only
`asyncio.create_task`s the synth and never awaits TTS, so the row was already within ~1 ms
of end-of-stream.

## Changes

- `helpers/utils.py::convert_to_request_log` — optional `ts` (epoch seconds) and `latency`
  (seconds). `ts` stamps `log["time"]` at the moment the event happened rather than the
  moment the row is written; `latency` overrides the `meta_info` lookups. Both default to
  `None`, so all existing call sites are unchanged.
- `agent_types/graph_agent.py` — every `routing_info` now carries `routing_started_at`
  (wall clock at hop start): `_router_hop_info` (new `started_at` kwarg; both
  `_resolve_router_chain` sites pass `hop_started_at`, captured next to the existing
  `hop_start` perf_counter), `_hold_routing_info`, the event-triggered hop, and the main
  hop in `generate()` (stamped before `decide_next_node_with_functions`).
- `agent_manager/task_manager.py` — `graph_routing` REQUEST stamped `ts=routing_started_at`;
  RESPONSE carries `latency=routing_latency_ms/1000`.
- `agent_manager/task_manager.py::__do_llm_generation` — tracks `llm_stream_end_ts` and
  `llm_first_token_latency` (from the chunk's `LatencyData.first_token_latency_ms`,
  converted to seconds to match every other component's `Latency`), threaded through
  `__store_into_history` via new `ts`/`llm_latency` kwargs. The `ts` is a correctness guard
  against future work landing between loop exit and the log call; the `latency` is the fix.

## Verification

Replaying the fix's semantics over the real trace: routing hops go from 0.0/0.1/0.1 ms to
691.8/895.8/783.7 ms, and the seq-3 routing request row moves from +896 ms after the ASR
final to **+1 ms**.

New `tests/test_trace_log_timestamps.py` (5 cases): explicit `ts` honoured, omitted `ts`
falls back to `now()`, explicit latency beats `meta_info`, `latency=0.0` survives (a
deterministic hop really is ~0 s), and a routing request stamped from `routing_started_at`
sorts before its response.

`1181 passed, 1 skipped`. ruff check + format clean.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.217"
+__version__ = "0.10.218"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.212"
+__version__ = "0.10.213"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -3510,6 +3510,8 @@ class TaskManager(BaseManager):
         reasoning_content=None,
         log_message=None,
         overflowed=False,
+        ts=None,
+        llm_latency=None,
     ):
         self.llm_response_generated = True
         # task 0 only, so aux LLMs (hangup/voicemail) never tally, and never an overflowed turn,
@@ -3533,6 +3535,8 @@ class TaskManager(BaseManager):
             reasoning_tokens=reasoning_tokens,
             cached_tokens=cached_tokens,
             reasoning_content=reasoning_content,
+            ts=ts,
+            latency=llm_latency,
         )
         turn_id = meta_info.get("turn_id")
         response_uid = meta_info.get("response_uid")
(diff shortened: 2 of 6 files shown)


SubhrajitPrusty (approved):

san0808 (approved):
lgtm

PR #995, opened 2026-08-28
fix: refresh s2s caller liveness on a provider speech start

san0808:
On a speech-to-speech agent, the provider's speech-start event only updated the interruption accounting. It did not refresh `time_since_last_spoken_human_word`, which the caller-liveness timers read. A caller who was actively talking but not barging in could therefore be treated as silent, triggering the "are you still there" prompt or an inactivity hangup mid-conversation.

A speech start is proof the caller is talking, so it now refreshes the liveness timestamp whether or not the turn is a barge-in.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.214"
+__version__ = "0.10.215"
 
 import os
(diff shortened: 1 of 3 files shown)


san0808:
Folded into #994.

PR #1009, opened 2026-09-02
fix: stall backstop can be permanently defeated by a hung LLM task (BOLNA-2563)

aasthachauhan-code:
## Problem

Reported on two calls (execs `d409c7cd…`, `7c70407a…`, both Plivo): the agent
never disconnected after the callee's voicemail response, and the call sat in ~90-130s of
dead silence until the carrier itself dropped the line. Both calls also came back with an
empty transcript. `hangup_after_silence` and voicemail detection were both enabled.

**Direct fix** — _run_llm_task now wraps generation in asyncio.wait_for(LLM_GENERATION_TIMEOUT_S) (60s). A hang raises a timeout, routes through the normal error-hangup path, and the call recovers on its own.
**Backstop** — _should_stall_hangup catches anything else that leaves the pipeline looking stuck (no audio playing, both sides silent) past STALL_HANGUP_HARD_CAP_S, now tuned down to match at 60s.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -3210,7 +3210,15 @@ class TaskManager(BaseManager):
 
             self.hangup_detail = HangupReason.END_CALL_TOOL
             self.call_hangup_message_config = None
-            await self.process_call_hangup()
+            # Detached, not awaited: this call site runs inside __do_llm_generation's own
+            # LLM_GENERATION_TIMEOUT_S window (the end_call tool is handled here, mid-generation).
+            # process_call_hangup's teardown (goodbye playout wait, hangup-chunk wait,
+            # stop_handler) can legitimately take longer than that and must not be cancelled
+            # mid-way - same failure mode as BOLNA-2563, reached via the end_call tool instead
+            # of the hangup-decision check. Held on self, not _usage_tasks, so a reference
+            # survives (see _usage_tasks' own comment) without an unrelated cleanup sweep
+            # ever cancelling this specific task.
+            self._end_call_hangup_task = asyncio.create_task(self.process_call_hangup())
             return
 
         if called_fun.startswith("transfer_call"):
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -42,9 +42,10 @@ from bolna.constants import (
     END_CALL_FUNCTION_PREFIX,
     END_CALL_TOOL_DEFINITION,
     RESPONSES_API_MODEL_PREFIXES,
+    LLM_GENERATION_TIMEOUT_S,
     S2S_GOODBYE_TIMEOUT_S,
     S2S_STREAM_SID_TIMEOUT_S,
-    STALL_HANGUP_FLOOR_S,
+    STALL_HANGUP_HARD_CAP_S,
     STUCK_AUDIO_GATE_RELEASE_S,
     WEB_BASED_CALL_PROVIDER,
     WEBCALL_TTS_SAMPLE_RATE,
@@ -4437,9 +4438,11 @@ class TaskManager(BaseManager):
 
         try:
             if self._is_extraction_task() or self._is_summarization_task():
-                await self._process_followup_task(message)
+                await asyncio.wait_for(self._process_followup_task(message), timeout=LLM_GENERATION_TIMEOUT_S)
             elif self._is_conversation_task():
-                await self._process_conversation_task(message, sequence, meta_info)
+                await asyncio.wait_for(
+                    self._process_conversation_task(message, sequence, meta_info), timeout=LLM_GENERATION_TIMEOUT_S
+                )
             else:
                 logger.error("unsupported task type: {}".format(self.task_config["task_type"]))
             self.llm_task = None
@@ -4448,6 +4451,21 @@ class TaskManager(BaseManager):
             self._synthesis_awaiting_first_audio = False
             await self._end_call_on_component_error(e, HangupReason.LLM_ERROR)
             raise
+        except asyncio.TimeoutError:
+            # The LLM never came back within LLM_GENERATION_TIMEOUT_S seconds - treat it as
+            # stuck and hang up, same as any other LLM failure, instead of waiting forever.
+            # BOLNA-2563.
+            logger.error(f"LLM generation stuck for {LLM_GENERATION_TIMEOUT_S}s (run_id={self.run_id}) - hanging up")
+            self.response_in_pipeline = False
+            self._synthesis_awaiting_first_audio = False
+            await self._end_call_on_component_error(
+                LLMError(
+                    f"LLM generation timed out after {LLM_GENERATION_TIMEOUT_S}s",
+                    provider=self.llm_config.get("provider", "unknown"),
+                    model=self.llm_config.get("model"),
+                ),
+                HangupReason.LLM_ERROR,
+            )
         except Exception as e:
             self.response_in_pipeline = False
             self._synthesis_awaiting_first_audio = False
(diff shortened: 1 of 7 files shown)


vikramsingh-bolna (approved):

bolna/constants.py
```diff
@@ -22,8 +22,17 @@
...
+# made no forward progress at all, e.g. a hung LLM call or an s2s tool task that never
+# completes. BOLNA-2563.
+STALL_HANGUP_HARD_CAP_S = 60.0
+# Max time to wait for one LLM reply (_run_llm_task). If the LLM never comes back - no reply,
+# no error, it just hangs - force it to fail after this many seconds and hang up, instead of
+# leaving the call stuck forever. BOLNA-2563.
+LLM_GENERATION_TIMEOUT_S = 60.0
+# Same idea, applied to the voicemail-check LLM call (agent_manager/voicemail_handler.py),
+# which has its own unguarded await and is not covered by _run_llm_task's timeout. BOLNA-2563.
+VOICEMAIL_CHECK_TIMEOUT_S = 12.0
```
san0808: This constant is not referenced anywhere. The voicemail check is still unguarded: `_background_check` awaits `check_for_voicemail` directly, and it runs in its own task so `_run_llm_task` wait_for does not reach it.

This is not theoretical for the reported calls. The voicemail check was triggered on both (7c70407a at t=2.2s, d409c7cd at t=5.0s) and never returned, at the same moment the conversation LLM hung. Half of what actually hung is still unguarded. Either wire it into `_background_check` or drop the constant, because right now the comment claims a fix that does not exist.

bolna/agent_manager/task_manager.py
```diff
@@ -4437,9 +4438,11 @@ async def _run_llm_task(self, message):
 
         try:
             if self._is_extraction_task() or self._is_summarization_task():
-                await self._process_followup_task(message)
+                await asyncio.wait_for(self._process_followup_task(message), timeout=LLM_GENERATION_TIMEOUT_S)
             elif self._is_conversation_task():
-                await self._process_conversation_task(message, sequence, meta_info)
+                await asyncio.wait_for(
```
san0808: This wraps more than generation. `_process_conversation_task` can reach `process_call_hangup`, which reaches `wait_for_current_message` and `__process_end_of_conversation` (goodbye playout, hangup-chunk wait, `stop_handler`). If the 60s budget runs out in there, the teardown is cancelled partway with `_end_of_conversation_in_progress` and `_hangup_processing` already True, and the `except asyncio.TimeoutError` handler then calls `_end_call_on_component_error`, which skips the hangup because of that same guard. A slow hangup becomes a permanent one, which is the failure mode this PR exists to fix.

Suggest scoping the timeout to `__do_llm_generation`.

bolna/agent_manager/task_manager.py
```diff
@@ -4448,6 +4451,21 @@ async def _run_llm_task(self, message):
             self._synthesis_awaiting_first_audio = False
             await self._end_call_on_component_error(e, HangupReason.LLM_ERROR)
             raise
+        except asyncio.TimeoutError:
```
san0808, line 4492: `_run_llm_task` is also the post-call extraction and summarization entry point, where there is no live call to rescue. The only effect of the cap there is turning a slow-but-successful extraction into a lost one, plus a `_report_provider_health(..., False)` and a `__process_end_of_conversation()` on a task manager that has no telephony input handler. Worth skipping the timeout for the followup task.

bolna/agent_manager/task_manager.py
```diff
@@ -7252,22 +7270,23 @@ async def _inject_and_run_llm(self, injected_message: str):
...
             return
 
-    def _should_stall_hangup(
-        self, audio_playing, has_pending_generation, time_since_last_spoken_ai_word, time_since_user_last_spoke
-    ):
-        """Hang up when the call has made no forward progress at all: no audio playing, nothing
-        in flight, and both sides silent past the floor. Runs above the audio/pipeline gate so it
-        still applies when a pipeline flag is stuck. Floor stays above hangup_after_silence so the
-        normal inactivity path takes precedence whenever it is reachable."""
+    def _should_stall_hangup(self, audio_playing, time_since_last_spoken_ai_word, time_since_user_last_spoke):
```
san0808, line 7313: Two things from dropping the floor and the pending-generation check:

- The cap no longer respects `hangup_after_silence`. #758 used `max(hang_conversation_after, floor)` on purpose so the configured path always won. 166 prod agents are configured above 60s (only ~6 calls in the last 7 days, so impact is small, but it is a one-line `max()` to keep). It also makes `trigger_user_online_message_after` and `repeat_after_silence_seconds` above 60 unreachable, since this check runs above them.
- The comment above `has_pending_generation` at the call site still says the s2s clause is there so a long tool call is not read as no forward progress and hung up mid-tool. That is now what happens at 60s, and the same goes for the transfer path that waits on `is_audio_being_played_to_user` before its POST. Either keep the clause with a longer cap, or update that comment so it stops contradicting the code.

bolna/agent_manager/task_manager.py
```diff
@@ -7348,14 +7367,13 @@ async def __check_for_completion(self):
...
-                has_pending_generation=has_pending_generation,
                 time_since_last_spoken_ai_word=time_since_last_spoken_ai_word,
                 time_since_user_last_spoke=time_since_user_last_spoke,
             ):
                 logger.warning(
                     f"Stall backstop: no forward progress for {time_since_last_spoken_ai_word:.1f}s "
-                    f"(audio_playing=False, no pending generation, response_in_pipeline={self.response_in_pipeline}) "
-                    f"- forcing hangup"
+                    f"(audio_playing=False, has_pending_generation={has_pending_generation}, "
+                    f"response_in_pipeline={self.response_in_pipeline}) - forcing hangup"
```
san0808, line 7421: Separate from this PR trigger but the same hole: this branch calls `_hangup_after_goodbye` and then breaks. That routes to `process_call_hangup`, which returns immediately if `_hangup_processing` or `conversation_ended` is set, and neither flag is ever reset. So against a hangup that is already half-done, the backstop fires once, does nothing, and kills the only watchdog. Worth making the last-resort path bypass those guards while this file is open.

tests/test_hung_llm_task_defeats_hangup_backstops.py
```diff
@@ -0,0 +1,115 @@
+"""BOLNA-2563: a hung LLM task (no token, no error, never completes) used to silently disable
```
san0808, line 1: These do not cover the fix. All five tests call the pure predicate; none go through the `wait_for` in `_run_llm_task`. `test_hung_llm_task_keeps_has_pending_generation_true_forever` asserts on `_has_pending_generation`, which is defined in this file, so it tests the test. The `hung_llm_task` fixture is requested by tests that never use it.

The docstring premise is also wrong: `response_in_pipeline` is cleared in the generation path and the audio-playback path too, not only in the exception handlers of `_run_llm_task`. And the line-number references are already stale, better to name the function.

bolna/agent_manager/task_manager.py
```diff
@@ -3209,7 +3210,15 @@ async def __execute_function_call(
...
-            await self.process_call_hangup()
+            # Detached, not awaited: this call site runs inside __do_llm_generation's own
+            # LLM_GENERATION_TIMEOUT_S window (the end_call tool is handled here, mid-generation).
+            # process_call_hangup's teardown (goodbye playout wait, hangup-chunk wait,
+            # stop_handler) can legitimately take longer than that and must not be cancelled
+            # mid-way - same failure mode as BOLNA-2563, reached via the end_call tool instead
+            # of the hangup-decision check. Held on self, not _usage_tasks, so a reference
+            # survives (see _usage_tasks' own comment) without an unrelated cleanup sweep
+            # ever cancelling this specific task.
+            self._end_call_hangup_task = asyncio.create_task(self.process_call_hangup())
```
san0808, line 3222: Detaching the teardown is the right call, but this `create_task` has no done callback and is never awaited or cancelled, so if `process_call_hangup` raises, the exception disappears with the task and a failed hangup looks identical to a successful one.

`_s2s_track_task` in this same file exists for exactly this ("A bare discard callback drops the exception along with the task, so a tool result that never reached the model looked exactly like one that succeeded"). Worth routing through it, or at least adding a callback that logs.

san0808 (approved):

PR #1010, opened 2026-09-02
feat: accept per-call Deepgram host override for multi-cloud ASR routing

SubhrajitPrusty:
## What
Teach `DeepgramTranscriber` to accept a **per-call host** so a caller can pin a call to a specific self-hosted Deepgram endpoint, instead of only the fleet-wide `DEEPGRAM_HOST` env.

- `bolna/models.py` — add `deepgram_host` / `deepgram_flux_host` / `deepgram_host_protocol` to the `Transcriber` model. Explicit fields (not kwarg passthrough) so they survive pydantic validation.
- `bolna/transcriber/deepgram_transcriber.py` — `__init__` prefers the kwarg, falls back to env (`kwargs.get(...) or os.getenv(...)`), so behaviour is unchanged when unset. Nova + Flux WS URL builders use the resolved protocol. Served host recorded in `meta_info["deepgram_host"]`.

## Backward compatibility
Fully backward compatible — all three fields default to `None`; with none set, host and protocol resolve exactly as before.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.230"
+__version__ = "0.10.231"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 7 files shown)


san0808 (approved):
lgtm

PR #1011, opened 2026-09-02
 fix(llm): stop speaking Azure content-filter errors to the caller

puneet-bolna:
**Problem** 

On few calls  the agent read an internal error out loud mid-conversation: "An error occurred: Error code: 400 - … The response was filtered due to the prompt triggering Azure OpenAI's content management policy…". Both runs broke at the identical turn — the caller said "my school was not good … planning to attempt it again" (re-sitting the GRE). Azure returned 400 ResponsibleAIPolicyViolation, param: 'prompt' (input filtered before any completion), one category: self_harm/medium.

**Root cause** 

 Two defects stacked. (1) AzureLLM._generate_stream_chat caught BadRequestError only to re-raise it; LiteLLM has handled this correctly since litellm.py:98 but the SDK-based providers never got the equivalent. (2) GraphAgent.generate's blanket except Exception yielded f"An error occurred: {e}" as an LLMStreamChunk — indistinguishable from model output, so it was synthesized and written into conversation_history, then re-sent to the LLM every following turn. knowledgebase_agent.py had the identical line; the simple agent never did, which is why it already ended the call cleanly.

**What changed** 

 LLM layer: is_content_filter_error() / note_content_filter() in openai_base.py, wired into the Azure, OpenAI, and Responses-API stream paths (the last placed ahead of the stale-previous_response_id check, which treats any 400 as stale and would burn a wasted retry). Agent layer: graph and KB agents log and raise instead of speaking. Trace: format_error_message now matches Azure's "content management policy" wording, which previously fell through and printed the raw 400 to the customer.

**Behaviour change to flag** 

 a graph-agent call hitting a non-content-filter LLM error now ends the call (LLM_ERROR) where it previously spoke the error and continued.

**Open decision** 

 whether to retry a filtered turn on the existing direct-OpenAI _overflow_client: A ship as-is/silent turn, B overflow retry, C neutral reprompt. Recommendation is A, with the real cure client-side — raising Meritto's Azure self_harm prompt threshold on the PTU deployment.

Tests — 14 new tests; red/green proven by reverting the behavioural files to HEAD (7 fail, graph agent reporting DID NOT RAISE). 1403 passed, 1 skipped, ruff clean. Includes the incidental stale-mock bug the old except had been hiding.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.224"
+__version__ = "0.10.225"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_types/graph_agent.py
+++ b/bolna/agent_types/graph_agent.py
@@ -1426,10 +1426,7 @@ class GraphAgent(BaseAgent):
                 yield chunk
 
         except Exception as e:
+            # Never yield the error as text: a chunk here is indistinguishable from model output
+            # and gets spoken. Propagate so the task manager ends the call with an LLMError.
             logger.error(f"Error in generate: {e}")
-            latency_data = LatencyData(
-                sequence_id=meta_info.get("sequence_id") if meta_info else None,
-                first_token_latency_ms=0,
-                total_stream_duration_ms=now_ms() - start_time,
-            )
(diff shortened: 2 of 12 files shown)


SubhrajitPrusty (approved):
LGTM

san0808 (approved):

PR #1012, opened 2026-09-02
fix(llm): stop a cancelled WS turn from leaking its answer into the next turn

puneet-bolna:
Problem 

 after a barge-in the agent answered the previous utterance (user: ठीक है / user: <loan question> / assistant: कुछ और पूछना था?). The per-sequence latencies prove no generation happened: seq 4 cancelled, seq 5 = 7.81 ms, seq 6 = 0.52 ms, seq 7 = 1160 ms (normal again). Loki shows zero LLM activity across the 4-second gap.

Root cause

OpenAIWSConnection holds one WebSocket for the whole call (connected once at 12:58:56.603, never reconnected across 8 sequences). stream_response yields every event with no response_id filter. llm_task.cancel() abandons the generator mid-stream, leaving the cancelled response's events queued; the next turn reads them and returns on their terminal event. The server-side cancel can't help — it's gated on previous_response_id, only set on CREATED, which hadn't arrived 6 ms after the create.

Fix 

 _needs_reset on the connection: set when the create is sent, cleared only when a terminal event is actually consumed, and ensure_connected reconnects when set. 17 insertions / 2 deletions, one class. Placed in stream_response rather than at a caller because kickoff_llm_generation also cancels llm_task directly — the overlapping-speech path, i.e. this exact bug. Notes the load-bearing clear-before-yield ordering.

Tests 

15 tests in three groups (healthy path unchanged / abandonment reconnects / end-to-end through the real consumer). Red/green proven by reverting to HEAD: 7 of 15 fail, the end-to-end one as assert 'loan answer' in 'कुछ और पूछना था?'. Suite 1404 passed / 1 skipped, ruff clean.

### DIFF
--- a/bolna/llms/openai_llm.py
+++ b/bolna/llms/openai_llm.py
@@ -51,6 +51,7 @@ class OpenAIWSConnection:
     def __init__(self, api_key: str):
         self._api_key = api_key
         self._ws = None
+        self._needs_reset = False
         self._connected_at: float = 0
         self._lock = asyncio.Lock()
         self._connect_task: Optional[asyncio.Task] = None
@@ -72,7 +73,12 @@ class OpenAIWSConnection:
             self._connect_task = None
 
         if self._ws is not None:
-            if self._ws.state is not WSState.OPEN:
+            if self._needs_reset:
+                # An abandoned stream left its events queued here; reading them as the next
+                # turn's would answer the previous utterance.
+                logger.info("WebSocket dirty after an abandoned stream, reconnecting")
+                await self._close_ws()
+            elif self._ws.state is not WSState.OPEN:
                 logger.info("WebSocket closed unexpectedly, reconnecting")
                 await self._close_ws()
             elif time.monotonic() - self._connected_at >= self.RECONNECT_BEFORE_SECS:
@@ -81,6 +87,7 @@ class OpenAIWSConnection:
             else:
                 return
 
+        self._needs_reset = False
         await self._connect()
 
     async def _connect(self):
@@ -99,13 +106,21 @@ class OpenAIWSConnection:
         async with self._lock:
             await self.ensure_connected()
             await self._ws.send(json.dumps({"type": ResponseStreamEvent.CREATE, **create_params}))
+            # Cleared only once a terminal event is actually consumed. Any other exit —
+            # cancellation, error, the consumer breaking early — leaves the response's
+            # remaining events on the socket, so it must not be reused.
+            self._needs_reset = True
 
             async for raw_msg in self._ws:
                 evt = json.loads(raw_msg)
                 evt_type = evt.get("type", "")
-                yield evt
                 if evt_type in self.TERMINAL_EVENTS:
+                    # Before the yield: consumers break on COMPLETED/INCOMPLETE and never
+                    # let this generator resume to run the line after.
+                    self._needs_reset = False
+                    yield evt
                     return
+                yield evt
 
     async def cancel_response(self, response_id: str):
         """Best-effort cancel: signal the server to terminate the in-flight
--- a/bolna/llms/openai_llm.py
+++ b/bolna/llms/openai_llm.py
@@ -74,8 +74,7 @@ class OpenAIWSConnection:
 
         if self._ws is not None:
             if self._needs_reset:
-                # An abandoned stream left its events queued here; reading them as the next
-                # turn's would answer the previous utterance.
+                # An abandoned stream's queued events would be read as the next turn's.
                 logger.info("WebSocket dirty after an abandoned stream, reconnecting")
                 await self._close_ws()
             elif self._ws.state is not WSState.OPEN:
@@ -106,17 +105,13 @@ class OpenAIWSConnection:
         async with self._lock:
             await self.ensure_connected()
             await self._ws.send(json.dumps({"type": ResponseStreamEvent.CREATE, **create_params}))
-            # Cleared only once a terminal event is actually consumed. Any other exit —
-            # cancellation, error, the consumer breaking early — leaves the response's
-            # remaining events on the socket, so it must not be reused.
-            self._needs_reset = True
+            self._needs_reset = True  # cleared only once a terminal event is consumed
 
             async for raw_msg in self._ws:
                 evt = json.loads(raw_msg)
                 evt_type = evt.get("type", "")
                 if evt_type in self.TERMINAL_EVENTS:
-                    # Before the yield: consumers break on COMPLETED/INCOMPLETE and never
-                    # let this generator resume to run the line after.
+                    # Before the yield: consumers break here and never let this resume.
                     self._needs_reset = False
                     yield evt
                     return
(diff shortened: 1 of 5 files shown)


SubhrajitPrusty (approved):
LGTM

bolna/llms/openai_llm.py
```diff
@@ -99,13 +105,17 @@ async def stream_response(self, create_params: dict):
         async with self._lock:
             await self.ensure_connected()
             await self._ws.send(json.dumps({"type": ResponseStreamEvent.CREATE, **create_params}))
+            self._needs_reset = True  # cleared only once a terminal event is consumed
```
san0808: Set this before the `send`, not after.

If the barge-in cancels the task while `await self._ws.send(...)` is suspended on `drain()` (plausible for a chained create carrying full history plus tools), the frame is already on the wire and the flag never gets set. Socket reads clean, no reconnect, next turn inherits this response's events: the original bug, unfixed on the narrower window. Reproduced by patching `send` to suspend after delivering the frame and then cancelling, which gives `needs_reset=False`, `connects=1`, and turn 2 reading the stale text.

The asymmetry favours marking first: a create that never went out costs one needless reconnect, a create that did costs a leaked turn. websockets also documents that a cancelled `send()` leaves the connection inconsistent, which is a second reason to mark before rather than after.

bolna/llms/openai_llm.py
```diff
@@ -99,13 +105,17 @@ async def stream_response(self, create_params: dict):
...
 
             async for raw_msg in self._ws:
                 evt = json.loads(raw_msg)
                 evt_type = evt.get("type", "")
-                yield evt
                 if evt_type in self.TERMINAL_EVENTS:
+                    # Before the yield: consumers break here and never let this resume.
+                    self._needs_reset = False
```
san0808: `TERMINAL_EVENTS` includes `ERROR`, so this clears the flag on any `error` event. `completed`/`failed`/`incomplete` are response-lifecycle terminals; `error` is an inference, and the consumer below handles `error` and `response.failed` in separate branches, which implies they are not always alternatives. A mid-stream `error` followed by a trailing terminal event leaves that event queued on a socket now marked clean, which is exactly the bug this PR fixes.

Suggest clearing only on `COMPLETED`/`FAILED`/`INCOMPLETE` and leaving `error` dirty. Worst case there is one reconnect on a turn that already failed.

bolna/llms/openai_llm.py
```diff
@@ -72,7 +73,11 @@ async def ensure_connected(self):
             self._connect_task = None
 
         if self._ws is not None:
-            if self._ws.state is not WSState.OPEN:
+            if self._needs_reset:
```
san0808, line 78: This puts a serial close handshake (`close_timeout=5`) plus a fresh TLS and WS upgrade inside the lock, before the create goes out, on the reply after every barge-in. TCP+TLS to api.openai.com measures 33-114ms on its own. It used to happen once per 55 minutes; now it is several times in a noisy call, and the caller hears it as dead air at the worst moment.

Two changes make it close to free: `ensure_future` the old socket's `close()` instead of awaiting it, and start the reconnect when the stream is abandoned rather than when the next turn needs it. `start_connect()` / `_connect_task` already exist for that, so the next turn finds a warm socket.

tests/test_ws_responses_stream_reset.py
```diff
@@ -0,0 +1,306 @@
+"""A cancelled WS Responses turn must not leak its events into the next turn.
+
+One socket serves every turn and `stream_response` reads it without a response_id filter, so
+an abandoned stream's leftovers get read as the next turn's (run 23323c9e). The guarantee: no
```
san0808: Drop the run id from this docstring and from `test_real_consumer_does_not_inherit_a_cancelled_turns_answer`; this repo is public. `INIT.md` is not in the repo either, so that pointer is dead for anyone reading the test.

puneet-bolna: done

bolna/llms/openai_llm.py
puneet-bolna: Close is handed to the loop instead of awaited, and the replacement is warmed at abandonment through the existing start_connect() / _connect_task, so the next turn finds a connected socket. The lazy branch in ensure_connected stays as a backstop — a generator abandoned without cancellation can be finalized late — but it now uses the same non-blocking discard rather than await self._close_ws().

bolna/llms/openai_llm.py
puneet-bolna: Added ResponseStreamEvent.response_terminal_events() = {COMPLETED, FAILED, INCOMPLETE}. The loop still returns on all four terminals; it only clears on a response terminal, so an error leaves the socket dirty.

bolna/llms/openai_llm.py
puneet-bolna: Mark before the send — added

san0808 (commented):
Two small things left, one of them worth fixing before merge.

bolna/llms/openai_llm.py
```diff
@@ -131,6 +149,23 @@ async def _close_ws(self):
...
             self._ws = None
 
+    def discard_socket(self, prewarm=True):
+        """Drop a socket carrying an abandoned response. The close is not awaited and the
+        replacement is warmed in the background, so no handshake lands on the next turn."""
+        ws, self._ws = self._ws, None
+        self._needs_reset = False
+        if ws is not None:
+            asyncio.ensure_future(self._close_quietly(ws))
+        if prewarm:
```
san0808, line 167: `disconnect()` races this prewarm and leaks an open socket.

`discard_socket()` sets `_ws = None` and fires `start_connect()`. If the call tears down while that handshake is in flight, `disconnect()` calls `_close_ws()`, which no-ops because `_ws` is already `None`, and then the prewarm completes and assigns a live socket that nothing ever closes. Verified: after `await disconnect()`, `_ws` holds an unclosed socket.

The window is one handshake wide, but it now opens after every barge-in rather than once at call start, and a caller interrupting and then hanging up is a normal way for a call to end. `disconnect()` should cancel `_connect_task` (and close whatever it produced) before `_close_ws()`.

tests/test_ws_responses_stream_reset.py
```diff
@@ -0,0 +1,415 @@
...
+    t = _transport([delta("hi"), completed()], [completed("r2")])
+    async for evt in t.stream_response({"input": "a"}):
+        if evt["type"] == ResponseStreamEvent.COMPLETED.value:
+            break
+    assert t._needs_reset is False
+    await _drain(t.stream_response({"input": "b"}))
+    assert t.connects == 1
+
+
+@pytest.mark.parametrize("terminal", ["response.failed", "response.incomplete", "error"])
```
san0808: The `"error"` param now passes for the wrong reason: the flag is zero after the drain because the `finally`'s `discard_socket()` zeroed it, not because `error` settled the response. With the test named `test_every_terminal_event_clears_the_flag`, it reads as asserting the opposite of the contract this PR just established.

Drop `"error"` from the parametrize; `test_an_error_event_does_not_mark_the_socket_clean` covers it correctly, and asserting on the discard rather than the flag is the right shape for it.

puneet-bolna: done

bolna/llms/openai_llm.py
puneet-bolna: fixed

san0808 (approved):
lgtm

PR #1015, opened 2026-09-03
fix: trim an interrupted static node to what was actually spoken

puneet-bolna:
Problem. Barge-in during a static (say) node left the whole node message in history and the transcript, not the part the caller heard.

Root cause. Static clips bypass the synthesizer, which is what normally stamps meta_info["text_synthesized"] (base_synthesizer.py:156), and __send_preprocessed_audio sends them with yield_in_chunks = False. So a static node produces exactly one mark, carrying text_synthesized = "" plus the whole clip's duration.

In sync_history that dead-ends twice: the pending_chunks loop skips marks with no text, so the time-proportional estimator never runs; and the duration-only fallback sees a single mark where pending_dur == total_dur, giving proportion 0. With one mark, no text, and heard-vs-pending being all-or-nothing, no partial is recoverable — you get the full row or none.

Fix. Stamp meta_info["text_synthesized"] = static_text on the static packet. Nothing else was needed: estimate_played_text_for_time already slices a single chunk proportionally by elapsed time and trims to whole words, and sent_ts/duration were already on the mark.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -6918,7 +6918,9 @@ class TaskManager(BaseManager):
     def _stamp_cached_clip_text(self, meta_info):
         """Give the mark its text so an interrupted clip can be trimmed. Cached-send branch
         only: the cache-miss path re-synthesizes live and would stamp every chunk. See INIT.md."""
-        if meta_info.get("message_category") != "static_node" or meta_info.get("is_silence_trigger"):
+        if meta_info.get("message_category") not in CACHED_SINGLE_MARK_CATEGORIES or meta_info.get(
+            "is_silence_trigger"
+        ):
             return
         meta_info["text_synthesized"] = meta_info.get("text", "")
 
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -6916,14 +6916,8 @@ class TaskManager(BaseManager):
             await self.tools["synthesizer"].cleanup()
 
     def _stamp_cached_clip_text(self, meta_info):
-        """A pre-generated clip skips the synthesizer, which is what normally stamps
-        text_synthesized, so its single mark would carry no text and an interrupted node could
-        not be trimmed. Only on the branch that sends the cached bytes: the cache-miss path
-        re-synthesizes live with this same meta_info, and every streaming provider but
-        ElevenLabs and Azure reuses one meta_info for all chunks, which would put the whole
-        message on every mark. Silence re-prompts stay unstamped: history deliberately never
-        records them, and mark text would let sync_history materialise an interrupted one.
-        """
+        """Give the mark its text so an interrupted clip can be trimmed. Cached-send branch
+        only: the cache-miss path re-synthesizes live and would stamp every chunk. See INIT.md."""
         if meta_info.get("message_category") != "static_node" or meta_info.get("is_silence_trigger"):
             return
         meta_info["text_synthesized"] = meta_info.get("text", "")
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -22,6 +22,7 @@ from pydub import AudioSegment
 
 from bolna.constants import (
     ACCIDENTAL_INTERRUPTION_PHRASES,
+    CACHED_SINGLE_MARK_CATEGORIES,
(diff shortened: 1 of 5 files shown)


SubhrajitPrusty (approved):
LGTM

bolna/agent_manager/task_manager.py
```diff
@@ -3821,6 +3828,9 @@ async def __do_llm_generation_impl(
 
                     meta_info["end_of_llm_stream"] = True
                     meta_info["text"] = static_text
+                    # The clip skips the synthesizer, which is what normally stamps this. Without
+                    # it the mark carries no text and an interrupted node cannot be trimmed.
+                    meta_info["text_synthesized"] = static_text
```
san0808: On a cache miss `__send_preprocessed_audio` re-synthesizes live with this same meta_info, and for streaming providers other than ElevenLabs/Azure nothing re-stamps `text_synthesized` per chunk: `_generate_ws_loop` pops it once into `self.meta_info` and yields it on every chunk, so all N marks carry the whole message and `estimate_played_text_for_time` concatenates it N times on barge-in. Should the stamp move into `__send_preprocessed_audio` on the branch that actually sends the cached clip, so it never reaches the live path? Repro: point a static node at a missing S3 clip on a Cartesia agent and barge in mid-clip.

san0808: `if not is_silence_trigger:` above deliberately keeps the static text out of history, but with text on the mark `sync_history` now reaches `upsert_assistant_for_response`, which appends when nothing matches the uid. So a silence re-prompt that gets barged into comes back into history, while one that plays in full stays out. Is that asymmetry intended?

bolna/constants.py
```diff
@@ -77,6 +77,9 @@
 SONIOX_DEFAULT_MULTILINGUAL_HINTS = ["en", "hi", "ta", "te", "kn", "ml", "mr", "bn", "gu", "pa", "ur"]
 SONIOX_AUTO_LANGUAGE_VALUES = {"", "multi", "auto", "multilingual", "unknown"}
 
+# Pre-generated clips that bypass the synthesizer and go out as one mark covering the whole
+# message, unlike streamed speech which arrives as many marks.
+CACHED_SINGLE_MARK_CATEGORIES = ("static_node", "event_proactive")
```
san0808: `event_proactive` sets `sequence_id: -1`, and all four output handlers blank `text_synthesized` when `sequence_id == -1`, so its mark never carries text: the hangup shortcut cannot fire for it and it cannot be trimmed either. Should it come out of the tuple until that path is stamped too?

puneet-bolna: removed

bolna/agent_manager/task_manager.py
puneet-bolna: now it carries is_silence_trigger onto the packet and the helper skips it — both decisions stay in one place rather than adding a second guard in sync_history.

puneet-bolna: stream_synthesizer._generate_ws_loop pops one meta_info per turn and yields it on every chunk; only ElevenLabs overrides _unpack_receiver_message to re-stamp per chunk (Azure stamps its own). Cartesia, Deepgram, Rime, Sarvam, Smallest, Maya and Kalpa all inherit the default return item, {}. So on a cache miss the whole message would land on every mark and estimate_played_text_for_time would concatenate it N times.

New _stamp_cached_clip_text(meta_info) helper, called on the two branches that send cached bytes —

bolna/agent_manager/task_manager.py
```diff
@@ -6907,6 +6915,13 @@ async def __listen_synthesizer(self):
         finally:
             await self.tools["synthesizer"].cleanup()
 
+    def _stamp_cached_clip_text(self, meta_info):
+        """Give the mark its text so an interrupted clip can be trimmed. Cached-send branch
+        only: the cache-miss path re-synthesizes live and would stamp every chunk. See INIT.md."""
+        if meta_info.get("message_category") != "static_node" or meta_info.get("is_silence_trigger"):
```
san0808: `CACHED_SINGLE_MARK_CATEGORIES` already names which categories are cached single marks, but this guard repeats it as a literal, so adding a second one would need both edited. Should the helper read the constant instead?

san0808 (approved):

bolna/agent_manager/task_manager.py

PR #1018, opened 2026-09-04
feat(graph): per-node overrides for interruption threshold and routing model

puneet-bolna:
Two per-node overrides on GraphNode, both inheriting the agent-level value when unset, both following the existing repeat_after_silence_seconds pattern:

*Per-node overrides*

• `number_of_words_for_interruption` — overrides `ConversationConfig.number_of_words_for_interruption`
    `None` = inherit  ·  `0` = barge-in off on this node

• `routing_model` — overrides the agent's routing model
    `None` = inherit  ·  runs on the agent's routing provider and credentials

Interruption threshold. task_manager remembers the call default and _apply_node_interruption_threshold(node) re-applies the landed node's value at the same three node-change sites that set repeat_after_silence_seconds. It writes both holders (task_manager's copy at the interim gate, and InterruptionManager's), since the value can now move mid-call. is None, not truthiness. The first node is covered by turn-1's routing_info; barge-in during the welcome is already gated.

Routing model. Scope is deliberately model-only on the agent's provider and creds — no per-node provider or keys (secrets in agent JSON, the whole credential matrix). _init_routing_client was split by pure extraction into _qualify_routing_model / _routing_kwargs_for_model / _routing_llm_for(node); the default LLM is built through the same helpers, so the default path is identical. _routing_llm_for returns the agent LLM by identity without an override, else a bounded per-model cache. Single call site → router chains covered. Telemetry reports the model actually used, with _reset_routing_identity() so a no-call turn reports the agent model as before. routing_info.routing_model is already what the executions view renders.

Safety. 125 existing routing tests pass untouched. No transport per cached instance (route() is non-streaming HTTP). Bounded cache.

Tests. 19 new across two files, all failing at the parent commit. One e2e stub gained the two attributes real __init__ sets. Suite 1491 passed, 1 skipped, ruff clean.


bolna/agent_types/graph_agent.py
```diff
@@ -157,11 +162,80 @@ def _initialize_llm(self):
...
+        kwargs = dict(self._conversation_base_kwargs)
+        kwargs["provider"] = provider
+        if provider != self._conversation_provider:
+            # Another provider cannot authenticate with this one's credentials; let its own
+            # env defaults serve the node instead of sending a key that will be rejected.
+            for key in ("llm_key", "base_url", "api_version"):
+                kwargs.pop(key, None)
+        for key in ("model", "temperature", "max_tokens"):
+            if key in override:
+                kwargs[key] = override[key]
```
san0808, line 276: `use_responses_api` rides in from `_conversation_base_kwargs`, which task_manager derived once from the agent-level model via `RESPONSES_API_MODEL_PREFIXES`. Overriding the model here never revisits it, so a node on gpt-6 under a gpt-4.1 agent gets chat completions with an effort plus transition tools, and a node on gpt-4.1 under a gpt-6 agent inherits Responses mode with its own `previous_response_id` chain. Should this re-derive the transport from the node model?

bolna/agent_types/graph_agent.py
```diff
@@ -1325,7 +1452,8 @@ async def generate(self, message: List[dict], **kwargs) -> AsyncGenerator:
                 tool_choice = self._get_tool_choice_for_node(history=message)
                 forced_name = tool_choice["function"]["name"] if tool_choice else None
                 node_tools = self._tools_for_node(current_node, forced_name)
-                async for chunk in self.llm.generate_stream(
+                node_llm = self._conversation_llm_for(current_node)
```
san0808, line 1561: Now that a turn can run on something other than `self.llm`, should the interruption hooks follow it? `invalidate_response_chain`, `set_interruption_hint` and `cancel_in_flight_response` in task_manager are all hardcoded to `llm_agent.llm`, so a barge-in on an override node cancels an idle client and leaves the override holding a `previous_response_id` for a response nobody consumed.

bolna/agent_types/graph_agent.py
```diff
@@ -157,11 +162,80 @@ def _initialize_llm(self):
...
+        if effort is not None:
+            kwargs["reasoning_effort"] = getattr(effort, "value", effort)
+        # An inherited effort would be rejected by a non-reasoning override model.
+        if not is_reasoning_model(canonical_model(kwargs.get("model") or "")):
+            kwargs.pop("reasoning_effort", None)
+
+        llm = llm_class(**kwargs)
+        if len(self._conversation_llm_cache) >= self._conversation_llm_cache_max_size:
+            self._conversation_llm_cache.pop(next(iter(self._conversation_llm_cache)))
+        self._conversation_llm_cache[cache_key] = llm
```
san0808: Should these get closed on teardown? It only walks `agent.llm`, `conversation_completion_llm` and `voicemail_llm`, and with the Responses API on, each instance holds an `OpenAIWSConnection` that only `close()` disconnects.

bolna/agent_types/graph_agent.py
```diff
@@ -157,11 +162,80 @@ def _initialize_llm(self):
...
+            kwargs["model"] = f"{provider}/{kwargs['model']}"
+
+        effort = override.get("reasoning_effort")
+        if effort is not None:
+            kwargs["reasoning_effort"] = getattr(effort, "value", effort)
+        # An inherited effort would be rejected by a non-reasoning override model.
+        if not is_reasoning_model(canonical_model(kwargs.get("model") or "")):
+            kwargs.pop("reasoning_effort", None)
+
+        llm = llm_class(**kwargs)
```
san0808: `_initialize_llm` falls back to OpenAI when construction throws, but this runs inline in `generate()` with no guard. A node on azure has its `base_url` popped just above, so `azure_endpoint.rstrip()` hits None and the call ends. Should this fall back to `self.llm` instead?

bolna/agent_types/graph_agent.py
```diff
@@ -157,11 +162,80 @@ def _initialize_llm(self):
...
-            return OpenAiLLM(model=self.llm_model or "gpt-4o-mini", llm_key=self.llm_key or os.getenv("OPENAI_API_KEY"))
+            fallback_kwargs = {
+                "model": self.llm_model or "gpt-4o-mini",
+                "llm_key": self.llm_key or os.getenv("OPENAI_API_KEY"),
+            }
+            self._conversation_provider = "openai"
+            self._conversation_base_kwargs = dict(fallback_kwargs)
+            return OpenAiLLM(**fallback_kwargs)
+
+    def _conversation_llm_for(self, node: Optional[dict]):
```
san0808, line 193: Building lazily puts a TLS handshake in the first-token path the first time a call reaches an override node. The nodes are known at graph load, so should these be constructed there rather than in the audio path?

bolna/agent_types/graph_agent.py
```diff
@@ -157,11 +162,80 @@ def _initialize_llm(self):
...
+            logger.warning(f"Unknown provider {provider!r} in node llm_config, using {self._conversation_provider!r}")
+            provider = self._conversation_provider
+
+        kwargs = dict(self._conversation_base_kwargs)
+        kwargs["provider"] = provider
+        if provider != self._conversation_provider:
+            # Another provider cannot authenticate with this one's credentials; let its own
+            # env defaults serve the node instead of sending a key that will be rejected.
+            for key in ("llm_key", "base_url", "api_version"):
+                kwargs.pop(key, None)
```
san0808: Dropping the credentials means the node falls through to the provider env default, so a BYOK customer whose agent runs on their own key gets a node quietly running on the platform key. Should a provider override with no credentials of its own be rejected instead?

bolna/agent_types/graph_agent.py
```diff
@@ -157,11 +162,80 @@ def _initialize_llm(self):
...
+        for key in ("model", "temperature", "max_tokens"):
+            if key in override:
+                kwargs[key] = override[key]
+
+        llm_class = SUPPORTED_LLM_PROVIDERS.get(provider, OpenAiLLM)
+        # LiteLLM addresses a backend by a "provider/model" string; a bare name resolves to OpenAI.
+        if llm_class is LiteLLM and "/" not in (kwargs.get("model") or ""):
```
san0808: This skips the re-prefix whenever the inherited model already has a `/`, so an agent on `groq/llama-3.3-70b-versatile` with a node setting only `provider: together` keeps dispatching to Groq while the Groq key was popped above. Should the bare model name be derived before re-qualifying?

tests/test_node_conversation_llm_override.py
```diff
@@ -0,0 +1,292 @@
...
+def _registry(real=None):
+    def for_provider(name):
+        def make(**kwargs):
+            client = MagicMock()
+            client.captured_kwargs = kwargs
+            client.generate_stream = MagicMock()
+            return client
+
+        return make
+
```
san0808, line 30: The registry is all MagicMock factories, so no real client is ever constructed and the transport choice is invisible here. Worth one case that builds a real `OpenAiLLM` per node and asserts which API it picked?

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

bolna/agent_types/graph_agent.py
puneet-bolna: fixed

san0808 (approved):

PR #1019, opened 2026-09-05
feat: route the gpt-6 family through the reasoning-model path and catalog gpt-6-astra

san0808:
`gpt-6-astra` is live on the OpenAI API but every reasoning branch here tested `model.startswith("gpt-5")`, so a gpt-6 model fell through to the legacy request shape. Verified against the live model, that path is rejected four ways:

- `max_tokens` instead of `max_completion_tokens`
- `stop: ["User:"]`
- any `temperature` other than 1
- function tools on chat completions, at every effort level

The last one has no escape here. gpt-5.4 and up can drop to `reasoning_effort: none` and stay on chat completions; gpt-6-astra has no `none`, so a tool-using agent has to reach the Responses API.

Replaces the six `startswith(GPT5_MODEL_PREFIX)` checks with `is_reasoning_model()` over `REASONING_MODEL_PREFIXES`, so the next generation is one tuple entry. `gpt-6` joins `RESPONSES_API_MODEL_PREFIXES`.

Catalog entry is `[low, medium, high, xhigh, max]`, measured against the API. gpt-6 drops the `none` and `minimal` floors, so `default_reasoning_effort` correctly returns `low`, and it adds `max`, which needed a new `ReasoningEffort` member.


prabalrastogi (approved):
LGTM

PR #1021, opened 2026-09-05
fix(synthesizer): record turn latency for the non-streaming HTTP path

san0808:
The one-shot HTTP synth loop (`_generate_http_loop`) never recorded into `turn_latencies`, so HTTP-only providers (gemini, openai) reported empty `synthesizer_latencies` in the execution telemetry while the streaming synths populated theirs.

Records a per-turn entry shaped like the streaming path's, measuring the render time. A non-streaming render produces the whole utterance at once, so `first_result_latency_ms` equals `total_stream_duration_ms`. Now HTTP-only TTS latency shows up in the usage/latency dashboards like every other provider.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.230"
+__version__ = "0.10.231"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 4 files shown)


san0808:
Closing: gemini is moving to the streaming path (streamGenerateContent), which records turn latency natively via StreamSynthesizer, so the HTTP-loop instrumentation isn't needed for it. Easy to reopen if the streaming switch doesn't ship (non-streaming openai/gemini would still benefit).

PR #1028, opened 2026-09-07
fix: don't show the injected silence nudge as a caller turn

puneet-bolna:
Problem

user: [silence] User was silent for 12.0 seconds
assistant: Sure, take your time. Tell me when you are ready.

The caller never said that line, and the recording has no matching 12-second silence.

Cause

It's an internal control token. When repeat_after_silence_seconds elapses, task_manager calls

await self._inject_and_run_llm(f"[silence] User was silent for {self.repeat_after_silence_seconds} seconds")

and that helper did conversation_history.append_user(...) — a real user row. That history is what format_messages renders into executions.transcript and what the progression tab renders, so the nudge surfaces as caller speech.

The number is wrong too: it interpolates self.repeat_after_silence_seconds, the configured threshold, never a measurement. On this call the real gaps were 16.87 s, 33.83 s and 15.52 s (from __check_for_completion at 13:28:40 / 13:28:57 / 13:29:58) — never 12.0 s. The agent was also speaking through those gaps, so the audio isn't silent there either.

Change

New exclude_from_transcript message key — the mirror of the existing exclude_from_llm, which keeps a row in the transcript but out of the LLM's view. This one does the inverse:

- added to INTERNAL_MESSAGE_KEYS, so strip_internal_keys drops it before any provider call
- format_messages skips flagged rows
- _inject_and_run_llm sets it

Re-prompt behaviour is unchanged. The conversation LLM reads conversation_history.get_copy(), which filters only exclude_from_llm, so it still sees the nudge and still re-prompts. Only the rendering changes.

Also silenced by the same filter, all reasonable: extraction/summarization input (:2704), one trace log line (:3199), and the check_for_completion prompt (:4142) — a synthetic line shouldn't feed extraction or bias the hangup decision. The conversation path doesn't use format_messages.

Tests — tests/test_silence_marker_hidden_from_transcript.py (5): absent from the transcript; still present for the LLM; the key never reaches a provider; an ordinary user turn mentioning silence is untouched; and the injection marks the turn. 3 of 5 fail at HEAD. Suite 1533 passed, 1 skipped, ruff clean.

### DIFF
--- a/bolna/agent_manager/task_manager.py
+++ b/bolna/agent_manager/task_manager.py
@@ -7377,7 +7377,9 @@ class TaskManager(BaseManager):
             logger.error(f"Error in processing message output: {str(e)}")
 
     async def _inject_and_run_llm(self, injected_message: str):
-        self.conversation_history.append_user(injected_message)
+        # exclude_from_transcript, not exclude_from_llm: the nudge is the whole point for the
+        # LLM, but the caller never spoke it, so it must not surface as a user turn.
+        self.conversation_history.append_user(injected_message, exclude_from_transcript=True)
         meta_info = self.__get_updated_meta_info(
             {
                 "io": self.tools["output"].get_provider(),
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.235"
+__version__ = "0.10.236"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/helpers/conversation_history.py
+++ b/bolna/helpers/conversation_history.py
@@ -148,8 +148,17 @@ class ConversationHistory:
         return popped
 
     def pop_and_merge_user(self, new_content: str) -> str:
+        """Fold a split-utterance tail back into the caller turn it belongs to.
+
+        A row flagged exclude_from_transcript is an engine control token (the silence nudge), not
+        caller speech. Merging it would paste the marker into a real utterance, where it surfaces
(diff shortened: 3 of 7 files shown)


bolna/agent_manager/task_manager.py
```diff
@@ -7295,7 +7295,9 @@ async def __process_output_loop(self):
             logger.error(f"Error in processing message output: {str(e)}")
 
     async def _inject_and_run_llm(self, injected_message: str):
-        self.conversation_history.append_user(injected_message)
+        # exclude_from_transcript, not exclude_from_llm: the nudge is the whole point for the
+        # LLM, but the caller never spoke it, so it must not surface as a user turn.
+        self.conversation_history.append_user(injected_message, exclude_from_transcript=True)
```
san0808, line 7382: Should `pop_and_merge_user` be taught about this flag too? This injection sets `response_in_pipeline = True`, so a caller final that lands while the nudge is still generating takes the `overlapped` branch in `_handle_transcriber_output`, which pops this row, concatenates its content and re-appends with only `asr_turn_id`. The flag is gone, so the transcript shows `user: [silence] User was silent for 12.0 seconds Hello.` and the real caller turn carries the marker into the LLM context as well. That merge branch fires often in practice, just never on the injected row so far. Discarding a flagged row's content instead of merging it would cover both.

puneet-bolna: done

san0808 (approved):
lgtm

PR #1032, opened 2026-09-08
fix: infer bool as bool, not int, in infer_type()

Arunendra21:
## What

`infer_type()` in `bolna/helpers/utils.py` maps a JSON value to a Pydantic field type for `json_to_pydantic_schema()`. It checks `int` before `bool`:

```python
def infer_type(value):
    if isinstance(value, int):
        return (int, ...)
    ...
    elif isinstance(value, bool):   # unreachable
        return (bool, ...)
```

In Python, `bool` is a subclass of `int`, so `isinstance(True, int)` is `True`. The `bool` branch is therefore **unreachable**, and every boolean value is inferred as `int`:

```python
infer_type(True)   # -> (int, ...)   expected (bool, ...)
infer_type(False)  # -> (int, ...)
```

Because `infer_type` feeds `json_to_pydantic_schema()`, a boolean field in the JSON is described as an integer in the generated Pydantic schema.

## Fix

Check `bool` before `int` — the standard ordering when an isinstance chain needs to tell the two apart:

```python
if isinstance(value, bool):
    return (bool, ...)
elif isinstance(value, int):
    return (int, ...)
...
```

Ints, floats, strings, lists and dicts infer exactly as before; only booleans change (from `int` to the correct `bool`).

## Tests

Added `tests/test_infer_type.py` covering booleans (the regression), ints, and the other scalar/container types.

Before: `infer_type(True) == (int, ...)`
After: `infer_type(True) == (bool, ...)`

### DIFF
--- a/bolna/helpers/utils.py
+++ b/bolna/helpers/utils.py
@@ -513,12 +513,15 @@ def has_placeholders(s):
 
 
 def infer_type(value):
-    if isinstance(value, int):
+    # bool must be checked before int: in Python bool is a subclass of int, so
+    # isinstance(True, int) is True. Ordering int first made the bool branch
+    # unreachable and inferred every boolean as int.
+    if isinstance(value, bool):
+        return (bool, ...)
+    elif isinstance(value, int):
         return (int, ...)
     elif isinstance(value, float):
         return (float, ...)
-    elif isinstance(value, bool):
-        return (bool, ...)
     elif isinstance(value, list):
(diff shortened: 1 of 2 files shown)


Arunendra21:
Friendly ping on this one — it's a small, self-contained bug fix and has been open for a few weeks. Happy to rebase or make any changes if a maintainer has a chance to take a look. Thanks!

PR #1034, opened 2026-09-09
fix: release the audio gate when the ASR stops producing new words [B…

puneet-bolna:
The agent went silent after the caller said हां जी and the call was abandoned. The reply was generated (2.66s) and synthesized (113ms) — but never played.

The caller started saying "hello" while the reply was still generating, so the audio gate held it (Audio status=WAIT - user is speaking) for 2.42s. When Deepgram finally finalized "hello", the overlap path merged the utterances and discarded the reply. sync_history logged marks=0, response_heard len=0 — zero frames reached the caller.

Root cause

The gate has an escape that releases held audio once the caller goes stale, but it could never fire. note_user_liveness() refreshed last_interim_ts on every interim, and Deepgram re-emitted an identical non-final hello every ~0.6s, so staleness kept resetting.

A duplicate interim means the ASR has nothing new — not that the caller is still talking.

Fix

Refresh liveness only when the transcript actually changes:

def note_user_liveness(self, transcript: str = "") -> None:
    if transcript and transcript == self.last_interim_text:
        return
    self.last_interim_text = transcript
    self.last_interim_ts = time.time()

STUCK_AUDIO_GATE_RELEASE_S 3.0 → 2.0. Needed, not cosmetic: with content-aware liveness the anchor sits at the first hello and Deepgram finalized 2.601s later, so 3.0s still missed by ~0.4s. 2.0s releases with 0.6s margin and stays above Deepgram's 1s utterance_end_ms floor, so healthy turns are unaffected.

This doesn't talk over a real speaker — new words reset the timer, so only a frozen transcript trips it. And once audio plays, the short hello final is correctly classified a false interruption and ignored, so the reply completes.

Tests

tests/test_stuck_audio_gate_release.py (9 cases): replay of the incident's real timings asserting release lands before the ASR final; the inverse guard (new words every 0.5s for 10s never goes stale); silent-caller case still works; back-compat for the no-arg call.

Full unit suite: 1561 passed, 1 skipped. ruff clean.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.234"
+__version__ = "0.10.235"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/agent_manager/interruption_manager.py
+++ b/bolna/agent_manager/interruption_manager.py
@@ -33,6 +33,7 @@ class InterruptionManager:
         self.incremental_delay: int = incremental_delay
         self.utterance_end_time: float = -1
         self.last_interim_ts: float = -1
+        self.last_interim_text: str = ""
 
         # Configuration
         self.number_of_words_for_interruption: int = number_of_words_for_interruption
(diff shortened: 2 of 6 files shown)


san0808 (approved):
lgtm

PR #1035, opened 2026-09-09
fix: don't regenerate a turn that has already been spoken [BOLNA-2621]

puneet-bolna:
Reported by Spinny (9b14c2b2-a997-4d67-9231-0ec999323607). The opening line after the welcome was spoken twice, back to back, with no interrupting user turn — plus a mismatch between the old and new transcript.

The caller's utterance arrived as two finals (हेलो।, then हेलो। हाँ, बताइए।). Reply #1 was generated, committed and marked at 12:08:22.068–.069. The regen debounce was armed at 22.070 — 2 ms later — fired at 22.774, and produced a near-identical second reply. Both were acked; the caller heard both. The transcript mismatch is the same cause: two near-duplicate assistant turns genuinely got committed to history.

Root cause

regen-settle exists so a late final can replace an in-flight turn rather than stack on it. But __regen_after_settle called kickoff_llm_generation unconditionally after the 0.7 s window, with no check on whether the turn it meant to replace had already been spoken. When it had, the regen replaced nothing and simply appended.

Regression from d43ba2f5 / PR #967 (0.10.202, 21 Aug), which introduced the debounce.

Fix

Record at episode start which sequence the debounce means to replace, and skip the regen if that sequence's audio has since shipped:

if superseded is not None and superseded in self._sent_audio_sequences:
    logger.info("... regen_settle skipped ... reason=already_spoken")
    return

Two things this gets right that the obvious version doesn't:

- Fire time, not arm time. When the second final is handled the first reply hasn't shipped yet — the incident logs audio_playing=False at 22.064, 4 ms before the send — so an arm-time check misses it. At fire time it's deterministic.
- Anchored at episode start, not seq - 1. A burst of finals re-arms with a fresh sequence each time, so seq - 1 would test a turn that was never spoken and let the duplicate through.

_sent_audio_sequences already exists and is populated in the output loop's SEND path, so no new state machinery. When the superseded reply genuinely hasn't shipped, the predicate is false and behaviour is unchanged.

### DIFF
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.233"
+__version__ = "0.10.234"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.236"
+__version__ = "0.10.237"
 
 import os
 from bolna.helpers.logger_config import configure_logger
--- a/bolna/__init__.py
+++ b/bolna/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.10.234"
+__version__ = "0.10.235"
 
 import os
 from bolna.helpers.logger_config import configure_logger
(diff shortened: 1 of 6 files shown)


san0808 (approved):
lgtm

PR #1041, opened 2026-09-15
fix: keep a refused Azure ASR socket from hanging up the call

puneet-bolna:
- Problem — the reference call timeline, plus the evidence the trigger was upstream (Deepgram + Sarvam: 0 connect failures in the same window while Deepgram was under heavy load; Azure TTS: 0).
- Root cause — is_active_transcriber_alive() probing transcription_task, which Azure never sets, so a standby death read as the active one dying. Plus the separate fact that single-language agents held a bare transcriber and died on the first refused socket.
- Fix — the two is_connected() definitions inline, reconnect_single_transcriber(), and the two changes that came along because they had to (cancellation logging, pump cancellation) with the reason each was unavoidable.
- Scale — 161 calls killed in the 6h window 2026-09-11 12:00–18:00 UTC. I queried that fresh for this — it's calls, not log lines (one terminal exception per dead call, 36.5M lines scanned), since a bare error-line count would have overstated it ~6×.
- Accepted residual — all five things still unfixed, stated plainly: no verification, no backoff, the SDK-thread race, frame counters, no second region. It says explicitly that a sustained refusal still ends the call.


bolna/agent_manager/task_manager.py
```diff
@@ -5003,6 +4965,28 @@ async def _end_call_on_component_error(self, error, hangup_detail):
             self.hangup_detail = hangup_detail
             await self.__process_end_of_conversation()
 
+    # Same budget as TranscriberPool._MAX_RECONNECTS_PER_CALL, for agents that hold a bare
+    # transcriber instead of a pool — those had no recovery at all, so one refused socket
+    # ended the call (every Azure agent during the 2026-09 centralindia outage).
+    MAX_SINGLE_TRANSCRIBER_RECONNECTS = 5
+
+    async def reconnect_single_transcriber(self):
```
SubhrajitPrusty: During a reconnect the connection_error is not cleared?

bolna/transcriber/transcriber_pool.py
```diff
@@ -184,8 +184,11 @@ def labels(self):
...
 
     def is_active_transcriber_alive(self):
-        """True if the active transcriber's connection task is still running."""
+        """True if the active transcriber's provider connection is still usable."""
         active = self.transcribers[self.active_label]
+        probe = getattr(active, "is_connected", None)
+        if probe is not None:
+            return probe()
         task = getattr(active, "transcription_task", None)
         return task is not None and not task.done()
```
SubhrajitPrusty: isnt is_connected and the code below here doing the same check? ie. looking for the transcription_task?

bolna/transcriber/azure_transcriber.py
```diff
@@ -82,6 +88,18 @@ async def run(self):
             meta["connection_error"] = self.connection_error
             await self.transcriber_output_queue.put(create_ws_data_packet("transcriber_connection_closed", meta))
 
+    async def cancel_audio_pump(self):
+        """Stop the pump feeding the previous connection, if one is still running."""
+        task = self.send_audio_to_transcriber_task
```
SubhrajitPrusty, line 93: what is the difference between this send_audio_to_transcriber_task and a transcription_task that is not being used by this class?

bolna/agent_manager/task_manager.py
puneet-bolna: cleared now

bolna/transcriber/transcriber_pool.py
puneet-bolna: removed it

san0808 (approved):

PR #1043, opened 2026-09-15
fix: stamp the LID switch trace rows at send and reply time

puneet-bolna:
Problem — the 37µs pair from run 7a07ae20, and why both readings (looks late, looks instant) are wrong.
- Root cause — _log_decision() writing both rows after the reply, quoting the contract it broke (utils.py:957: callers that log after the fact must pass ts). Also notes self.latency_ms was being measured correctly all along but only reached the app log.
- Fix — the before/after code, why _log_request returns the shared meta_info, and the seconds unit matching utils.py:1049. Plus the two gaps found while wiring it: no response row on the failure/decline paths, and a throttled judge being indistinguishable from a valid decline because _hedged_generate swallows the exception.
- Accepted residual — states plainly that this fixes reporting, not latency: ~1.36s end-to-end, 300ms of it the deliberate settle, and the caller still heard English → handoff → Hindi. Points at the loop-blocking write_request_logs as a likely contributor, with the existing doc reference.
- Tests — including why the four stub renames mattered (they'd have passed as dead assignments while exercising real log I/O).


bolna/helpers/language_switcher.py
```diff
@@ -238,10 +244,12 @@ async def decide(
             logger.info(
                 f"LanguageSwitcher decision: {result} (latency_ms={self.latency_ms:.0f}, hedge_won={self.hedge_won})"
             )
-            self._log_decision(detector_transcript, result)
+            self._log_response(result, log_meta, start_time)
             return result
         except Exception as e:
```
pavankapoor31, line 239: The request row is now written before the await, but a cancelled decide() raises CancelledError, which `except Exception` does not catch, so a timed-out decide leaves a request row with no reply.

pavankapoor31 (approved):

bolna/helpers/language_switcher.py
```diff
@@ -238,10 +244,12 @@ async def decide(
...
             logger.error(f"LanguageSwitcher decision error: {e}")
+            # Without this a failed decision left a request row with no reply in the trace.
+            self._log_response({"error": str(e)}, log_meta, start_time)
```
san0808, line 241: `wait_for` cancels `decide()` at 6.0s (`LANGUAGE_SWITCH_DECIDE_TIMEOUT_S`, and `_hedged_generate`'s docstring notes an observed 5.9s tail), and `CancelledError` is a `BaseException` so `except Exception` misses it, leaving a request row with no reply on exactly the slow decides this is meant to catch. Should we also catch `asyncio.CancelledError`, log a cancelled response and re-raise?

bolna/helpers/language_switcher.py
```diff
@@ -238,10 +244,12 @@ async def decide(
             logger.info(
                 f"LanguageSwitcher decision: {result} (latency_ms={self.latency_ms:.0f}, hedge_won={self.hedge_won})"
             )
-            self._log_decision(detector_transcript, result)
+            self._log_response(result, log_meta, start_time)
```
san0808: If `_log_response` itself raises here (say `last_usage` is not a dict and `usage.get` throws), we fall into `except Exception` below and write a second response row while dropping a decision the model already returned. Should the `try` wrap just the `await`?

bolna/helpers/language_switcher.py
```diff
@@ -224,10 +224,16 @@ async def decide(
...
+        # Logged before the await so the row carries the send time, not the reply time.
+        log_meta = self._log_request(detector_transcript, start_time)
         try:
             result = await self._hedged_generate(messages)
             if result is None:
-                # A parsed `null` is the model validly declining, not a broken judge.
+                # A parsed `null` is the model validly declining, not a broken judge — say which,
+                # or a throttled judge looks identical to a decline in the trace.
+                self._log_response(
+                    {"error": "generate_failed"} if self.last_generate_errored else None, log_meta, start_time
```
san0808: `write_request_logs` normalizes `message=None` to `""`, so a valid decline lands as a blank Data cell that reads like a truncated row. Should this send something explicit like `{"target_language": None}` instead?

tests/test_language_switch_log_timestamps.py
```diff
@@ -0,0 +1,67 @@
+"""The LID trace rows are stamped when each event happened, not both at reply time.
+
+_log_decision() used to write the request and response rows together after the model replied, so
+both carried the completion timestamp: in the trace the pair landed after the main LLM's response
+and showed ~0ms apart (run 7a07ae20: request 26.609920, response 26.609957 — 37 microseconds).
```
san0808: Can we drop the run id before merge since this repo is public? The "37 microseconds apart, both at reply time" detail carries the symptom on its own.

puneet-bolna: done

bolna/helpers/language_switcher.py
puneet-bolna: done

bolna/helpers/language_switcher.py
puneet-bolna: done

bolna/helpers/language_switcher.py
puneet-bolna: done

san0808 (approved):
lgtm

PR #1045, opened 2026-09-16
feat: LLM request-id logging + empty-response observability on the chat-completions path

san0808:
Two small observability changes on the LLM path.

**1. Log the provider request/response id once per turn, for every provider.**
`BaseLLM._log_llm_request_id` reads the HTTP request-id header (`x-request-id` / `apim-request-id` / `x-ms-request-id`) when the stream exposes it, and always falls back to the response object id (`chatcmpl-*` on chat completions, `resp_*` on the Responses API). One helper on the base class, called at the six streaming sites (OpenAI chat, Azure chat, LiteLLM, Gemini, Responses HTTP, Responses WS). Custom / self-hosted OpenAI-compatible endpoints get the response id. Defensive `getattr` throughout, so it can never raise inside a turn; nothing is logged when no id is available.

**2. Record an `empty_response` non-fatal event when a turn returns no output on the chat-completions path.**
Previously only the Responses API recorded an empty turn (`incomplete_empty_response`); chat-completions empties were logged but left no structured event, so they were invisible to downstream observability. The event is appended in the single place empty output is detected, only when nothing else already explained the empty turn, so the Responses API path is not double-counted.


bolna/llms/llm.py
```diff
@@ -46,3 +46,15 @@ async def close(self):
         No-op by default. Override in subclasses that hold closeable resources.
         """
         pass
+
+    def _log_llm_request_id(self, stream=None, response_id=None):
+        """Log the provider's request/response id once per turn so it is greppable for support."""
+        request_id = None
```
vikramsingh-bolna: can we wrap this in try exception rest LGTM

vikramsingh-bolna (approved):

bolna/llms/llm.py
san0808: Done, wrapped in try/except so logging can never break a turn.

PR #1046, opened 2026-09-17
perf: stream the graph routing call and dispatch on the function name (BOLNA-2744)

san0808:
The graph routing hop runs serially in front of the answering LLM, so what matters is time to decision, not time to last token. It was issued as a non-streaming completion, so it paid for the whole response — including `reasoning` and `confidence`, which the model emits *after* the function name they describe and which nothing reads to make the routing decision.

Under `tool_choice: "required"` the function name is streamed before any argument. This streams the routing call and resolves as soon as the function name and the edge's own parameters have arrived, then drains the remainder in a background task that folds `reasoning`, `confidence` and the usage record back into `routing_latencies[]` before the call ends.

#972 had made `reasoning` opt-in behind `GRAPH_ROUTER_RATIONALE`, off by default, to avoid paying for those tokens. That also removed the rationale that a misrouting agent is debugged from, and dropping it measurably shifts routing decisions. The flag is gone and the tool schema is byte-identical to the pre-#972 one, so routing behaviour is that schema's rather than something new to re-establish.

Reordering the schema so the rationale comes last was tried first and rejected: on a replay of recorded routing requests it moved decisions materially more than run-to-run variance, so property order is a perturbation in its own right. Cutting the stream instead leaves the request untouched.

### Verification

`RoutingStreamReader` is unit-tested for early resolution, out-of-order arguments, a rationale-free response, malformed JSON and the no-tool-call case.

Beyond that, recorded routing requests were replayed against the model with `stream=True` and the chunks fed through the reader, comparing the early decision against a full parse of the same response:

| corpus | n | decided early | decision preserved | key order held |
|---|---|---|---|---|
| in-distribution | 2146 | 100% | 2146/2146 | 2146/2146 |
| unseen agents | 1227 | 100% | 1227/1227 | 1227/1227 |

No response put `reasoning` or `confidence` ahead of an edge parameter. The decision resolved at 5 streamed chunks at both p50 and p90, against a 44-token full completion. Where the cut would leave a required edge parameter missing the reader falls back to parsing the whole response, so an out-of-order model degrades to the previous behaviour rather than to a wrong decision.

### Review notes

`reasoning` and `confidence` are absent from `routing_info` when the hop is yielded and are patched into the `routing_latencies[]` entry by the tail task. Outstanding tails are settled just before the latency snapshot is taken, next to the existing voicemail-check cancellation, so a hangup within a few hundred ms of the last routing hop does not lose its rationale or its token accounting; a tail that has not landed within `ROUTING_TAIL_SETTLE_TIMEOUT` is cancelled rather than delaying the hangup. The rationale is written as its own `GRAPH_ROUTING` row stamped at the hop start rather than appended to the routing response line.

Azure token metering for the routing hop moves into the tail, since usage only arrives in the final chunk.

Gemini routing stays non-streaming; its `route()` builds a result from a different SDK shape.


puneet-bolna:
lgtm

puneet-bolna (approved):

PR #1047, opened 2026-09-17
fix: book each send for the audio it actually carries on sip-trunk and web calls (BOLNA-2726)

amansrivstv:
Reported ASR latencies ramp with call elapsed time instead of measuring anything. Measured drift slope in prod: **0.61 on sip-trunk** (p50 24.8s, max 221s, 95.8% of interims above 2s) and **0.64 on web calls** (p50 33.1s, max 205s, 87.6% above 2s). The calls themselves are healthy — LLM TTFT 716ms, TTS first result 201ms, ASR connect 22ms.

Turning a transcript position into a latency means looking up when the audio at that position was sent, so the position-to-send-time map is only as good as the duration each websocket send is booked for. Every send is booked at a fixed `audio_frame_duration`, but two families of input handler batch something else:

| path | handler batches | booked at |
| --- | --- | --- |
| sip-trunk | `AUDIO_BATCH_MS = 80` (4 x 20ms ulaw = 640B) | 0.200 |
| web_based_call | `INGEST_CHUNK_BYTES = 6400` (linear16@16k = 200ms) | 0.256 |
| freeswitch | same 6400B = 200ms | 0.500 |

So the cursor runs fast and every lookup resolves to a packet from earlier in the call. The same constant places `connection_start_time`, which feeds `user_stop_ts_wall` and so every `user_bot_latencies` entry — sip-trunk's user-bot latency p50 is 5,840ms against 1,432-2,782ms on other providers.

`MEASURED_FRAME_PROVIDERS` is the single home for the paths whose handler batches something other than the constant it is booked at; `_audio_frame_seconds()` measures the payload for those and returns the legacy constant for everything else.

Nothing else moves. Simulated old against new per provider on both deepgram and soniox:

| provider | old | new |
| --- | --- | --- |
| twilio / plivo / exotel / vobiz | 0.200 | 0.200 |
| playground | 0.000 | 0.000 |
| sip-trunk | 0.200 | **0.080** |
| web_based_call | 0.256 | **0.200** |
| freeswitch | 0.500 | **0.200** |

Verified end to end on staging: interim latencies went from 2,097 -> 20,659 -> 21,860ms across a 43s call on the stock build, to a flat 17-47ms across an 87s call with this change.

On deploy, `voiceai.user_bot.latency_ms` steps down for sip-trunk and web calls as the metric becomes correct.


san0808 (approved):
lgtm

PR #1048, opened 2026-09-18
fix: keep credentials out of the logs and cut ws-server log volume by about a third (BOLNA-2415)

san0808:
Several log lines interpolated whole tool and LLM config objects, so fields like the tool's `api_token`, its auth `headers` and the LLM `api_key` ended up in the log message along with full conversation history. They now log identifying fields only.

The rest is log volume. The mark trail logged the same `(mark_id, seq, turn_id, response_uid)` tuple at three or four layers, so the send-side line moves into `update_data` (which also gives web, SIP and FreeSWITCH calls coverage they never had) and the duplicates go. `update_is_audio_being_played` now records which of its 16 call sites fired instead of a bare True/False. Agent-config and per-chunk payload dumps are trimmed to ids and shape.

About 23% fewer lines and 35% fewer bytes from ws-server.


SubhrajitPrusty (approved):
LGTM

PR #1049, opened 2026-09-18
fix(graph-agent): log the model that served each turn, not the base config

puneet-bolna:
## Context
Graph agents can override the LLM per node.a `negotiate`
node ran on `azure/gpt-5.4`, but every conversation turn was logged, metered, and
displayed as the agent base model `azure/gpt-4.1-mini`. Root cause: the conversation-turn
log/latency producers stamped `self.llm_config["model"]` (base config) instead of the
client that actually served the turn. Routing already recorded the override correctly.

Since `azure/gpt-5.4` costs ~6× (input) / ~9× (output) of `gpt-4.1-mini`, this
under-billed every override turn and hid the real model in logs and the progression trace.

## Change
- Add `TaskManager._active_llm_model()` — resolves the active client via the existing
  `_active_llm_client()` → graph agent's `current_conversation_llm()`, prefers
  `.request_log_model` then `.model`, and falls back to `self.llm_config["model"]` for
  non-graph agents / no active client (behaviour-preserving there).
- Use it at the six graph-relevant producers: RESPONSE `__store_into_history` and the
  non-stream RESPONSE (billing), the graph REQUEST log (display), and the two latency
  stamps `_stamp_llm_latency_dict` / `_append_eager_llm_stub` (progression trace).
- No schema change: `latency_dict` already carried per-turn `model`; only the value was wrong.

## Tests
- `tests/test_active_llm_model.py` (5) — override client reported, `request_log_model`
  preferred, `.model` fallback, base-config fallback when no client/labels.
- Existing graph/routing suites green (52). ruff check + format clean.


san0808 (approved):
lgtm

bolna/agent_manager/task_manager.py
```diff
@@ -4200,7 +4129,7 @@ def _append_eager_llm_stub(self, meta_info):
                 "sequence_id": meta_info.get("sequence_id"),
                 "turn_id": meta_info.get("turn_id"),
                 "asr_turn_id": getattr(_t_s, "turn_counter", None),
-                "model": self.llm_config.get("model") if self.llm_config else None,
+                "model": self._active_llm_model(),
```
san0808, line 4213: `_append_eager_llm_stub` runs at the top of `_process_conversation_task`, before `generate()` calls `_conversation_llm_for()`, so `current_conversation_llm()` still holds the previous turn's client and the stub gets that model. The completion stamp only replaces the stub on a turn that finishes with the same sequence_id, which leaves the stale model exactly on the hung and barged-in turns the stub exists for. Should the stub leave `model` null and let `_stamp_llm_latency_dict` fill it?

bolna/agent_manager/task_manager.py
```diff
@@ -1238,6 +1233,16 @@ def _active_llm_client(self):
...
 
+    def _active_llm_model(self):
+        """Model label of the client serving this turn. A per-node override (graph agent) runs on a
+        different model than the base config, so logs/billing must record the client's own model,
+        not self.llm_config. Falls back to the base config for non-graph agents or a missing client."""
+        base = self.llm_config.get("model") if self.llm_config else None
+        client = self._active_llm_client()
+        if client is None:
+            return base
+        return getattr(client, "request_log_model", None) or getattr(client, "model", None) or base
```
san0808, line 1249: `GeminiLLM` extends `BaseLLM` so it has no `request_log_model`, and its `__init__` rewrites `self.model` to strip a `provider/` prefix and a leading `models/`. A google-provider agent configured with a prefixed model id would have every row relabelled to the stripped value, which is a new bucket key that misses `bolna_cp["llm"]` on the backend and prices at zero. Should `GeminiLLM` set `_request_log_model = model` the way `AzureLLM` does?

PR #1052, opened 2026-09-18
 fix(synthesizer): Sarvam receiver survives a mid-call socket drop in…

puneet-bolna:
…stead of muting the call## Problem

`SarvamSynthesizer.receiver()` did `break` on `ConnectionClosed`. `SynthesizerPool._run_generate`
iterates `generate()` exactly once, so that ended the audio path for the **rest of the call** while
`monitor_connection` kept redialling a socket nobody read — silently. No error, no
`_end_call_on_component_error`, no hangup.

Prod run `03d94c42` (2026-09-04, Plivo outbound): the Sarvam WS went half-open during a 16s gap
between turns. The user's turn at 2:55 was answered by the LLM in 620ms, the text was sent into the
dead socket, and the caller then heard nothing for the remaining 56s while the agent kept generating
replies into a void. The trace showed synthesizer REQUEST rows with no `tts_start` and no
`agent_speech_start`, because the REQUEST row is written at push time and the RESPONSE row only when
real audio arrives.

## Fix

Three changes in `receiver()`, plus one shared helper:

1. **`ConnectionClosed` keeps looping rather than breaking.** Returns only on `conversation_ended` /
   `connection_error`, so the loud give-up path (raise → `SynthesizerError` → hangup) stays
   reachable. Matches `kalpa_synthesizer` and elevenlabs v3, which already carry this fix and
   document the same `generate()`-iterated-once reasoning.

2. **Settles the in-flight turn with `b"\x00"` on disconnect** — in both the `ConnectionClosed`
   handler and the top-of-loop "not connected" branch, since a socket dying *between* recvs surfaces
   there instead. Guarded by the new `StreamSynthesizer.has_unsettled_turn()` so an idle socket drop
   emits no spurious eos.

   This half is not optional. Without it the lost turn never emits `b"\x00"`, so
   `_record_turn_latency()` never runs and `current_turn_start_time` stays latched — which disables
   the whole new-turn block in `_stamp_turn_start`, *including the stale `text_queue` drop*. A revived
   receiver alone would then hand the next turn's first audio chunk a dead `sequence_id` and
   `__listen_synthesizer` would filter it out. Still mute.

3. **Pins `ws = self.websocket` per iteration**, dropping frames from a replaced socket. A closing
   socket drains buffered frames before raising, and post-fix those could otherwise play as the next
   turn — a regression the `break` used to mask. `kalpa` guards this the same way.

## Scope

Reconnect path only. Interruption behaviour (`number_of_words_for_interruption`, the `overlapped`
drop policy, the audio gate) is deliberately untouched — those amplified the silence in `03d94c42`
but did not cause it.

Seq 21 (the 2:55 turn) is still lost; its text is not replayed. But the socket redialled at
12:36:40.954 would now be read, so seq 22 at 12:36:43 would have been spoken. Recovery moves from
*never* to *next user turn*.

## Testing

`tests/test_sarvam_receiver_reconnect.py` — 10 cases: resume-after-drop (replaying the incident's
shape), no-sentinel-when-idle, drop-between-recvs, the replaced-socket guard, both give-up paths, and
turn-state unlatching end-to-end through the real `_generate_ws_loop`.


SubhrajitPrusty (approved):

san0808 (approved):
lgtm

PR #1053, opened 2026-09-18
feat: AssemblyAI universal-3.x streaming models and ASR keyword/context biasing

san0808:
AssemblyAI shipped `universal-3-6-pro` without announcing it. This transcriber never sent `speech_model`, so every call ran on the server default `universal-streaming-english` and the configured model was ignored.

### Models

Send `speech_model`, aliasing the legacy catalog values to the model the API picks when the parameter is absent, so existing agents are unchanged. Model-specific params are scoped, verified against the live socket rather than the docs, which stop at 3.5:

- `prompt` is universal-3-x only; a legacy model rejects the session when it is set.
- `format_turns` is legacy-only.
- `language_codes` takes bare codes from a fixed set, so the region is stripped (`en-IN` would reject the session) and the param is skipped for a language outside that set.

### Two crashes fixed

**Every interim was emitted as a final.** `turn_is_formatted` is true on every Turn message including partials, so keying finality off it made the agent answer fragments. The signal is `end_of_turn`. A live call now gives 10 interims and 1 final where all 11 were finals.

**A call could die right after the welcome message.** The receiver forwarded a session event carrying `self.meta_info`, which `BaseTranscriber` initialises to `None`. When `Begin` beats the first audio frame, `_set_call_details` raises `TypeError` on the None and the conversation ends. Nothing consumes the event, so it is no longer forwarded. This is a race, so it reproduces intermittently and predates this branch.

### Keyword and context biasing

`bolna/helpers/asr_keywords.py` parses the shared `keywords` field once. Deepgram, ElevenLabs, Gladia and Soniox drop their copies of the same split.

Newly forwarding keywords: `assembly` (`keyterms_prompt`), `azure` (`PhraseListGrammar`), `smallest` (`keywords`, weights preserved). New `transcriber.context` field for engines taking free-form biasing text: `assembly` (`prompt`), `soniox` (`context.text`), `sarvam` (`prompt`). Gemini's Live API has no biasing key.

Each transcriber clamps to its own provider's rules at the point of use, so a multilingual leg on an unweighted engine strips weights a different top-level provider was allowed to keep.

Sarvam's socket URL was hand-joined without encoding, which a free-form prompt would corrupt; it now uses `urlencode`.

### Call behaviour

New parameters are set at connection time, so first-audio latency and barge-in are unchanged. The finality fix changes what the caller hears on AssemblyAI agents: the agent waits for the turn to end instead of answering fragments. On a language the model has no code for, the agent connects and auto-detects rather than failing the call.

Verified on a dev box across ten agents covering every transcriber touched plus an untouched Deepgram baseline: all ten completed with the agent replying.


bolna/transcriber/assemblyai_transcriber.py
```diff
@@ -13,13 +13,24 @@
...
+from bolna.helpers.asr_keywords import keyword_terms
 from bolna.helpers.logger_config import configure_logger
 from bolna.helpers.ssl_context import get_ssl_context
 from bolna.helpers.utils import create_ws_data_packet, timestamp_ms
 
 logger = configure_logger(__name__)
 load_dotenv()
 
+# Catalog rows predating `speech_model`, mapped to the model the API defaults to.
+LEGACY_SPEECH_MODEL_ALIASES = {
```
prateeksachan, line 26: are we keeping the legacy ones?

bolna/transcriber/sarvam_transcriber.py
```diff
@@ -59,6 +61,8 @@ def __init__(
         self.high_vad_sensitivity = high_vad_sensitivity
         self.vad_signals = vad_signals
         self.disable_sdk = disable_sdk
+        # Sarvam biases on free-form text; it takes no term list on the socket.
```
prateeksachan, line 64: I believe sarvam's latest model support keyterms.

> documentation for keyterm promoting: https://docs.sarvam.ai/api-reference/speech-to-text/transcribe

prateeksachan: ^ ignore this. I guess they don't have this for streaming.

bolna/transcriber/assemblyai_transcriber.py
san0808: mainly for backward compatibility like existing agents still carry model: "universal" which is what's replayed at call time and AssemblyAI's enum has no universal. Without the alias those calls stop connecting the moment we start sending speech_model.

prateeksachan (approved):

